PR #1845, opened 2013-11-07
Update pods-ui-list-table.css

mrDlef:
Fix image not found if wordpress is not on the root.


szepeviktor:
Please do stop distributing 404s.
Could you comment out those PNGs from the CSS?

sc0ttkclark:
The code is already in Pods 2.5.2 I believe, http://pods.io/latest/

szepeviktor:
Thank you!
Maybe on WP.org also? https://wordpress.org/plugins/pods/changelog/

PR #2229, opened 2014-06-10
Update password.php

orchid-hybrid:
Initial commit adding password hashing support to password fields.

This adds two new options to the password field: Hash (a boolean) and Salt (a string). Hashing is enabled by default but is optional (for backwards compatability) and can be turned off.

We've chosen one iteration and 32 bytes for the size of the derived key. I think these values would be suitable in general.

Note that this means users should not inspect the value of a password field anymore: The way to interface with hashed passwords is to hash the value you want to test it against with the same salt and compare.

Reference for the choice of pbkdf2 with sha1: http://security.stackexchange.com/a/31846/45116

### DIFF
--- a/classes/fields/password.php
+++ b/classes/fields/password.php
@@ -97,19 +97,17 @@ class PodsField_Password extends PodsField {
      * @since 2.0
      */
     public function schema ( $options = null ) {
-        if( pods_var( self::$type . '_hash', $options ) ) {
-            return 'VARCHAR(32)';
-        }
-        else {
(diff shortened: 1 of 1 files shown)


lougreenwood:
Hey, I just wanted to bump this conversation and see whether the core implementation of password hashing is something that could be sponsored so it can be released ASAP?

lougreenwood:
bump... :)

sc0ttkclark:
@lougreenwood contact me on our Slack to discuss further

PR #2355, opened 2014-09-17
Don't force a default value in a format function

jamesgol:
This seems like a more sane fix to so many of the bugs related to default values. If
someone really wants to make 0, 0.00, etc. the default let them force it via the "Default
Value" attribute


Shelob9:
@jamesgol I tested this and still got an empty string for the field with a default value.

jamesgol:
@Shelob9 Did you have the field marked as required?  Otherwise an empty string is valid

Shelob9:
@jamesgol No, I'm wrong, I was testing this against another issue. Sorry. Will retest.

PR #2358, opened 2014-09-17
Change the default of display_errors to true

jamesgol:
A previous changed corrected the reversed logic that was used in display_error.  This should keep the previously expected behavior the same


pglewis:
Related: https://github.com/pods-framework/pods/issues/3060

Issue #2360, opened 2014-09-23
Simple Relationship (list) won't save a value with double quotes


maiki:
This may be related to the spirit of #1943. I've also posted this at https://wordpress.org/support/topic/how-do-i-get-the-lable-instead-of-the-value-from-a-list-with-get_post_meta.

My issue is that if I have double quotes in the value of a list item, it won't save when I update the post; it returns an empty field instead. In the past I got away with these items:

1|5.5" x 8.25"
2|6" x 9"
3|5.5" x 7.5"
4|5.25" x 7"
5|5.375" x 7"
6|5.5" x 7"
7|7" x 9"
8|7.25" x 9.5"
9|8" x 10"
10|9" x 7"

Those are the trim of a book (its size), and the reason this worked is because I would use a magic tag, `{@trim}`, and it gets the label. That works, so I didn't think about it much. Now I am moving my templates into the theme, and `<?php echo get_post_meta( get_the_ID(), 'trim', true); ?>` returns the value, which in this case isn't helpful.

As @Shelob9 pointed out, there are ways to get the label, but I'd rather figure out how to save the value I want, and just use `get_post_meta`.

I've tested the following modifications to the list:

|5.5" x 8.25"
6" x 9"|6" x 9"
5.5 x 7.5|5.5" x 7.5"
5.25" x 7"

Of those four, only the third, `5.5 x 7.5|5.5" x 7.5"`, will save the value. That may work in the meantime, but of course we would prefer to have the double quotes to refer to the inches in the size.

I've tried escaping the double quotes, with back slashes, but that didn't change anything. Suggestions? ^_^

quasel:
any plans to tackle this or shall we just close it - there is a PR since 2016 time to make a decision ;) @sc0ttkclark @jamesgol - or just add a doc with it's a known issue don't fo it :P

Issue #2042, opened 2014-02-19
comma character breaks custom post type dropdown


andsat:
extended post type (I extended post to allow for a "location" dropdown)

if you try to make the dropdowns contain a comma, for example: "Seattle, WA" on one line, "Renton, WA" on the 2nd line, it'll break.

this is entirely in the wordpress back-end. not theme-related.

seems to work fine if I just type "Seattle WA" instead. tried using &#44; (escape code) but it converts it to a comma and breaks it again.

PR #2361, opened 2014-09-23
Properly format custom-simple fields for comparison

jamesgol:
The foreach loop here breaks out the custom into an array, then the next foreach loop trys to compare the value
against this list but the array is sanitized so if there is any special chars in there it won't match.

Related to #2360

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -268,7 +268,9 @@ class PodsAPI {
 
 				if ( $simple ) {
 					delete_post_meta( $id, $meta_key );
-
+					if ( !has_filter( 'sanitize_post_meta__pods_' . $meta_key, 'pods_sanitize' ) ) {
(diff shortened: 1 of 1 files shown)


Shelob9:
@jamesgol Before I test this, can you resolve merge conflicts please? I will re-test right ASAP once you do.

sc0ttkclark:
@jamesgol revisit this when you're able

jamesgol:
Will do, going to be a bit.  I'm off to Seattle for the weekend and then I'll be in Baltimore all next week.

jamesgol:
I split this patch up since this PR covered 2 issues.  #3616 fixes having quoted strings in realtionship fields.  The other part of the patch covered #2042, but by changing the field separator to something other than a comma which I'm not sure if @sc0ttkclark was going to be happy with.

jamesgol:
Now that select2 claims to have fixed their issue in 4.0, I'll see if I can work back on the other issue and implement it without changing the field separator

Issue #2557, opened 2014-12-31
Update version constant


Shelob9:
PODS_VERSION should be 2.5 in master (it is 2.5-b1 now which is wrong). In 2.x branch should be 2.5-alpha1

PR #2563, opened 2014-12-31
set version to 2.5.1 alpha in 2.x branch 

Shelob9:
#2557

### DIFF
--- a/init.php
+++ b/init.php
@@ -3,7 +3,7 @@
 Plugin Name: Pods - Custom Content Types and Fields
 Plugin URI: http://pods.io/
 Description: Pods is a framework for creating, managing, and deploying customized content types and fields
-Version: 2.5.1-alpha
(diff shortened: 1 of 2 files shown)


init.php
```diff
@@ -37,7 +37,7 @@
 }
 else {
     // Current version
-    define( 'PODS_VERSION', '2.5-b-1' );
+    define( 'PODS_VERSION', '2.5-a-1' );
```
sc0ttkclark: 2.5.1-a-1

sc0ttkclark:
Places where versions need changing:

1. init.php in plugin header
2. init.php in PODS_VERSION constant
3. README.md at bottom in current version / stable etc
4. README.md at bottom in `~` area
5. readme.txt stable tag

Only 1, 2, and 4 are needed for the test builds.

pglewis:
Can we "DRY" some of these?

PR #2565, opened 2015-01-02
Potential fix for hidden fields show up for an admin user bug #2560

### LINKED ISSUE #2560: Hidden fields show up for an admin user
I'm doing quality control before I upgrade Pods from 2.4 -> 2.5 (I did not do interim updates).

I created a new pod called 'test' and created a default text field called 'private'.  I then checked 'Hide field from UI'.  As an administrator (the user who set it up), clicking 'Add new' for test shows that field.  This is different than functionality in v2.4, where the field would be hidden.  Can provide screenshots if necessary, but I think it is really simple to duplicate.

sc0ttkclark:
Could fix #2560


Shelob9:
I can still see the hidden field after eba54273cf73297a4aa57fc7331cf20aa7fa426f

sc0ttkclark:
Closing this as #2613 resolves the issue.

Issue #2277, opened 2014-06-30
Can't orderby with pods::field for relationship field


Shelob9:
When setting orderby in pods::field() params to order the results of a related field items it returns false. For example:

``` php
<?php
$pods = pods('author', array('limit'=> -1 ) );
if ( $pods->total() > 0 ) {
    while( $pods->fetch() )  {
        $books = $pods->field( array('name' => 'books', 'orderby' => 't.post_title ASC' ));
        var_dump( $books );
    }
}
```

returns false once per loop iteration. Taking out the orderby returns the related field values as expected. Is my usage wrong, or is this a bug?

PR #2570, opened 2015-01-03
Allow fields to be sorted by weight or by orderby

### LINKED ISSUE #2567: Drag & Drop multiple image field not saving order
I have added a multiple image field to my pod. When I try and re-order the images using the drag and drop facility it works just fine, however when I save the post the order is not saved it reverts to the original order. See screenshots below:

before
required_order

The first screenshot is the original order, the second is what I want but when the post is saved it just reverts to the original.

I have the latest version of Wordpress and Pods installed and very few other plugins so cannot see any problem from my side. Could this issue be tested?

Thanks
Robert

jamesgol:
In PR #2350 I attempted to fix bug #2277, and it did but also removed the sorting by the weight.

This should fix #2567 as well as not regressing the fix for #2277

### DIFF
--- a/classes/Pods.php
+++ b/classes/Pods.php
@@ -1380,9 +1380,19 @@ class Pods implements Iterator {
 									unset( $item_data );
 
 									// Return all of the data in the order expected
-									foreach ( $items as $id => $v ) {
-										if ( in_array( $id, $ids ) )
-											$data[ $id ] = $v;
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Will merge pending Travis-CI

sc0ttkclark:
I love that!

PR #2596, opened 2015-01-11
Fixing delete_field issue

### LINKED ISSUE #2593: Possible Issue with pods_api->delete_field()
I'm always getting 'Field not found' when using this function.  I think there's a bug here: https://github.com/pods-framework/pods/blob/2.x/classes/PodsAPI.php#L4549

Here's the code I'm using: https://gist.github.com/ckpicker/8eae792e7a5f2058ad6b

Any issues with that?

ckpicker:
This fixes #2593


sc0ttkclark:
Oops, sorry, very close :)

Just need to make sure the branch that you are doing the PR against is pods-framework:2.x and not master.

PR #2613, opened 2015-01-14
Change $group to be reference

### LINKED ISSUE #2560: Hidden fields show up for an admin user

jamesgol:
I believe this is what is intended, otherwise around line 510 $group gets blown away

Fixes #2560

### DIFF
--- a/ui/admin/form.php
+++ b/ui/admin/form.php
@@ -15,7 +15,7 @@ $groups = PodsInit::$meta->groups_get( $pod->pod_data[ 'type' ], $pod->pod_data[
 $group_fields = array();
 $submittable_fields = array();
 
-foreach ( $groups as &$group ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Rather than reference, can we just set $groups[ $g ] = $group; at the end of that foreach loop? Sorry, bad experience with references in the past.

jamesgol:
That works too, I'm not a huge fan of references either.

Issue #2534, opened 2014-12-18
Update Pods Logos


Shelob9:
Anything in https://github.com/pods-framework/pods/tree/master/ui/images that is a Pods logo should get updated with a new and improved one, using the same size, layout of image.

Shelob9:
New branding: https://drive.google.com/open?id=0B2iTBwlukqFtQTNYTXBkZF9rQlk&authuser=0

PR #2615, opened 2015-01-15
Initial pass at adding icon font

ghost1227:


Shelob9:
Related #2534 (update images for icons)

PR #2621, opened 2015-01-19
Make sure related_items are only cached for valid fields

### LINKED ISSUE #2620: Magic tag for Custom Taxonomy displays no content in Template after 2.5 upgrade
Magic tag for Custom Taxonomy displays no content in Template after 2.5 upgrade submitted via Slack by jimtrue

Problem represents that using a magic tax {@custom-taxonomy} that successfully output taxonomy information before upgrading to Pods 2.5 no longer displays any content after upgrading. 

Back revving to Patch 2.4.3 immediately resolves the issue. I'll test locally with debug console to see if any errors are being represented.

jamesgol:
I think the only case this gets hit is for custom taxonomy that isn't table based, since there is no pod associated
with it.

Fixes #2620

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -6634,7 +6634,7 @@ class PodsAPI {
         $ids = array_unique( array_filter( $ids ) );
 
 	    $idstring = implode( ',', $ids );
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
I haven't merged this file into Pods 3.0 so it's still fair game. Fix looks good to me.

Issue #1240, opened 2013-04-10
Display Field in selection list


MaartenvanLoon:
I'm not able to choose a field of a related pod to show. The direct database fields (of the related pod) I can show in the list, but not the related ones (so, these are related, to the related pod).

I'll give an example with some screenshots, because it sounds difficult like this :-)

I have a pod "employee", which had a related field to a WP user (an employee is a WP user and works for a certain company ... just one company = single select).

![Employee](https://f.cloud.github.com/assets/3957956/361440/eca59804-a1c3-11e2-9be0-0a029092eb5e.JPG)

bi-directional relationship company <--> employees (1 company --> many employees)

![Company_1](https://f.cloud.github.com/assets/3957956/361448/1f3d47ee-a1c4-11e2-9d50-66c5cfa2b793.JPG)

Now, to select/show employees for a company, I want to show fields of the WP user (like display_name or user_login)

![Company_2](https://f.cloud.github.com/assets/3957956/361466/62228d80-a1c4-11e2-87d4-04965c1fdbbf.JPG)

At this moment, on company level, I only see the emplyee ID's, which doesn't say much.

![Company_3](https://f.cloud.github.com/assets/3957956/361472/83fa979a-a1c4-11e2-90f4-a43bd099d2c7.JPG)

So, on company level, I want to show the employee with its name, which is not stored directly on "Employee" but on the "WP User".

Is that possible one way or another?

DavidRies:
Hi, I used this patch: https://github.com/pods-framework/pods/tree/feature/1240-displayfield and it solved the following for me.

SETUP
POD A: custom post type with text field "label"
POD B: custom post type with relationship field named "group" related to POD A and the relationship field's "Display Field in Selection List" set to {@label}

CURRENT RELEASE
On the POD B edit screen, the dropdown options all said "(No Title)"

WITH THIS PATCH
On the POD B edit screen, the dropdown displayed the values of the label field as expected.

Thanks!!

Shelob9:
fixed via #2654

jimtrue:
This has not been fixed. This is the question about reaching across in a Relationship Field to a connected table's relationship field and showing values from THAT table.

FLHerne:
Per 'Jim True' on Slack, after trying to use it here, this isn't actually fixed. Trying to get the string via a relationship in the target still fails.

jimtrue:
The workaround for this one is as indicated above by @mastef 
[...]

Issue #2652, opened 2015-01-26
Improper prefixing used for "Display Field" for relationship


Shelob9:
As [reported in the forums](http://pods.io/forums/topic/relationships-no-longer-working/) a user has a relationship between meta storage custom post types, and is setting a custom display field. In this case `{@company_name}`. Then when they got to the post editor, they see this error:

`WordPress database error: [Unknown column ‘t.company_name’ in ‘field list’]
SELECT DISTINCT t.ID, t.company_name, t.post_type FROM wp_posts AS t WHERE ( ( t.post_status IN ( “publish” ) ) AND ( t.post_type = “company” ) ) ORDER BY t.menu_order, t.post_title, t.post_date /* From [www.devkitmagazine.com/wp-admin/post-new.php?post_type=product] in [/nas/wp/www/cluster-2214/devkit/wp-content/plugins/pods/classes/PodsData.php:2055] */
Database Error; SQL: SELECT DISTINCT t.ID, t.company_name, t.post_type FROM wp_posts AS t WHERE ( ( t.post_status IN ( “publish” ) ) AND ( t.post_type = “company” ) ) ORDER BY t.menu_order, t.post_title, t.post_date; Response: Unknown column ‘t.company_name’ in ‘field list’
`

Looks to me like a prefixing problem as the unknown column `t.company_name` wouldn't exist. The relationship is from a field called "company_name" to a Pods that has a field called "company_name". So the query should be for "company_name.company_name" right?

And yes, maybe the issue is that this is an edge case where same.same traversal causes a problem.

Shelob9:
My theory on the edge case seems to be invalid: http://pods.io/forums/topic/relationships-no-longer-working/#post-188880

Shelob9:
Debug information form reporting user:

Debug Information

WordPress Version: 4.1

PHP Version: 5.3.2-1

MySQL Version: 5.5.40

Server Software: Apache

Your User Agent: Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.91 Safari/537.36

Session Save Path: /var/lib/php5

Session Save Path Exists: Yes

Session Save Path Writeable: Yes

Session Max Lifetime: 1440

WPDB Prefix: wp_

WP Multisite Mode: No

WP Memory Limit: 40M

Pods Network-Wide Activated: No

Pods Install Location: /nas/wp/www/cluster-2214/devkit/wp-content/plugins/pods/

[...]

PR #2654, opened 2015-01-27
Allow Display Field to work with relationships on CPT

jamesgol:
Related to #1240 and #2652


sc0ttkclark:
What should we add for unit testing on this?

jimtrue:
This did not fix #1240 because this only fixed display field for CPT's, not for reaching across and providing values through a relationship field into another connected Pod.

Issue #2376, opened 2014-10-02
first_id and last_id should check if total_found is equal to total, if so, use local result set and do not perform a separate find() query


sc0ttkclark:
first_id and last_id should check if total_found is equal to total, if so, use local result set and do not perform a separate find() query

PR #2664, opened 2015-01-30
Make last_id return a value

jamesgol:
Somewhat related to issue #2376


sc0ttkclark:
Not sure who removed it, whoever did gets -- in Slack

Issue #2665, opened 2015-01-30
CodeMirror fields won't save


jamesgol:
Can't reproduce on CPT for some odd reason, but on ACT codemirror field isn't saved.

Looks like CodeMirror dropped the 'onBlur' at 3.0

PR coming

sc0ttkclark:
Closed by #2666

cspicuzza:
I upgraded a multisite to 2.5.1.2 and noticed the CodeMirror fields still are not saving.  Was this fix included with that release?

sc0ttkclark:
No it's in Pods 2.5.2 http://pods.io/latest/

cspicuzza:
I probably should of picked up on that... woops.  Thanks!

PR #2666, opened 2015-01-30
Change CodeMirror to not use onBlur

jamesgol:
Apparently around CodeMirror 3.0 they dropped onBlur.  So the value wasn't being saved


sc0ttkclark:
Fixes #2665

PR #2674, opened 2015-02-03
Don't let Pods overwrite admin menu items

### LINKED ISSUE #2673: Too many Pods will overwrite admin menu
Using add_object_page doesn't check the count of items in the admin menu so it can overwrite.  I noticed that my plugins and themes menus disappeared when I had a lot of Pods on my test setup.

PR incoming

jamesgol:
Fixed #2673

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -224,9 +224,8 @@ class PodsAdmin {
                                 $parent_page = $page = 'pods-manage-' . $pod[ 'name' ];
 
                                 if ( empty( $menu_position ) )
-                                    add_object_page( $page_title, $menu_label, 'read', $parent_page, '', $menu_icon );
-                                else
-                                    add_menu_page( $page_title, $menu_label, 'read', $parent_page, '', $menu_icon, $menu_position );
+                                    $menu_position = null;
+                                add_menu_page( $page_title, $menu_label, 'read', $parent_page, '', $menu_icon, $menu_position );
 
                                 $all_title = $plural_label;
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
I'm not 100% sure yet on this, since if a object doesn't have a menu position set, it will go to the bottom of the menu items, instead of the bottom of the object menu group. Moving all current Pods for people who may be used to finding them where they're at now. Does it sound better if we just internally track menu position via a local incremental variable, if empty, just increment and use a decimal to avoid conflicts.

So 3 menus:

* 6.001
* 6.002
* 6.003

jamesgol:
On my test setup (with about 65 pods that are a mix of everything) it doesn't really change the order except not overwriting anything.  I think one of them has a position explicitly set but the rest are default.  The groupings are the same.  CPT's up top below "Comments".  ACT start after "Settings" and go up to "Pods Admin"

sc0ttkclark:
Line 207 has the $menu_position null bit, so you can take out the others.

I'm still wondering why add_menu_page is performing like that, when it's null it appends to the end:

```php
$new_menu = array( $menu_title, $capability, $menu_slug, $page_title, 'menu-top ' . $icon_class . $hookname, $hookname, $icon_url );

	if ( null === $position )
		$menu[] = $new_menu;
	else
		$menu[$position] = $new_menu;
```

sc0ttkclark:
Let's get one more person doing a test on this with lots of Pods to be sure.

Shelob9:
I tested and this works as expected. Merging.

PR #2678, opened 2015-02-04
fixes codepress/admin-columns-pro-issues/issues/80

tobiasschutter:
Improved integration with Admin Columns. Added custom fields for Taxonomy support. Meta keys will now be listed for the Custom Field type and it's column value displayed on the taxonomy listings screen.


sc0ttkclark:
Fixed via rebased 2.x PR #2682

Issue #988, opened 2013-01-15
Validate one of multiple fields is not empty


mertnuhoglu:
I have a form where I want to validate that either one of two fields is not empty. 

I wrote the following javascript function to do this:

```
$("form.pods-form-pod-katilimci").submit(function(event){
    telefon = $("form.pods-form-pod-katilimci #pods-form-ui-pods-field-oncelikli-telefon");
    email = $("form.pods-form-pod-katilimci #pods-form-ui-pods-field-email");
    if( $.trim(telefon.val()).length == 0 && $.trim(email.val()).length == 0) {
        alert("You have to fill either telefon or email.");
        event.preventDefault();
        return false;
    }
});
```

The function works properly except that it cannot prevent submitting of the form when validation fails. 

What should I do to stop the form submitting?

PR #2687, opened 2015-02-05
Add a css helper like ".pods-form-front-success" but for errors

developez:
Usefull for this kind of situations:

https://github.com/pods-framework/pods/issues/988

### DIFF
--- a/ui/css/pods-form.css
+++ b/ui/css/pods-form.css
@@ -290,6 +290,14 @@ input.pods-form-ui-field-type-color {
     background: #DFF2BF url(../images/green-check.png) no-repeat 10px center;
 }
 
+.pods-form-front-error {
+    color: #8A1010;
+    border: 1px solid;
+    margin: 10px 0;
+    padding: 15px 10px 15px 50px;
+    background: #F2D8BF url(../images/red-x.png) no-repeat 10px center;
+}
+
 .pods-form-ui-row-type-hidden {
     display: none;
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Please base your PRs / commits off of the "2.x" branch if it's a fix, or "release/3.0" if it's an enhancement or feature (3.0 is the next major release).

Also, I'm not sure yet what we want to do to address multiple error validation and the class applied, we're still figuring a few things out in that part of the code for Pods 3.0.

developez:
The point is that if you want to do an additional validation in the pods forms using jquery (like here: https://github.com/pods-framework/pods/issues/988), it would be a good idea to get a similar style to show the validation error.

The thing is to do this:

	$(document).ready(function() {
														
		pods_admin_submit_validation = function ( valid_form, $submittable ) {
			var user_pass = $.trim( $submittable.find( '#pods-form-ui-pods-field-user-pass' ).val() ),
				confirm_user_pass = $.trim( $submittable.find( '#pods-form-ui-pods-field-confirm-user-pass' ).val() );

			if ( user_pass !=  confirm_user_pass) {
				$("#message-form-container").html('<div id="message" class="pods-form-front-error"><?php _e('blah ..','foo'); ?></div>');
				return false;
			}

			return valid_form;
		};
		
	});

"pods-form-front-error" style draws a red square where put the msg, like pods-form-front-success.

PR #2706, opened 2015-02-09
Updating .button.button-primary in  ui/css/pods-wizard.css

NikV:
Making the button more better and simple by taking away gradients. Better matches the redesigned Pods Wizard.


sc0ttkclark:
All CSS PRs need a screenshot of at least the patched version. It's nice-to-have a screenshot of how it was before too for reference, but not required.

PR #2722, opened 2015-02-12
Make sure autoloader doesn't load the same class more than once

### LINKED ISSUE #2721: 3.0 autoloader dies if trying to load the same class more than once
PR incoming

jamesgol:
Two distinct but related issues here.  First the code would try to alias classes more than once which caused php to complain that the class already existed.  Second the autoloader would try to 'require' files more than once which caused php to complain.

Fixes #2721

### DIFF
--- a/classes/Pods/ClassLoader.php
+++ b/classes/Pods/ClassLoader.php
@@ -133,7 +133,7 @@ class Pods_ClassLoader {
 		}
 
 		if ( $file = $this->findFile( $className ) ) {
-			require $file;
+			require_once $file;
 
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


jamesgol:
Leaving this for @sc0ttkclark to merge.  I'm not sure why require was used instead of require_once but maybe there was a good reason

jamesgol:
From @unknownnf:  require was used for simplicity, but in the case of aliases it should have been require_once aliases were added after the autoloader. It's still PSR-0 either way. Ideally the class loader would not be called for a class that has already been loaded.

So I'm going to merge this.

PR #2736, opened 2015-02-15
Use composer for including cmb2

### LINKED ISSUE #2710: Use Composer For CMB2 Dependency & Set Correct Install Path
A part of  #2517

Shelob9:
resolves #2710 

MERGE BLOCK: no longer valid.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "includes/cmb2"]
-	path = includes/cmb2
-	url = git@github.com:WebDevStudios/CMB2.git
--- a/README.md
+++ b/README.md
@@ -3,6 +3,13 @@
 
 **Pods is a development framework for creating, extending, managing, and deploying customized content types in WordPress.**
 
+## 3.0 Is In Development & Is Not Functional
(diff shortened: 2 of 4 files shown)


Shelob9:
Assigning to @sc0ttkclark  for merge

m-e-h:
You could use grunt to move the cmb2 folder back to the includes folder with https://github.com/gruntjs/grunt-contrib-copy
You could even use grunt to initiate the composer command https://www.npmjs.com/package/grunt-composer
I'm not that familiar with grunt but I've done this in projects with the gulp equivalent.

Shelob9:
@m-e-h I could move it with Grunt, and if you're interested in how, I'll show you.

Right now I'm setting a custom install path of vendor, since it's type is wordpress-plugin, which is a problem.

I could move its path to includes, but I don't see why it should be there. Dependencies go in vendor. Includes is for our own code. If you want to make a reasoned argument against that opinion, please do.

m-e-h:
Oh ok. I saw in a comment somewhere on this issue mention of grunt and since the folder used to be in includes... just thought I'd throw it out there in case you weren't aware.
Really appreciate all you guys do. Trying to help anyway I can.

PR #2739, opened 2015-02-16
Properly build related object table

### LINKED ISSUE #2737: pods_register_related_object stopped working with 2.5
I posted on the pods.io forum but someone mentioned I try posting here as well. Here's my initial post:
http://pods.io/forums/topic/ok-pods_register_related_object-really-not-working-at-2-5-1-1/#post-189173

I’m dealing with a bug that is extremely weird, among the most frustrating I’ve encountered.

When I say bug, I don’t mean Pods offhand, my system is complicated, but what I do know is if I roll back Pods to 2.4.4-alpha or even 2.4.3 my custom object pick select lists are populated.

Here’s where it gets weird. If I swap out 2.4.3 for 2.5 (or 2.5.1.1) and refresh the page the pick objects continue to work and are populated. Refresh after refresh, no problem… until I clear the Pods cache, then the lists won’t populate.

I’ve traced this down a lot into the code. In /pods/classes/fields/pick.php PodsField_Pick::register_related_object() if I dump put $related_object after the array merge my data for the list is there. It seems to get lost however somewhere with the handling of self::$custom_related_objects.

Is there anything you can tell me offhand what might be causing this? Anything that changed between 2.4.3 and 2.5 that might cause this kind of behavior?

I know this is crazy, I’ve been working on resolving this for days, any help would be greatly appreciated. 

Thanks,

-Marc

jamesgol:
If pods_register_related_object() is called late in the game then the objects aren't actually added to the array
because the optimizations added for 2.5 keep the tables from constantly being rebuilded.  This patch lets the items to be added without doing a full rebuild.  Also, a force rebuild before would have loaded the transient and thrown the data away.

Fixes #2737

### DIFF
--- a/classes/fields/pick.php
+++ b/classes/fields/pick.php
@@ -353,6 +353,7 @@ class PodsField_Pick extends PodsField {
         }
 
         if ( empty( self::$related_objects ) ) {
+                // Do a complete build of related_objects
             $new_data_loaded = true;
 
             // Custom
--- a/classes/fields/pick.php
+++ b/classes/fields/pick.php
@@ -339,15 +339,22 @@ class PodsField_Pick extends PodsField {
      * @since 2.3
      */
     public function setup_related_objects ( $force = false ) {
-	    if ( !$force && ! empty( self::$related_objects ) ) {
-		    // Check if we've already been setup
-		    return false;
-	    }
-        $related_objects = pods_transient_get( 'pods_related_objects' );
+        $new_data_loaded = false;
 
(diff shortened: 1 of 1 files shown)


jamesgol:
Unless I'm missing something else the Travis CI build failed because it couldn't connect to develop.svn.wordpress.org

```
svn: OPTIONS of 'https://develop.svn.wordpress.org/trunk/tests/phpunit/includes': could not connect to server (https://develop.svn.wordpress.org)
```

PR #2764, opened 2015-02-28
travis-ci: unit test php 7

Ramoonus:


sc0ttkclark:
Should we follow what bbPress did by using `nightly` instead and also adding it to allowed failures?

https://bbpress.trac.wordpress.org/changeset/5619

Ramoonus:
I agree
since PHP 7 is not a final version but a nightly

sc0ttkclark:
I'll leave that change to you, then I'll merge.

sc0ttkclark:
Please recreate patch based on 2.x branch instead please.

Issue #2271, opened 2014-06-27
@{post_content} strips out all tags in a Template when there are no Fields


szepeviktor:
`@{post_content}` strips out e.g. `<p style="text-align: center;">`
If I add a Field to the pod it returns the raw content.

logoscreative:
I can confirm this.

If there are no custom fields, and I use `{@post_content}` in a template, the HTML in the post content is stripped out. As soon as I add one field, the HTML is outputted properly.

jamesgol:
I couldn't reproduce this either.  I tried enabling/disabling anything that should be relevant and couldn't come up with anything.

If you can give more specific steps to reproduce that would go a long way.

logoscreative:
Ah! Looks like this happens when listing multiple posts, like `[pods name="test" template="Test"]`.

I was able to reproduce this again. Only Pods and Frontier Auto Template are enabled.
1. Create a new Pods CPT (don't do anything else to it). Add one instance of the CPT, and put some HTML inside of the content area.
1. Add a new template that uses `{@post_content}`.
1. Insert the shortcode that lists the pods with the template into a page/post/whatever.

The shortcode that lists the Pods will strip out the HTML. However, if you add literally any field to the Pod and save it, the HTML will output correctly for the list shortcode.

Shelob9:
@logoscreative Did you test pull request #2766 to see if it resolves your issue?

jamesgol:
I can confirm that I was able to reproduce the issue with the additional info @logoscreative provided.

Unfortunately when updating to the last version that includes #2766, now no data is passed at all

jamesgol:
#2765 broke this

sc0ttkclark:
For #2765, just need to ensure if form="1" is set then we require $params to not be empty on that line.

Shelob9:
Fixed via #2769

szepeviktor:
Thank you!
When do you plan to release 3.0?
Will it be faster?

Shelob9:
@szepeviktor We will release 3.0 when it is ready. I know that sounds dismissive, but it's the best answer we can give.

szepeviktor:
It is promising that tha wheels are turning. I've never thought it'll be done. But it is! 10 months.

szepeviktor:
Thank you again.

jamesgol:
Sorry haven't had a chance to figure out the right way to test this. Way too much life stuff going on right now

PR #2769, opened 2015-03-02
Make sure the fields array is properly built

jamesgol:
Related to #2271


sc0ttkclark:
Pending Travis-CI pass

jamesgol:
This doesn't fix the bug introduced by #2765 but it fixes the original issue.  This one was interesting.  What was happening is that when there was no ID passed then PodsData:select() gets called and it tries to add a $pods->field array.  The only data in the array it sets is 'label' so when PodsForm:fields_setup() gets called there is no $field[ 'type' ] set so the default of 'text' is used.   Which will make field() strip out the html when display is called

PR #2780, opened 2015-03-05
Feature/2779 term link

### LINKED ISSUE #2779: .permalink traversal for related taxonomies don't work
.permalink traversal for related taxonomies don't work submitted via Slack by josh

Shelob9:
Should fix #2779


Shelob9:
@sc0ttkclark or @jamesgol please see my line note in  3d2494b and let me know if you agree with my logic. I don't see any way that this new conditional could affect any other usage of `Pods::field()` since it would only run in this very specific case, but let's see what traversal tests say.

PR #2782, opened 2015-03-06
travis-ci: add php nightly but allow failures

Ramoonus:
this time validated using lint
#2774  #2781

PR #2785, opened 2015-03-07
travis-ci: only build php nightly with latest stable wordpress

Ramoonus:
this removed 8 builds / run
which saves nearly 2 hours per UT session

#2782


sc0ttkclark:
Let's do this with hhvm too.

Do we have a 3.0 version of this PR too yet?

Ramoonus:
not yet, it`s on my todo list

Issue #2778, opened 2015-03-05
Clarify what does and doesn't get deleted on reset.


Shelob9:
What we have now:

![screen shot 2015-03-05 at 3 06 41 pm](https://cloud.githubusercontent.com/assets/1994311/6513471/4124a566-c349-11e4-810e-c6c37cae6646.png)

"This does not remove any items from any Post Types, Taxonomies, Media, Users, or Comments data you have added/modified." Is technically correct, but it could be clearer, as anything in a custom table will be deleted.

The following sentence needs append:

"Any custom fields stored using the table storage component, content in Advanced Content Types, and relationships between posts will be lost."

Needs 2.x and 3.x patch.

Shelob9:
Fixed in 2.x and 3.0 by @NikV via #2792 (3.0) and #2784 (2.x)

PR #2792, opened 2015-03-10
Clarify what does and doesn't get deleted on reset

NikV:
https://github.com/pods-framework/pods/issues/2778


Shelob9:
@NikV On this and the 2.x one, can you please redo to move the HTML out of the translation string. I know it was like that when you found it, but might as well improve it while we are in there.

NikV:
Should I remove completely, or just rearrange it?

Shelob9:
Use sprinf() to separate markup from the translation string.

NikV:
Hey @Shelob9, before I fix the 2.0 branch I just added a commit for 3.0, does that look OK?

Shelob9:
@NikV Perfect. Please apply same fix to 2.x

Issue #2658, opened 2015-01-28
Network paths on windows system gets stripped in PodsView::locate_template


cocoanton:
Using Pods 2.5.1 on a clients windows server with WP4.1. 
They are using a network path as the real path.
This ends up being an problem when pods try to render the templates because locate_template strips the leading \\.

This is due to the replacement happening in the locate_template [function at row 576](https://github.com/pods-framework/pods/blob/master/classes/PodsView.php#L576)
I actually don't think these are necessary since realpath() already does the stripping and optimizing of the paths and the function should rely on that instead. Removing these two lines fixes the problem for my clients install. 

Is there any other reason these two replacements are done? Or should we just let realpath() do the lifting in that matter.

sc0ttkclark:
Fixed via #2793 and #2795

PR #2793, opened 2015-03-12
Removing double-slash trimming

cocoanton:
This fixes issue #2658  
The removed trimming was causing issues on windows network-mounted filesystems. the function realpath() already does this for us.


sc0ttkclark:
Needs a patch for release/3.0 too

cocoanton:
So a new pullrequest?

sc0ttkclark:
Yes, in addition to this one if you've got the time. Otherwise we can't merge this into Pods 2.x until we have a identical PR for Pods 3.0, which we can do for you but won't have it prioritized over existing work we're doing at the moment.

PR #2805, opened 2015-03-21
deprecated phpdoc typos fixed

Ramoonus:


sc0ttkclark:
This should only be done for 2.x and release/3.0 instead of master.

PR #2824, opened 2015-04-03
NikV's 3.0 docs revamp

NikV:
This is just me fixing a bunch of random docs stuff. Also, there could be more PRs like this, beware...

### DIFF
--- a/classes/Pods/Field/Website.php
+++ b/classes/Pods/Field/Website.php
@@ -264,7 +264,7 @@ class Pods_Field_Website extends
 	}
 
 	/**
-	 * Build an url
+	 * Build a url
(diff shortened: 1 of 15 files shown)


sc0ttkclark:
Looks good

Shelob9:
I'm not going to revert this, as that would make @pglewis life even more difficult, but @NikV can you please do a new Pull Request to complete the inline docs you started here, so they meet the WordPress inline docs standards? IE Add an @since for every function/method/property you are documenting (@since unknown is OK for now) and add at least a short description (that is a complete sentence, ending in a period) for each one. Also @return needs to describe what is returned.

PR #2825, opened 2015-04-03
Translation fixes for help-resources.php

NikV:
And, by accident, hook docs for the pods_meta_init action ...


NikV:
unfinished, just noticed :/

PR #2827, opened 2015-04-05
update the importer to the newest version

tommiekeeston:
cptui changed the name of their option vars, now are :
cpt_custom_tax_types -> cptui_post_types
cpt_custom_post_types -> cptui_post_types

i've tested the script and its importing correctly i guess, maybe its missing some field but its working.

the function to clean after import doesn't work and need to be checked

### DIFF
--- a/components/Migrate-CPTUI/Migrate-CPTUI.php
+++ b/components/Migrate-CPTUI/Migrate-CPTUI.php
@@ -33,8 +33,8 @@ class Pods_Migrate_CPTUI extends PodsComponent {
      * @since 2.0
      */
     public function __construct () {
-        $this->post_types = (array) get_option( 'cpt_custom_post_types', array() );
-        $this->taxonomies = (array) get_option( 'cpt_custom_tax_types', array() );
+        $this->post_types = (array) get_option( 'cptui_post_types', array() );
+        $this->taxonomies = (array) get_option( 'cptui_taxonomies', array() );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


NikV:
I have tested this PR with taxonomies and custom post types with the simple importer and can confirm that so far, it works with the new version. Not sure of the problems with the clean function. Also, what missing field problems did you have @tommiekeeston?

sc0ttkclark:
We need to find a way to make this version dependent on what version CPTUI they are actually running. Currently this code will make older versions of CPTUI no longer work to migrate, so that would require them upgrade CPTUI first before importing into Pods.

@tw2113 Does the new CPTUI remove it's old options? Basically wondering if we could run into any potential issues with CPT / CT that may have been deleted since they upgraded -- and them showing up as the result of our backwards compatibility fix.

tw2113:
Nope, we leave the options alone for the 0.9.x and earlier versions. The user would have to manually delete them to get them removed, or have some sort of data loss occur. Not something we trigger at all ourselves though.

sc0ttkclark:
@tommiekeeston we'll want to check the CPT_VERSION constant here.

```php
if ( defined( 'CPT_VERSION' ) && version_compare( '0.9.0', CPT_VERSION, '<=' ) ) {
    // New options get / update
} else {
    // Old options get / update
}
```

Do two if/else statements, one when migrating, one when cleanup (update).

The only difference for the cleanup is that we want to delete both the old and the new versions of the CPTUI options.

sc0ttkclark:
We also need this PR redone for two branches instead of `master`.

One for Pods 2.x (`2.x` branch)

One for Pods 3.0 (`release/3.0` branch)

Issue #2660, opened 2015-01-28
Prepare Pods for Term Splitting in WP 4.2


PodsBot:
```php
/**
 * Fires after a previously shared taxonomy term is split into two separate terms.
 *
 * @since 4.2.0
 *
 * @param int $term_id ID of the formerly shared term.
 * @param int $new_term_id ID of the new term created for the $term_taxonomy_id.
 * @param int $term_taxonomy_id ID for the term_taxonomy row affected by the split.
 * @param string $taxonomy Taxonomy for the split term.
 */
do_action( 'split_shared_term', $term_id, $new_term_id, $term_taxonomy_id, $taxonomy );
```

## Pod table

- [x] If $taxonomy is a Pod and storage_type = table `UPDATE wp_pods_{pod_name} SET id = $new_term_id WHERE id = $term_id`

## Pods relationship table

- [x] Replace all podsrel IDs in `wp_podsrel` where related to that taxonomy + $term_id (any relationship field for pick_object = taxonomy, pick_val = {$taxonomy})
- [x] Replace all podsrel IDs in `wp_podsrel` where related from that taxonomy + $term_id (any relationship field for pick_object = taxonomy, pick_val = {$taxonomy})

## Pods (type=post_type) w/ relationship fields to that taxonomy

- [x] Replace all post meta related IDs in where related to that taxonomy + $term_id (any relationship field for pick_object = taxonomy, pick_val = {$taxonomy}) [keys = `_pods_{field_name}` has serialized array, `{field_name}` has individual values in separate rows]

## Pods (type=comment) w/ relationship fields to that taxonomy

- [x] Replace all comment meta related IDs in where related to that taxonomy + $term_id (any relationship field for pick_object = taxonomy, pick_val = {$taxonomy}) [keys = `_pods_{field_name}` has serialized array, `{field_name}` has individual values in separate rows]

## Pods (type=user) w/ relationship fields to that taxonomy

- [x] Replace all user meta related IDs in where related to that taxonomy + $term_id (any relationship field for pick_object = taxonomy, pick_val = {$taxonomy}) [keys = `_pods_{field_name}` has serialized array, `{field_name}` has individual values in separate rows]

## Pods (type=setting) w/ relationship fields to that taxonomy

- [x] Replace all options related IDs in where related to that taxonomy + $term_id (any relationship field for pick_object = taxonomy, pick_val = {$taxonomy}) [keys = `_pods_{field_name}` has serialized array, `{field_name}` has serialized array]

[...]

PR #2856, opened 2015-04-22
Support for Term Splitting in WP 4.2

sc0ttkclark:
This code addresses #2660

### DIFF
--- a/classes/PodsTermSplitting.php
+++ b/classes/PodsTermSplitting.php
@@ -30,6 +30,7 @@ class Pods_Term_Splitting {
 			}
 		}
 
+		// Track down all fields related to the target taxonomy and update stored term IDs as necessary
 		self::update_relationships_to_term( $term_id, $new_term_id, $taxonomy );
 
 	}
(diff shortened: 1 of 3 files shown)


classes/PodsTermSplitting.php
```diff
@@ -0,0 +1,515 @@
...
+			$wpdb->update( $table, $data, $where, $format, $where_format );
+
+			$this->append_progress( $task );
+		}
+
+		// Fix up the serialized data
+		$task = "update_setting_meta_{$pod_name}_{$field_name}_serialized";
+		if ( ! $this->have_done( $task ) ) {
+
+			$target_serialized = sprintf( ';i:%s;', $this->term_id );
```
sc0ttkclark: These probably should have been `%d`

pglewis: No, the fields are string fields and need quoted, string values.

sc0ttkclark: But `;i:%s;` is saying integer, so it would be `;i:%d;`

pglewis: Integer is how it's to be interpreted when unserialized.  It is still stored as a string in the database and thus we need to build a string.

PR #2870, opened 2015-04-28
Add Pod name to cache key

### LINKED ISSUE #2869: Pods caching with 'expires' returns cache for wrong post type
We've found a bug in the PodsData select function. Consider the following example. I have a page with multiple pods calls which are retrieved from the cache:
```
$werkplaatsacties = pods('werkplaatsacties', array('where' => 'merk.slug = "'.$merk.'"', 'limit' => '10', 'expires' => HOUR_IN_SECONDS));
$werkplaatstips   = pods('werkplaatstips', array('where' => 'merk.slug = "'.$merk.'"', 'limit' => '10', 'expires' => HOUR_IN_SECONDS));
$downloads        = pods('downloads', array('where' => 'merk.slug = "'.$merk.'"', 'limit' => '10', 'expires' => HOUR_IN_SECONDS));
```
As you can see the params for these three calls are exactly the same. The only difference is the pod name. Fetching this without cache works, fetching this with cache results in only the first query working. After investigation I found that the second and third query contained the results for the first query. Disabling expire or changing some params around fixed this issue, but not permanently. 

After a lot of debugging we found the culprit:
https://github.com/pods-framework/pods/blob/master/classes/PodsData.php#L635

The params for the query are serialized to generate a cache key, but not the pod name. This explains the behavior in my example.
A workaround that fixes this is adding a random extra parameter to each query: 
```
$werkplaatsacties = pods('werkplaatsacties', array('where' => 'merk.slug = "'.$merk.'"', 'limit' => '10', 'expires' => HOUR_IN_SECONDS, 'pod' => 'werkplaatsacties'));
$werkplaatstips = pods('werkplaatstips', array('where' => 'merk.slug = "'.$merk.'"', 'limit' => '10', 'expires' => HOUR_IN_SECONDS, 'pod' => 'werkplaatstips'));
$downloads = pods('downloads', array('where' => 'merk.slug = "'.$merk.'"', 'limit' => '10', 'expires' => HOUR_IN_SECONDS, 'pod' => 'downloads'));
```
I've added 'pod' => 'podname' to each call, but it doesn't really matter what you add, as long as it's key-value and different to the other queries it works. The definite fix should be adding the pod name to the MD5 hash that is used for caching. I would've done a pull request, but I'm not quite sure where the cache is being set.

Shelob9:
Fixes #2869

### DIFF
--- a/classes/PodsData.php
+++ b/classes/PodsData.php
@@ -632,7 +632,7 @@ class PodsData {
 
         // Get from cache if enabled
         if ( null !== pods_v( 'expires', $params, null, false ) ) {
-            $cache_key = md5( serialize( $params ) );
+            $cache_key = md5( (string) $this->pod . serialize( $params ) );
 
             $results = pods_view_get( $cache_key, pods_v( 'cache_mode', $params, 'cache', true ), 'pods_data_select' );
 


sc0ttkclark:
3.0 pr?

PR #2910, opened 2015-05-09
Stop saving _pods_[field_name] serialized when saving single select pick field values (2.0)

NikV:

PR #2925, opened 2015-05-13
Revert "Stop saving _pods_[field_name] serialized when saving single …

Shelob9:
Reverts pods-framework/pods#2910

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -3792,18 +3792,16 @@ class PodsAPI {
                 $object_type = 'post';
 
             delete_metadata( $object_type, $id, $field[ 'name' ] );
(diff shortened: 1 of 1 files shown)


classes/PodsAPI.php
```diff
@@ -3792,18 +3792,16 @@ public function save_relationships ( $id, $related_ids, $pod, $field ) {
                 $object_type = 'post';
 
             delete_metadata( $object_type, $id, $field[ 'name' ] );
-            delete_metadata( $object_type, $id, '_pods_' . $field[ 'field' ] );
```
sc0ttkclark: I thought this only needed changing to `$field['name']`

PR #2933, opened 2015-05-14
Fixes  #2932

### LINKED ISSUE #2932: shortcode pagination=false sets the flag as true
[pods name="foo"  pagination=false] or [pods name="foo"  pagination="false"]
  
This will show the pagination links.

str:
[pods name="foo" pagination=false] or [pods name="foo" pagination="false"]

This will show the pagination links because (boolean) "false" == true;

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -639,8 +639,8 @@ function pods_shortcode ( $tags, $content = null ) {
 
     $tags = apply_filters( 'pods_shortcode', $tags );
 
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
This change looks great, but can you commit and PR against the `release/3.0` branch instead? `master` is reserved for releases and `2.x` is something we're phasing out to focus more on getting Pods 3.0 ready to go beta.

PR #2935, opened 2015-05-14
Fixed  #2934

### LINKED ISSUE #2934: shortcode forces a limit when no pagination was requested
Only showing 15 elements is confusing when in the shortcode there was no limit set nor pagination set. It should either set no limit or set pagination true

str:
$limit should only be set if pagination was set

### DIFF
--- a/classes/Pods.php
+++ b/classes/Pods.php
@@ -2121,8 +2121,12 @@ class Pods implements Iterator {
 	 * @since 2.0
 	 * @link http://pods.io/docs/find/
 	 */
-	public function find ( $params = null, $limit = 15, $where = null, $sql = null ) {
-
+	public function find ( $params = null, $limit = 0, $where = null, $sql = null ) {
+        
+        // we only set a default limit if a pagination is set to true
+		if (!empty($params['pagination'])) {
+			$limit = 15;
+		}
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Really appreciate the PR but I don't see this being able to change. Pods handles so many use-cases that this could explode a lot of things out there already and having no predetermined limit would put new coders at risk of taking down their site in some cases when dealing with lots of data in their Pods and memory limits.

Still very interested in any other ideas you have, this is just a no-go for us right now. Thanks for your effort though, I really appreciate you jumping into GitHub and putting out a PR so quickly.

str:
But current defaults are broken.  If I create a very simple shortcode like this: [pods name="foo"]  it will show a page of 15 elements with no pagination links. I understand it could explode for all the cases where people haven't seen they are wrongly only showing 15 elements.  

If you think having a preset limit is needed, then please have the pagination="true" by default too.

sc0ttkclark:
Pods has always shown 15 at a time, WordPress shows whatever the default posts per page is set to in the settings page. The difference here is that when using the shortcode, you have to specify whether you want to show pagination or not (default is off), and in PHP you have to manually put in pagination for both WordPress and Pods, neither just include pagination.

Issue #2931, opened 2015-05-14
improve grunt release process


PodsBot:
improve grunt release process submitted via Slack by josh

Shelob9:
Need to:
1) Fix svn checkout
2) Include dependencies in package.json
3) Make sure master branch Github updater meta in master has the right branch `GitHub Branch: master`
4) Set 2.x to next version a-1 immediately.

PR #2938, opened 2015-05-15
each branch designation must match the branch

afragen:
The `GitHub Branch: master` must be the branch listed in version on wp.org.

### DIFF
--- a/init.php
+++ b/init.php
@@ -9,7 +9,7 @@ Author URI: http://pods.io/about/
 Text Domain: pods
 Domain Path: /languages/
 GitHub Plugin URI: https://github.com/pods-framework/pods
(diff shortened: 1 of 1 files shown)


Shelob9:
This shows a deficiency in the version numbering process. I can adapt the grunt task to also fix this meta data as well (woo! more metadata!) But does this mean fully compatibility needs to wait until 2.5.3 @sc0ttkclark  ?

afragen:
I'm not so sure about version numbering, but please also refer to #2939

Shelob9:
I will update the Grunt release process so it sets all the meta data right in the future, see:

See this comment: https://github.com/pods-framework/pods/issues/2931#issuecomment-102529692

Issue #2945, opened 2015-05-15
Local dev environment and symlink'd Pods plugin


afragen:
I use DesktopServer to create local dev environments. All my local git repos are in a different folder. When I'm testing I usually create a symlink to wp-plugins from my local git plugin repo.

In Pods when I do this I have empty Pods Add New, Settings and Help pages.

pglewis:
If we're to support a Pods plugin directory that is symlinked to a location outside `.../wp-content/plugins` then we cannot blindly set `PODS_URL` based on `__FILE__`, because that magic constant will already have resolved the symlink.

pglewis:
An additional whitelist for views might be a good feature as well, but I think that's a separate feature request from this. I think there should be a workable solution to figure out if the pod directory is a symlink and not dereference it when setting PODS_DIR if so.

pglewis:
From the convo on the original pull request: 

`if ( false === strpos( $_real_view, realpath( WP_PLUGIN_DIR ) ) && false === strpos( $_real_view, realpath( WPMU_PLUGIN_DIR ) ) && false === strpos( $_real_view, PODS_DIR ) )` 

is probably the "proper" and easiest fix here.

pglewis:
Still the 3.0 PR and then this one should be closed.

afragen:
The PR seems to have been reversed in 5ff23512dcf759bf56f9f4cca9e160f80b10d44e

afragen:
Though merged PR should be changed to reflect new logic as in current `2.x`

pglewis:
Yeah, my mistaken direct push to 2.x instead of a pull request muddied the waters even more.  I've verified the fix looks correct to me now in both 2.x and 3.0.  Away she goes.

PR #2946, opened 2015-05-15
Fix for symlinkd pods in local dev environment

afragen:
See #2945

### DIFF
--- a/classes/PodsView.php
+++ b/classes/PodsView.php
@@ -605,7 +605,8 @@ class PodsView {
 			// Allow includes within plugins directory too for plugins utilizing this
 			if ( file_exists( $_view ) ) {
 				$located = $_view;
-			} else {
+			}
+			else {
 				$located = apply_filters( 'pods_view_locate_template', $located, $_view );
 			}
 		}
--- a/classes/PodsView.php
+++ b/classes/PodsView.php
@@ -600,12 +600,14 @@ class PodsView {
 				$located = realpath( get_template_directory() . '/' . $_real_view );
 			}
 		}
-		// Allow includes within plugins directory too for plugins utilizing this
-		elseif ( file_exists( $_view ) ) {
-			$located = $_view;
-		}
-		else {
-			$located = apply_filters( 'pods_view_locate_template', $located, $_view );
+
+		if ( ! $located ) {
+			// Allow includes within plugins directory too for plugins utilizing this
+			if ( file_exists( $_view ) ) {
+				$located = $_view;
+			} else {
+				$located = apply_filters( 'pods_view_locate_template', $located, $_view );
+			}
 		}
 
 		return $located;


sc0ttkclark:
Looks good, can you do a PR for `release/3.0` as well?

afragen:
Sure, will do.

afragen:
`release/3.0` PR #2947

pglewis:
This change defeats the enforcement of the view being within the plugin directory of the install or within the theme.  I think we need to fix this at the source: making sure `PODS_DIR` points to wp-content/plugins/... and not the path of the symlink.

afragen:
Really all this does is uncouple an `if elseif else` into some good simpler. I found the `elseif` wasn't working when $view had a valid file path. No idea why it wasn't but it wasn't.

pglewis:
The initial `if` is hit here because the the path of the view is outside `WP_PLUGIN_DIR` and `WPMU_PLUGIN_DIR`, thus why the `elseif`doesn't get hit.  

`$located` is returning `false` from that initial `if` because the view's path is also determined to be outside `get_stylesheet_directory()` and `get_template_directory()`.  

The change defeats this enforcement of acceptable locations for the view and would allow a view in any location the server can read.

pglewis:
There should be comments to make it clear what's happening there, but it's ultimately whitelisting acceptable locations.

sc0ttkclark:
We should update the code here in 2.x and release/3.0 to reflect as @pglewis is suggesting. We should then allow either a filter (array) to add additional whitelisted locations, or a constant/filter (boolean) or something allowing outside locations.

pglewis:
I'm shifting convo for the solution over to #2945

pglewis:
An additional whitelist for views might be a good feature as well, but I think that's a separate feature request from this.  I *think* there should be a workable solution to figure out if the pod directory is a symlink and not dereference it when setting `PODS_DIR` if so.

pglewis:
I'd like to suggest reverting this as it could be a potential security risk.  I'd also like to submit a separate pull request that makes the logic there a little more clear in the future.

afragen:
The following is the logic, isn't it.

```php
$located = false;
 
 		if ( false === strpos( $_real_view, realpath( WP_PLUGIN_DIR ) ) && false === strpos( $_real_view, realpath( WPMU_PLUGIN_DIR ) ) ) {
 			$_real_view = trim( $_real_view, '/' );
 
 			if ( empty( $_real_view ) ) {
 				return false;
 			}
```
Code makes it here with file path.
```php 
 			if ( file_exists( realpath( get_stylesheet_directory() . '/' . $_real_view ) ) ) {
 				$located = realpath( get_stylesheet_directory() . '/' . $_real_view );
 			}
 			elseif ( file_exists( realpath( get_template_directory() . '/' . $_real_view ) ) ) {
 				$located = realpath( get_template_directory() . '/' . $_real_view );
 			}
```
It skips through that last if elseif because it's not in the proper path and should then get the next elseif. since the next elseif only checks to see that the file exists it shouldn't miss this.
```php
 		}
		// Allow includes within plugins directory too for plugins utilizing this
		elseif ( file_exists( $_view ) ) {
			$located = $_view;
		}
		else {
			$located = apply_filters( 'pods_view_locate_template', $located, $_view );
 		}
 
 		return $located;
```

As an abstract

```php
if ( false === strpos( $_real_view, realpath( WP_PLUGIN_DIR ) ) && false === strpos( $_real_view, realpath( WPMU_PLUGIN_DIR ) ) ) {

} else if ( file_exists() ) {

} else {

}

```
As the file exists it should match the `else if`

Please let me know where I'm wrong here.

afragen:
I think it's really just the fact that I'm using it as a symlink locally that uncovered this. As `$located = false` immediately prior to this block the check in the patch keeps the logic consistent.

pglewis:
The view's file can exist but not be *allowed*.  Allowable locations are somewhere in the plugin directory tree or your theme directory.  Because of the symlink dereferencing you are ending up with a file path in an unallowable location.  The logic change would allow a view in any location the server can read.  While it solves your particular issue it opens up a potential security hole by allowing views to be anywhere.

sc0ttkclark:
Maybe we need one more `&&` in that non plugins/mu-plugins check to see if it's also not PODS_DIR, that should suffice here.

pglewis:
Really, that may be all it needs: 

`if ( false === strpos( $_real_view, realpath( WP_PLUGIN_DIR ) ) && false === strpos( $_real_view, realpath( WPMU_PLUGIN_DIR ) ) && false === strpos( $_real_view, PODS_DIR ) )`

To whitelist whatever PODS_DIR is

PR #2957, opened 2015-05-21
Don't over escape icon URL in file fields previewer

Shelob9:
#2995

We were using `esc_url` after handlebars escaped it, so it was always coming out as `http://icon`

Not sure if there are other instances where this could be a problem, but it fixed the issue for me in the CPT editor.

Assigning to @sc0ttkclark to consider the ramifications of this change and worry about any possible side-effects.

### DIFF
--- a/classes/fields/file.php
+++ b/classes/fields/file.php
@@ -498,7 +498,7 @@ class PodsField_File extends PodsField {
             <?php } ?>
 
             <li class="pods-file-col pods-file-icon">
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Consider this part one, part two should be adding an else after line 480 which esc_url on the valid non-empty `$icon`.

Shelob9:
Someone should test this PR with all Pod types, both single select and multiple select before merge. 

someone == me || someone == @NikV I assume.

sc0ttkclark:
Needs a 3.0 version as well

PR #2984, opened 2015-05-29
Fix current user can in 3.0

afragen:
current_user_can not yet defined and causes a PHP fatal for me locally. Moving the check inside the `admin_notices` hook fixes this.

### DIFF
--- a/init.php
+++ b/init.php
@@ -84,14 +84,16 @@ else {
         if ( file_exists( $cmb2_init_path ) ) {
             require_once( $cmb2_init_path );
         } else {
-            if ( is_admin() && current_user_can( 'edit_options' ) ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Code itself needs clean up, it wasn't my original implementation it was copied from some test code.

afragen:
Does it cause PHP fatal for you?

sc0ttkclark:
It only fatals when you don't have the CMB2 files existing, so it doesn't fatal for me.

sc0ttkclark:
I'm going to rewrite the code though, it shouldn't be like that anyways

sc0ttkclark:
(we can't use anonymous functions in Pods because we still support 5.2.x like WP does)

PR #3001, opened 2015-06-01
Fixed #3000 Update post filed in pod instance before saving

### LINKED ISSUE #3000: Metadata (post field) is not flushed after updating (2.4.3, 2.5.2)
I've tested with 2 versions: 2.4.3 and 2.5.2
Whenever you read a meta_data field (1) (get_post_meta) and then update it (update_post_meta) (2). The value in (1) will be stored in cache and we will never have the newly updated value. However, if you do an update first and then get it, it will work as expected. 

Steps to reproduce:
- Create a post field called ```bob``` within a post type
- Create a post under the post type
- Execute this script using the post id (replace $postId val in the script) created in the previous step, you would see two identical values printed in screen. 
```php
<?php
global $wp_did_header;

if (isset($wp_did_header)) {
    throw new Exception("Function called from the wrong place.");
}

// Load wordpress
$wp_did_header = true;
require_once dirname(__FILE__) . '/wp-load.php';
wp();

$postId = 61117;

$bob = get_post_meta($postId, 'bob', true);
echo "$bob <br>";

update_post_meta($postId, 'bob', md5(time()) . " Update 1");

sleep(1);
$bob = get_post_meta($postId, 'bob', true);
echo "$bob <br>";
```
- Repeat previous N number of times, old cached value appears instead of new one
- Execute the following script N times, it will work as expected (at least, different values at different refreshes):
```php
<?php
global $wp_did_header;

if (isset($wp_did_header)) {
    throw new Exception("Function called from the wrong place.");
}

// Load wordpress
$wp_did_header = true;
require_once dirname(__FILE__) . '/wp-load.php';
wp();

$postId = 61117; //replace your postId here

update_post_meta($postId, 'bob', md5(time()) . " Update 1");

$bob = get_post_meta($postId, 'bob', true);
echo "$bob <br>";

$bob = get_post_meta($postId, 'bob', true);
echo "$bob <br>";
```

This is causing particular 2 problems (which are highly related):
- After an update following by a read, in the same request, if someone retrieves meta_data again in the same HTTP request, the returned value even won't be the newly updated value
- Any subsequent reads in subsequent HTTP requests will return old value. 

Here is the timeline when executing update_post_meta. 
```update_post_meta --> PodsMeta::update_meta --> Pods::save --> Pods::fetch```
I believe the root cause is in Pods::fetch function. This function tries to get the latest values of a post from DB and then does a merge with current data with the same instance. However, it does not get the latest field values, as a result, merging keeps old values and they remain in cache. 

```php
$find = get_posts( $args );
if ( !empty( $find ) )
   $this->row = get_object_vars( $find[ 0 ] );
```
```php
if ( !$explicit_set && !empty( $this->row ) && is_array( $this->row ) && !empty( $old_row ) ) {
   $this->row = array_merge( $old_row, $this->row );
}
```

And perhaps, there are two ways to solve this.
1. Do a proper merge in Pods::fetch. This will include all fields from db, it's quite redundant and is a waste
2. in PodsMeta:update_meta, set the field, e.g. ```bob``` to new value before update. This will make sure everything gets merged and cached correctly. Only works with modified field so won't affect anyone else and optimised. 

Thank you for being patient to read my post :)

tuanmh:
- This will make sure that pod instance will hold modified field value
and the modified value go into the cache instead of old one.

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -2707,6 +2707,8 @@ class PodsMeta {
         if ( !is_object( self::$current_field_pod ) || self::$current_field_pod->pod != $object[ 'name' ] )
             self::$current_field_pod = pods( $object[ 'name' ] );
 
+        $pod = self::$current_field_pod;
+
         if ( ( isset( $pod->fields[ $meta_key ] ) || false !== strpos( $meta_key, '.' ) ) && $pod->row !== null) {
 
             $key = $meta_key;
@@ -2722,8 +2724,6 @@ class PodsMeta {
             }
 
         }
-
-        $pod = self::$current_field_pod;
 
         $pod->save( $meta_key, $meta_value, $object_id );
 
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -2707,6 +2707,22 @@ class PodsMeta {
         if ( !is_object( self::$current_field_pod ) || self::$current_field_pod->pod != $object[ 'name' ] )
             self::$current_field_pod = pods( $object[ 'name' ] );
 
+        if ( ( isset( $pod->fields[ $meta_key ] ) || false !== strpos( $meta_key, '.' ) ) && $pod->row !== null) {
+
+            $key = $meta_key;
+            if(false !== strpos( $meta_key, '.' )){
+                $key = current( explode( '.', $meta_key ) );
+            }
+
+            $pod->row[ $meta_key ] = $meta_value;
+
+            if ( isset( $pod->fields[ $key ] ) ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
This needs a Pods 3.0 PR too before we can merge, base it against `release/3.0` and we'll take a look at both.

sc0ttkclark:
3.0 patch: #3002

sc0ttkclark:
Oops, this is based off of the `master` branch, can you rebase off of `2.x` instead? That's where we're doing all development for the upcoming 2.5.3 release.

PR #3003, opened 2015-06-01
2.x Update post filed in pod instance before saving

tuanmh:
This will make sure that pod instance will hold modified field value and the modified value go into the cache instead of old one.


sc0ttkclark:
Related: #3000 
3.0 patch: #3002

Issue #2807, opened 2015-03-22
Pagination & Shortcode issues


quasel:
josh [9:47 PM] 
@datasplit: [pods name="videos" page="2" limit ="2"] That will give you videos 3 and 4
or [pods name="videos" limit="10" offest="2"] will give you 3 through 12

datasplit [9:48 PM] 
page=2" and limit="4" is giving me videos 1 through 4
it isn't giving videos 5 through 8

datasplit [10:02 PM] 
so this [pods name="videos" limit="10" offest="2"] will give you 3 through 12 doesn't work
it would have to be [pods name="videos" limit="10" offest="2" pagination="true"] to work but then it gives you the pages... not ideal for a sidebar unless it was in a slider or something
just as a test, I did pagination="false" and it gave pagination anyway

pglewis [11:13 PM] 
sounds like we need a GitHub issue to look into those params
and tests in the unit test while we're at it
if it reproduces, which I don't see why it wouldn't

Sheriziya:
After talking with @jimtrue about this in the chat, here's my issue with pagination in the shortcode:
I've got this code:
[pods name="gameprojects" orderby="post_title DESC" where="pods_aft_tutorial_example='1'" limit="5" pagination="true" template="Games_overview"]

I've got 8 gameprojects in that particular pod. I only get to see 5 projects and NO pagination.

I've also tried:
pagination="false"
pagination="1"
Limit removal + pagination="false"
Limit removal + pagination="true"

None of these options work.

I have checked to see if the theme's pagination (theme 2014) does produce pagination. This function works perfectly. It's only the pagination inside the shortcode which doesn't work at all for me.

Luckily I don't have that many projects yet, so it's not a problem yet :D

pglewis:
My quick testing shows that setting pagination to anything other than an empty string will turn on pagination.  The pagination will not be displayed if the limit is greater than or equal to the number of items (as in this case of 5).

I would test without your where clause, which should return all 8 records.  If so, there appear to be three records that don't match your where criteria.

pglewis:
As for the original report: I think `pagination` must be set for limit/page to work properly.  If so, I think that's a bug.  Clearly there are use cases for people supplying their own pagination mechanism.

pglewis:
Note that the `pagination` param's behavior has just changed as per #2932.  The `filter_var` behavior from the [PHP manual for FILTER_VALIDATE_BOOLEAN: ](http://php.net/manual/en/filter.filters.validate.php)

[...]

PR #3004, opened 2015-06-01
Fix for pagination handling in shortcodes

sc0ttkclark:
Shortcodes currently use 'pagination' for two contexts (display and data) but if `page` or `offset` is supplied, it's only *meant* for one context (display).


pglewis:
#2807 (2.x branch)

PR #3031, opened 2015-06-19
Avoid PHP Notice: Undefined index 'pick_object'

ctimoteo:
Hello,

i am using your plugin with pods-json-api plugin

i have created one custom taxonomy in pods,

when i try to retrieve json data with pods-json-api i get one PHP Notice of a missing index 'pick_object' 

on file pods/classes/Pods.php on line 791

applied one quick patch to solve it, and i share it here.

Thanks,

good work!

### DIFF
--- a/classes/Pods.php
+++ b/classes/Pods.php
@@ -788,7 +788,7 @@ class Pods implements Iterator {
 		}
 
 		// Simple fields have no other output options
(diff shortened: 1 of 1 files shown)


quasel:
Seems legit to me, but PR needs to go to 2.x Branch and it needs the same fix for 3.0 as well 
btw we are working on a successor of the api based on wp rest api 2 - https://github.com/pods-framework/pods-rest-api any help aprecciated!

sc0ttkclark:
Needs to be rebased on 2.x, needs 3..0 PR

PR #3033, opened 2015-06-21
Add query_field tests (meta_query syntax for where)

sc0ttkclark:
Related to #2582 and #2689

Issue #1662, opened 2013-08-29
WHERE related field has ALL values in list (AND)


brockfanning:
Sorry for asking a blatant support question here!

I am trying to get all pods of post_type "news" that are tagged as BOTH "term_1" and "term_2". I'm able to do ORs, but I can't seem to figure out the AND. I had been trying to do this all in the "where" statement, but with no success. Do I actually need to do some manual JOINs?

Any tips to point me in the right direction?  Thank you!

sc0ttkclark:
Great deal of ideas and information here: #1978

pcfreak30:
pods-framework/pods/pull/3036

sc0ttkclark:
Fixed by #3033

sc0ttkclark:
Following up with the example usage here:

`$pod->find(
    [
        'where' => [
            [
                'field' => 'category.name',
                'value' => [
                    'first',
                    'second',
                ],
                'compare' => 'ALL',
            ],
        ]
    ]
);`

PR #3037, opened 2015-06-23
Add support for compare ALL

pcfreak30:
Allows for compare ALL (needs to match all items) with meta_query syntax.

### DIFF
--- a/classes/PodsData.php
+++ b/classes/PodsData.php
@@ -2467,18 +2467,20 @@ class PodsData {
 		}
 
         // Restrict to supported comparisons
-        if ( !in_array( $field_compare, array( '=', '!=', '>', '>=', '<', '<=', 'LIKE', 'NOT LIKE', 'IN', 'NOT IN', 'BETWEEN', 'NOT BETWEEN', 'EXISTS', 'NOT EXISTS', 'REGEXP', 'NOT REGEXP', 'RLIKE' ) ) )
+        if ( !in_array( $field_compare, array( '=', '!=', '>', '>=', '<', '<=', 'LIKE', 'NOT LIKE', 'IN', 'NOT IN', 'BETWEEN', 'NOT BETWEEN', 'EXISTS', 'NOT EXISTS', 'REGEXP', 'NOT REGEXP', 'RLIKE', 'ALL' ) ) )
             $field_compare = '=';
 
         // Restrict to supported array comparisons
-        if ( is_array( $field_value ) && !in_array( $field_compare, array( 'IN', 'NOT IN', 'BETWEEN', 'NOT BETWEEN' ) ) ) {
+        if ( is_array( $field_value ) && !in_array( $field_compare, array( 'IN', 'NOT IN', 'BETWEEN', 'NOT BETWEEN', 'ALL' ) ) ) {
             if ( in_array( $field_compare, array( '!=', 'NOT LIKE' ) ) )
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Original issue #1662

classes/PodsData.php
```diff
@@ -2501,6 +2503,12 @@ public static function query_field ( $field, $q, $pod = null, &$params = null )
             }
         }
 
+        if ( $field_compare == 'ALL' ) {
+            if ( 1 == count( $field_value ) ) {
+                $field_compare = '=';
```
sc0ttkclark: This is a good choice!

classes/PodsData.php
```diff
@@ -2534,7 +2542,12 @@ public static function query_field ( $field, $q, $pod = null, &$params = null )
             	$field_query = $field_cast . ' ' . $field_compare . ' "' . $field_value . '"';
 			}
 		}
-        elseif ( in_array( $field_compare, array( 'IN', 'NOT IN' ) ) ) {
+        elseif ( in_array( $field_compare, array( 'IN', 'NOT IN', 'ALL' ) ) ) {
+            if ( $field_compare == 'ALL' ) {
+                $field_compare = 'IN';
+                $params->having[] = 'count(DISTINCT ' . $field_cast . ') = ' . count( $field_value );
+                $params->groupby[] = '`t`.`ID`';
+            }
```
sc0ttkclark: Please keep a line between each `if` statement, so it just needs one added below this.

classes/PodsData.php
```diff
@@ -2534,7 +2542,12 @@ public static function query_field ( $field, $q, $pod = null, &$params = null )
             	$field_query = $field_cast . ' ' . $field_compare . ' "' . $field_value . '"';
 			}
 		}
-        elseif ( in_array( $field_compare, array( 'IN', 'NOT IN' ) ) ) {
+        elseif ( in_array( $field_compare, array( 'IN', 'NOT IN', 'ALL' ) ) ) {
+            if ( $field_compare == 'ALL' ) {
+                $field_compare = 'IN';
+                $params->having[] = 'count(DISTINCT ' . $field_cast . ') = ' . count( $field_value );
```
sc0ttkclark: Please uppercase all MySQL functions, so this would be:

`$params->having[] = 'COUNT( DISTINCT ' . $field_cast . ' ) = ' . count( $field_value );`

sc0ttkclark:
Refactored for testing in #3033

sc0ttkclark:
Implemented in #3033 and #3052

sc0ttkclark:
Here's a very basic example of where ALL comes in handy for custom field lookups which get joined, making it hard to do what you want because it's just one join:

```php
$params = array(
	'where' => array(
		array(
			'key' => 'favorite_color',
			'value' => array( 'Red', 'Green', 'Blue' ),
			'compare' => 'ALL'
		)
	)
);

// Get all items that have *ALL* three of these favorite colors:
// Red, Green, and Blue
$pod->find( $params );
```

sc0ttkclark:
This is super useful for relationships, since they get joined by one LEFT JOIN in the final DB query. Because of that join, it's impossible to check that an item has multiple values in that context, as it will only ever have one value -- unless you `GROUP BY` and do a `GROUP_CONCAT` an alias in select, and plug all this into a `'having'` instead of `'where'`.

Issue #3041, opened 2015-06-23
Support WP_Query like orderby for meta fields


Shelob9:
Should be able to do `orderby => 'field_name.meta_value_num'` and have Pods autocast it as a number. Also should work for dates.

pglewis:
Has this been merged and tested?  Is it ready to close, or what is left?

PR #3043, opened 2015-06-23
Add support for autoCAST()ing meta_value orderby for date and numbers

Shelob9:
This isn't ready to get merged yet. Really just for @sc0ttkclark feedback and for testing. Also needs unit tests and 3.0 patch.
#3041

### DIFF
--- a/classes/PodsData.php
+++ b/classes/PodsData.php
@@ -885,7 +885,7 @@ class PodsData {
 	    }
 
         if ( !empty( $params->orderby ) ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Looks good at first review

jimtrue:
@Shelob9 Do we know if this made it into the default logic for Widgets? Because using a numeric field sorts ANSI (ie text 1, 12, 2, 25). Date works, but then date is stored like yyyy-mm-dd hh:mm:ss

Shelob9:
@jimtrue This happens on any Pods query no matter what its origin.

Issue #3055, opened 2015-07-04
Pods Widgets Use PHP4 Constructor


Shelob9:

pglewis:
I think this is ready to close?

PR #3056, opened 2015-07-04
Get rid of PHP4 constructor in widgets

Shelob9:
#3055

### DIFF
--- a/classes/widgets/PodsWidgetField.php
+++ b/classes/widgets/PodsWidgetField.php
@@ -8,7 +8,7 @@ class PodsWidgetField extends WP_Widget {
      * Register the widget
      */
     public function PodsWidgetField () {
-        $this->WP_Widget(
+	    parent::WP_Widget(
             'pods_widget_field',
             'Pods - Field Value',
             array( 'classname' => 'pods_widget_field', 'description' => "Display a single Pod item's field value" ),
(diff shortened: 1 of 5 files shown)


sc0ttkclark:
Shouldn't we also be using __construct in our widgets here? Instead of the php4 constructors?

Our extend would still have to call parent::__construct with the args, but we also need to declare the args exactly as the parent class does in the function / phpdoc.

And is this backwards compatible with previous versions of WP?

Shelob9:
@sc0ttkclark Good point re: backwards compat. Looking at the changeset in core, I think you are right: https://core.trac.wordpress.org/changeset/32990/trunk/src/wp-includes/widgets.php

Shelob9:
@sc0ttkclark I refreshed with a proper `__construct()` the fact that `WP_Widget::__construct()` doesn't exists in 4.2 and earlier shouldn't be an issue right? Don't constructors always exist by magic?

pglewis:
An omitted constructor is effectively the same and an empty constructor.

sc0ttkclark:
If that's the case @pglewis then we should be set. Can someone verify though, just to be sure, that's what actually happens for < 4.3

Shelob9:
Works as expected in WordPress 4.2.

PR #3058, opened 2015-07-05
Add support for autoCAST()ing meta_value orderby for date and number in 3.x

### LINKED ISSUE #3041: Support WP_Query like orderby for meta fields

Shelob9:
Fixes #3041 in 3.x

Is 3.x sibling of PR #3043

### DIFF
--- a/classes/Pods/Data.php
+++ b/classes/Pods/Data.php
@@ -978,6 +978,8 @@ class Pods_Data {
 		}
 
 		if ( ! empty( $params->orderby ) ) {
+			$params->orderby = (array) $params->orderby;
+			
 			if ( 'post_type' == $pod[ 'type' ] && 'meta' == $pod[ 'storage' ] && is_array( $params->orderby ) ) {
 				foreach ( $params->orderby as $i => $orderby ) {
 					if ( strpos( $orderby, '.meta_value_num' ) ) {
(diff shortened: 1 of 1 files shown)


classes/Pods/Data.php
```diff
@@ -978,7 +978,20 @@ public function build( $params ) {
...
+						$params->orderby[ $i ] = 'CAST(' . str_replace( '.meta_value_num', '.meta_value', $orderby ) . ' AS DECIMAL)';
+					} elseif ( strpos( $orderby, '.meta_value_date' ) ) {
+						$params->orderby[ $i ] = 'CAST(' . str_replace( '.meta_value_date', '.meta_value', $orderby  ) . ' AS DATE)';
+					}
+
+				}
+
+			}
+			
 			$params->orderby = (array) $params->orderby;
```
sc0ttkclark: This should go above the orderby handling, as it casts orderby as an array

Shelob9: Hmm thats, how I did it here: https://github.com/pods-framework/pods/commit/f6652895a1274c1305a170d67fec6c4f9593d508#diff-a96f9b1c49edf7ebf160a14d0b7b3f2aR901

$params->orderby is already an array by the time it gets there, at least in 2.x

sc0ttkclark: Right, it was good in 2.x, but in this case $params can't be assumed it's an array yet.

Shelob9: Fixed via 5ed9757

PR #3073, opened 2015-07-23
Code Coverage

Ramoonus:
for future code coverage


sc0ttkclark:
We should only run coverage for the one build combo, instead of all build combos, right? (Latest WP, Latest PHP, Multisite)

Ramoonus:
@sc0ttkclark indeed, it needs some bash if condition

PR #3091, opened 2015-08-07
scrutinizer: drop external code coverage - timeout

Ramoonus:


sc0ttkclark:
`release/3.0` version? or was it not put into 3.0 yet anyway?

Ramoonus:
Could indeed be used for both
I noticed scrutinizer halted on unit testing timeout
Therefore it had to be replaced

Ramoonus:
checked, was not implemented in 3.x yet

Issue #3018, opened 2015-06-09
Each Tag Traversal does not work with Taxonomy - Needs some love.


PodsBot:
Each Tag Traversal does not work with Taxonomy - Needs some love. submitted via Slack by jimtrue

Related to Forum Post:
http://pods.io/forums/topic/template-each-spits-unexpected-results-on-taxonomy-rel/

This is related to SEVERAL support forum posts, but basically:

- [each] tag traversal doesn't work with Taxonomy at all. Obviously listing `{@taxonomy}` does a display array list which is fine, but doesn't give people what they're looking for: links to the taxonomy pages.
- As a workaround, people have attempted to create relationship fields to Custom Taxonomy and traversing that with the [each] tag. That actually returns a matching entry for EVERY taxonomy term for the number of actual selected taxonomy terms.

My suggestion is we treat Taxonomy as a different animal and treat it differently, either give it it's own style of output tag for traversal or make the each tag recognize that it's traversing a taxonomy.

Shelob9:
This is not a template/frontier issue. `Pods::find()` only gets one associated term, no matter how many terms are associated with the post. That's the actual root issue.

jimtrue:
Hence why taxonomy handling in template magic tags needs some love. This is feature/enhancement. It doesn't work the way users expect, so we need to give them something that will.

Shelob9:
It's not a feature/enhancment its a bug in Pod::find() that has nothing to
do with templating. If it worked in Pods::find() then it would work in the
templates

--
JoshPress.net <http://joshpress.net/>
CalderaWP.com <http://calderawp.com/>
@Josh412 <http://twitter.com/Josh412>

jimtrue:
Closed as this is related to open issue #2915 @Shelob9 There was no supporting or related commentary on the other bug so I wasn't aware that was the issue in reference here, I just knew that whatever it was, it didn't get fixed in the last patch set.

pglewis:
I cannot reproduce a bug in `Pods::find()` (#2915) so I'm re-opening this one.

PR #3106, opened 2015-08-19
Fixes for #3018, `[each]` tag traversal does not work with taxonomy (2.x)

pglewis:
2.x branch patch for #3018

Needs review.

### DIFF
--- a/components/Templates/includes/functions-view_template.php
+++ b/components/Templates/includes/functions-view_template.php
@@ -232,47 +232,72 @@ function frontier_do_subtemplate( $atts, $content ) {
 	);
 
 	$entries = $pod->field( $atts[ 'field' ] );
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Sorry, merged before I saw that you should be looking at `$pod->object_fields` and not `$pod->fields` for taxonomy fields. Can you update the branch and submit a new PR? I'll leave the 3.0 one unmerged for the same reason too.

sc0ttkclark:
Let's add a code comment above the `field()` call that calling `field()` for a taxonomy object field will populate `$pod->fields[ $field_name ]` for the object field's data, in this case a taxonomy object field. Without calling `$pod->field( $field_name )`, it would not normally be available in the `$pod->fields` array and is something to not expect to be there in 3.0 as this was unintentional.

PR #3107, opened 2015-08-19
Fixes for #3018, `[each]` tag traversal does not work with taxonomy (3.0)

pglewis:
3.0 patch for #3018


sc0ttkclark:
Needs to look at $pod->object_fields for taxonomy fields.

pglewis:
Per our discussion, I think we leave this as is for now?

sc0ttkclark:
Holding for further review for 3.0 compatibility with lookups for object_fields

PR #3111, opened 2015-08-20
Notes about the workaround for taxonomies (2.x)

pglewis:
#3018

Future selves will thank you.


sc0ttkclark:
You done good past selves, you done good

Issue #3109, opened 2015-08-19
Support for CPT UI 1.0.x in CPT UI migration component


PodsBot:
Support for CPT UI 1.0.x in CPT UI migration component submitted via Slack by sc0ttkclark

sc0ttkclark:
https://wordpress.org/support/topic/migrate-cpt-ui-not-working

sc0ttkclark:
```
Old: cpt_custom_post_types
New: cptui_post_types

Old option rows in the db haven't been deleted by me, so if you're on 1.0.x+ you should have both at this point.
```

pglewis:
Also:

Old:  `cpt_custom_tax_types`
New `??`

sc0ttkclark:
Yup

PR #3112, opened 2015-08-21
Issue #3109: Support multiple possible option names for post types and taxonomies (2.x)

pglewis:
#3109

### DIFF
--- a/components/Migrate-CPTUI/Migrate-CPTUI.php
+++ b/components/Migrate-CPTUI/Migrate-CPTUI.php
@@ -21,7 +21,29 @@ if ( class_exists( 'Pods_Migrate_CPTUI' ) )
 
 class Pods_Migrate_CPTUI extends PodsComponent {
 
+    /** @var array
+     *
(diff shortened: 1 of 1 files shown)


components/Migrate-CPTUI/Migrate-CPTUI.php
```diff
@@ -32,9 +54,14 @@ class Pods_Migrate_CPTUI extends PodsComponent {
...
      */
-    public function __construct () {
-        $this->post_types = (array) get_option( 'cpt_custom_post_types', array() );
-        $this->taxonomies = (array) get_option( 'cpt_custom_tax_types', array() );
+    public function __construct() {
+
+        $this->post_option_name = $this->get_option_name( $this->post_option_name_list );
+        $this->taxonomy_option_name = $this->get_option_name( $this->taxonomy_option_name_list );
+
+        $this->post_types = (array) get_option( $this->post_option_name, array() );
```
sc0ttkclark: We should check if we found a valid option name before doing this and the next line.

Issue #3105, opened 2015-08-19
Merge Auto Template into core


PodsBot:
Merge Auto Template into core submitted via Slack by josh

sc0ttkclark:
Needs 3.0 PR

PR #3125, opened 2015-08-30
Merge auto template into Templates component 

Shelob9:
2.x PR for #3105

Assigning to @sc0ttkclark for review 

Needs 3.0 patch post-said review

### DIFF
--- a/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
+++ b/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
@@ -99,22 +99,26 @@ class Pods_Templates_Auto_Template_Front_End {
 
 			//start output array empty
 			$auto_pods = array();
-
+			
+			//get pods api class
+			$api = pods_api();
+			
 			//loop through each to see if auto templates is enabled
 			foreach ( $the_pods as $the_pod => $the_pod_label ) {
-				$pods = pods_api( $the_pod );
+				//get this Pods' data.
+				$pod_data = $api->load_pod( array( 'name' => $the_pod ) );
 
 				//if auto template is enabled add info about Pod to array
-				if ( 1 == pods_v( 'pfat_enable', $pods->pod_data[ 'options' ] ) ) {
+				if ( 1 == pods_v( 'pfat_enable', $pod_data[ 'options' ] ) ) {
 					//check if pfat_single and pfat_archive are set
-					$single = pods_v( 'pfat_single', $pods->pod_data[ 'options' ], false, true );
-					$archive = pods_v( 'pfat_archive', $pods->pod_data[ 'options' ], false, true );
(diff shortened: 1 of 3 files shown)


components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -0,0 +1,368 @@
...
+		if ( $auto_pods === false  ) {
+			//get possible pods
+			$the_pods = $this->the_pods();
+
+			//start output array empty
+			$auto_pods = array();
+
+			//loop through each to see if auto templates is enabled
+			foreach ( $the_pods as $the_pod => $the_pod_label ) {
+				$pods = pods_api( $the_pod );
```
sc0ttkclark: @josh This should be set above the foreach as `$api = pods_api();` and inside the foreach on this line it should have `$pod_data = $api->load_pod( array( 'name' => $the_pod ) );`

Further below instead of `$pods->pod_data` you would just use `$pod_data`

Maybe @pglewis can make this tweak for you today so we can get the release out today.

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -0,0 +1,368 @@
...
+	 * @uses 'the_content' filter
+	 *
+	 * @return string Post content with the template appended if appropriate.
+	 *
+	 * @since 2.4.5
+	 */
+	function front( $content ) {
+
+		// cet the current post type
+		$current_post_type = $this->current_post_type();
```
sc0ttkclark: @josh Is current_post_type() right for this? It checks current queried object but that's in WP_Query, that's not the current post data set on the screen.

We're running this on a post content / excerpt filter, we should just pull `get_post_type()`, and support a second parameter (`the_content` default passes in `$post_id` as second parameter in some cases). Also, if we add support for the optional second `$post_id = null` parameter in front(), we should be sure to update our usage of add_filter / remove_filter to include priority: 10 / arguments: 2 otherwise we'll only ever get the `$content` from the filter.

sc0ttkclark: Actually nevermind, I see you're covering taxonomy archives as well.

sc0ttkclark:
No release today, but plan is to have this ready by mid-next week if possible. Only needs that PodsAPI usage fixed and it's all set for 2.x merge and for you to create the `release/3.0` PR.

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
Shelob9: Updated.

sc0ttkclark:
Ready for 3.0 PR too

Issue #3122, opened 2015-08-28
Custom settings page need css fix


namaless:
The current version: http://take.ms/s6giL

Need changes to this:

.pods-manage-field .pods-field-option input[type="text"], .pods-manage-field .pods-field-option select, .pods-manage-field .pods-field-option textarea, .pods-manage-field .pods-field-option .pods-field.pods-boolean, .pods-manage-field .pods-pick-values, .pods-manage-field .pods-field-option .pods-form-ui-field-type-file, .pods-manage-field .pods-slider-field

and remove this 2 settings:

float: left;
margin-left: 22px;

With this modification the options appear ok: http://take.ms/ViHUc

Give this example from wp settings: http://take.ms/B56hQ

nicdford:
I tested this in Chrome/osx 10.10, firefox/osx 10.10, safari/osx 10.10 and wasn't able to recreate. What browser are you using?

namaless:
Chrome: http://take.ms/NmN2X

nicdford:
I tested on chrome/win10 and wasn't able to recreate. Can you make sure there aren't other plugins/code causing the issue? 
![image](https://cloud.githubusercontent.com/assets/7292444/9551388/bae0d7e8-4d65-11e5-9fd0-ee6d3f5dae58.png)

namaless:
I think you not understand, sorry for this, the problem are into custom options page not into metabox.

nicdford:
Ahhh i'm sorry, you're right. I'll make a pull request for this fix.

namaless:
No worries, i put the issue and not a pull request because i don't know where i can fix it, i post the solution without code change, sorry for this.

nicdford:
I made the fix for 2.x in #3124, i'll make another for 3.x release

pglewis:
Fixed in the upcoming 2.7 release

PR #3140, opened 2015-09-15
Removed styles causing issues on settings pages.

nicdford:
Removing float left and margin-left force the field description to
remain under the field instead of awkwardly wrapping. Ran grunt styles. Fix for #3122


sc0ttkclark:
How far back does this work for WP versions?

sc0ttkclark:
I'd like to confirm this on 2.x that the styles still work on previous WP versions, or at which WP version it stops working on. 2.x supports 3.8 or higher, but Pods 3.0 will be last two versions of WP and higher.

PR #3150, opened 2015-09-18
grunt: ignore dockunit.json

Ramoonus:


sc0ttkclark:
Should be added to ignore list for .gitattributes too

Ramoonus:
there it is

Issue #3157, opened 2015-09-23
Auto Template Options only shows first 5 alphabetical Templates in drop-down select


PodsBot:
Auto Template Options only shows first 5 alphabetical Templates in drop-down select submitted via Slack by jimtrue

jimtrue:
Related to this support thread over on the forums:

http://pods.io/forums/topic/switch-from-frontier-to-auto-templates-doesnt-show-all-choices/

If you have more than 5 templates in Pods, Templates, only the first alphabetical 5 are available in the drop-down select list.

jimtrue:
Found it, this just needs `numberposts => -1` added to the get_template_titles() function at:
https://github.com/pods-framework/pods/blob/master/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Settings.php#L354

Postscape:
Hi Jim,

Sorry having trouble getting this fixed. Any way that you could show exactly where to add numberposts => -1 should b in the code for the fix?

Thanks for getting on this so quick :)

jimtrue:
@Postscape you can check this PR to see what line I changed:

https://github.com/pods-framework/pods/pull/3160

But it's the /Pods/Components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Settings.php

Line 354 in the function get_template_titles(), change it to:

`$templates = get_posts( array( 'post_type' => '_pods_template', 'order'=> 'ASC', 'orderby' => 'title', 'numberposts' => -1 ));`

Trying to get this merged so it'll be in http://pods.io/latest/

PR #3158, opened 2015-09-23
Remove Limit of 5 on get_template_titles #3157

jimtrue:


jimtrue:
Jim is a noob. This was supposed to be pulled against the 2.x branch. @sc0ttkclark when you see in, walk me through 'getting rid of it' and I'll do it the correct way. I also need to replicate this against 3.x, I believe, right?

jimtrue:
Closing PR and re-creating against 2.x branch

PR #3160, opened 2015-09-23
Remove Limit of 5 in get_template_titles for #3157

jimtrue:


sc0ttkclark:
I'm down with increasing this, but I'd caution to set some sort of limit, perhaps a high limit of 100? Overridden by a filter if someone *really wants* more.

sc0ttkclark:
Fixes #3157

sc0ttkclark:
`release/3.0` PR - #3166

jimtrue:
The functionality of being able to select your templates from the list was broken as it currently stood. If the template exists, it needs to be in the drop down. This should be modified to use the same logic as the choose template in the Pods Shortcode dialog.

PR #3162, opened 2015-09-23
scrutinizer: dump composer

Ramoonus:
as manual says its already run
https://scrutinizer-ci.com/docs/guides/php/continuous-integration-deployment


sc0ttkclark:
Should be based off of 2.x

Ramoonus:
i really need to adapt my workflow
will issue again

PR #3167, opened 2015-09-24
scrutinizer: remove composer

Ramoonus:
since manual of scrutinizer says its already run


sc0ttkclark:
Needs refresh, will merge after that

Ramoonus:
will do!

PR #3168, opened 2015-09-24
scrutinizer: use phpcs standard wordpress

Ramoonus:


Ramoonus:
followup: remove phpcs from composer

@sc0ttkclark your opinion on this

Ramoonus:
i think this is why the first one fails #3170 / #3171 / #3172

sc0ttkclark:
I lean on your experience, if we don't need / use phpcs, and use Scrutinizer for that, then we can remove.

Ramoonus:
only reason to keep it is that people could run it locally

PR #3173, opened 2015-09-24
dockunit: equal naming

Ramoonus:


sc0ttkclark:
At some point we'll have multiple containers, for different WP versions etc. So the naming here should reflect both PHP version and WP version. I'm not sure how to denote multisite/non-multisite in our configs yet either.

PR #3184, opened 2015-10-03
Add REST Support To Post Types, Taxonomies, Users

Shelob9:
#3182

This patch does the following things:
1) All post type and taxonomy Pods can <em>optionally</em> use core routes, using core controllers. IE if you have a post type Pod called "Scott" then you can optin to having a route at `wp-json/wp/v2/scott" to show all of your Scotts. It works just like posts and pages.

Note: These only show core fields, no custom fields by default. See thing 2.

2) All post type and taxonomy Pods as well as user Pods have the ability to optin to using these routes for reading and writing, on an optin basis for fields. There is a global option and an individual field option.

Some clarifying questions:
1) What about Advanced Content Types?
You should really make your own routes for these. They are custom and need custom design. See question 3 & 4 for some help in that matter.

2) What kind of queries can I do?
The same queries as you can with any core REST API route? Yes, that means you're using WP_Query, WP_User_Query, etc, not Pods class. If you need custom queries, you should make your own routes. See questions 3 & 4 for help in that matter.

3) How Do I Make Custom Routes?
There is excellent documentation here:
http://v2.wp-api.org/extending/adding/

Also I wrote an article about doing so here:
http://torquemag.io/adding-custom-routes-wordpress-rest-api/

4) Can't You Give Me Some Help For My Custom Routes?
Totally, we need to publish the stuff @quasel and I did while trying to make "The One True Pods API" (see question 5) as a Composer package or something to make that easier for you.

5) Hold On, Why Can't We Have Pod Routes?
Good question. Couple reasons I don't like making a generic `/pods/your-pod` type plugin, again:
a) For WordPress objects, the core API handles that really well. For anything else, it's not generic, it's totally custom and trying to make everyone happy with a generic tool is a fool's quest.
b) Doing really complex SQL queries, like Pods class allows via HTTP requests, in an abstract enough way to make everyone happy is not going to happen, especially not securely. Nor is it particularly RESTful. If you need custom collection, make that query -- using the Pods class -- and wrap it up in a custom route.

6) What Do I Need To Make This Work?
As of today, you need the REST API plugin (version 2). Sadly, that's not going to change until probably WordPress 4.5 because this relies on default routes and their controllers which are not proposed for inclusion in WordPress 4.4


Ramoonus:
dockunit tripped at:
Testing on container PHP 7.0 RC-1 + WordPress Nightly

classes/PodsAdmin.php
```diff
@@ -2441,4 +2444,197 @@ function configuration( $pod = null, $full_field_info = false ){
...
+    protected function restable_pod( $pod ) {
+        $type =  $pod[ 'type' ];
+        if( in_array( $type, array(
+                'post_type',
+                'user',
+                'taxonomy'
+            )
+        )
+        ) {
+            return true;
```
sc0ttkclark: Should this be filtered, as when the Pods REST API v2 comes out enabling custom endpoints would find this handy?

classes/PodsInit.php
```diff
@@ -1160,6 +1161,13 @@ public function run () {
         // Show admin bar links
         add_action( 'admin_bar_menu', array( $this, 'admin_bar_links' ), 81 );
 
+        if( function_exists( 'register_api_field' ) ) {
```
sc0ttkclark: Should this happen in `add_rest_support` `if ( is_admin() )`?

sc0ttkclark:
Let's get another dev or two to sign off on this, from a security standpoint in the implementation of REST API stuff, just to be sure. Maybe @quasel can chime in with the "all clear"?

classes/PodsAdmin.php
Shelob9: Not sure I follow you. This check is to see if the Pod could extend a core route. Unrelated to custom routes...

classes/PodsInit.php
Shelob9: This is needed if is admin and if not it admin

classes/PodsAdmin.php
```diff
@@ -2441,4 +2444,197 @@ function configuration( $pod = null, $full_field_info = false ){
...
+     * Build UI for extending REST API, if makes sense to do so.
+     *
+     * @todo maybe disable if default routes don't exist since WP < 4.5 and plugin isn't installed?
+     *
+     * @since 2.5.6
+     *
+     * @access protected
+     */
+    protected function rest_admin() {
+        if( function_exists( 'register_rest_field' ) ) {
```
jamesgol: I couldn't locate a register_rest_field function anywhere.  Is this supposed to be register_api_field?  Maybe this is an old (or potentially new) function?

jamesgol:
I can confirm that all of the security aspects of the API are inherited, so if a user is not allowed to edit posts they are not allowed to run any of the update functions.  Also, verified that specific fields are marked "Read via REST API?" they are viewable and ones that are not marked are not viewable.

I noticed numerous code styling issues that would need to get resolved for sure, "if(" is one that always looks terrible.  I think some of the 

There is also some code duplication, the stuff the restable_pod function does is repeated in a few places.  

I'm not convinced that this couldn't/shouldn't be applied to ACT as well.  I agree that people that want to get really advanced would be better off writing their own endpoints but it would be good to support at least the simplest operations for everything across the board.

I haven't looked at the Pods 3.0 code in forever, so I'm not sure if there is anything in here that deviates from the way things are going to be and if there would be anything that needs to change to make the transition more seamless.

includes/general.php
```diff
@@ -2048,3 +2048,79 @@ function pods_session_start() {
...
+ *
+ * @since 2.5.6
+ *
+ * @param string $field_name The field name.
+ * @param object|Pods $pod Pods object.
+ * @param bool|true $read Are we checking read or write?
+ *
+ * @return bool If supports, true, else false.
+ */
+function pods_rest_api_field_allowed_to_extend( $field_name, $pod, $read = true ) {
```
quasel: maybe rename? it's about if a field can be accessed  for write and or read
i know true / false is more performant but it's harder to understand then 
e.g  $check = write | read  instead of $read = true | false  
but it's harder to read especially line 83/84 @PodsRESTFields

classes/PodsRESTFields.php
```diff
@@ -0,0 +1,129 @@
...
+	 *
+	 * @since 2.5.6
+	 *
+	 * @access protected
+	 */
+	protected function add_fields() {
+		$fields = $this->pod->fields();
+		foreach( $fields as $field_name => $field ) {
+			$read = pods_rest_api_field_allowed_to_extend( $field_name, $this->pod, true );
+			$write = pods_rest_api_field_allowed_to_extend( $field_name, $this->pod, false );
```
quasel: what about combine line 83+84 and just return an $access array with read = true | false and write = true | false
or is there a special reason to separate the check? (including the suggested rename )

quasel: ah got it used in the PodsRestHandlers too hum

classes/PodsRESTHandlers.php
```diff
@@ -0,0 +1,195 @@
...
+					 */
+					$depth = apply_filters( 'pods_rest_api_depth_for_relationship_response', $depth, $field_name, $pod, $related_pod, $id, $request );
+
+					$params = array(
+						'fields' => $fields,
+						'depth' => $depth
+
+					);
+
+					$data = $pod->api->export_pod_item( $related_pod, $params );
```
quasel: with the current implementation this can lead to huge loading times due to the way it's currently implemented iterating over and over the same pods objects especially if someone increases the depth ...

classes/PodsRESTHandlers.php
```diff
@@ -0,0 +1,195 @@
...
+	 * @param string $field_name Name of field
+	 *
+	 * @return bool|int
+	 */
+	public function write_handler( $value, $object, $field_name ) {
+		$pod_name = pods_v( 'type', $object );
+		$id = pods_v( 'id', $object );
+		$pod = self::get_pod( $pod_name, $id );
+		if ( $pod && pods_rest_api_field_allowed_to_extend( $field_name, $pod, false ) ) {
+			$saved_id = $pod->save( $field_name, $value, $id );
```
quasel: $saved_id - not used anywhere ?

classes/PodsRESTHandlers.php
```diff
@@ -0,0 +1,195 @@
...
+	 * Handler for updating custom field data.
+	 *
+	 * @since 2.5.6
+	 *
+	 * @param object $object The object from the response
+	 * @param string $field_name Name of field
+	 *
+	 * @return bool|int
+	 */
+	public function write_handler( $value, $object, $field_name ) {
```
quasel: $value misses phpdoc

quasel:
Sorry didn't have the time to test it did just read the Code :/ will try tomorrow 
Question:
What happens if a user extend's the post with pod's - any interference with WP-API ^^
Why not just add ACT to the endpoints too? just like the CPT 

And if i did read the Code correctly we would Output some things different to the WP-APIv2 e.g. author and all the related stuff is handled differently to wp-api ...

classes/PodsAdmin.php
Shelob9: @jamesgol  +1 Should have been `register_api_field`

Shelob9: Fixed via d56274b

sc0ttkclark:
Looks like one or two actionable things here to do, I'll try and commit changes as needed when I have some time this week.

classes/PodsAdmin.php
sc0ttkclark: Gotcha, nevermind then

classes/PodsInit.php
sc0ttkclark: Adding some tweaks here, commit coming shortly.

includes/general.php
sc0ttkclark: Cleaned up with string 'read' | 'write' checks

classes/PodsRESTHandlers.php
sc0ttkclark: Cleaned up

classes/PodsRESTHandlers.php
sc0ttkclark: Need to further review alternatives for this, right now this is the best we've got, especially since depth is filterable right now

classes/PodsRESTFields.php
sc0ttkclark: Cleaned up usage

sc0ttkclark:
Ready for merge as soon as it passes tests.

sandhya09:
It seems that pods and REST API request is merged.
But I tried wp-json/wp/v2/scott and it is not working?
Do I have to write custom routes?

arikogan:
Hi guys,

I'm very excited about this feature, thanks @Shelob9 for the hard work.

@sandhya09: There are some good instructions on how to set up the plugin [here](http://pods.io/docs/build/extending-core-wordpress-rest-api-routes-with-pods/). I don't think the code is available as a release yet so what I did is to install it in WordPress from a checkout of the 2.x branch of the GitHub project.

I've given this a try. I'm a little bit confused about the naming of the endpoints but that's not the most relevant thing now. I have a pod with type "Post Type (extended)" and I've exposed it with the name "mypod". The API call I make is below.

http://myserver.com/wp-json/wp/v2/mypod/123

where 123 is the id of an existing post. I don't get any response and looking on the logs I can find the following.

`PHP Fatal error:  Call to a member function fields() on a non-object in /home/ubuntu/wordpress/wp-content/plugins/pods/classes/PodsRESTHandlers.php on line 69`

The pod does have a field, it's a Relationship - Simple (custom defined list).

Has anyone run into this? I'd very much appreciate any help on this. Many thanks,

Ariel

jaavid:
@arikogan i have the same problem!

after change "Response Type" on the relational field to "ID Only", the problem fixed.

arikogan:
Thanks @jaavid, will give it a try.

pglewis:
I'm closing this out.  Bugs that turn up should be in their own separate issues.

PR #3192, opened 2015-10-09
Use proper tests checkout for WP_VERSION=latest

danielbachhuber:
This also removes some customizations you've made to the installation script, so you may not want to merge as-is.

### DIFF
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -12,20 +12,37 @@ DB_HOST=${4-localhost}
 WP_VERSION=${5-latest}
 
 WP_TESTS_DIR=${WP_TESTS_DIR-/tmp/wordpress-tests-lib}
-WP_CORE_DIR=/tmp/wordpress/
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Needs further tweaks to merge into 2.x branch, as we have some customizations to install-wp-tests.sh to allow for 'trunk' as an option, versus latest stable.

danielbachhuber:
> as we have some customizations to install-wp-tests.sh to allow for 'trunk' as an option, versus latest stable.

This would be nice to get upstream https://github.com/wp-cli/wp-cli/issues/1860

Ramoonus:
Looking into it

sc0ttkclark:
The one we were working with to test our Term meta code and our own tests against was this:

https://github.com/pods-framework/pods/blob/feature/3163-support-term-meta/bin/install-wp-tests.sh

sc0ttkclark:
Looks like we've got this manually merged with what we have for bleeding/bleeding-maintenance, will follow up when we improve the install-wp-tests.sh as right now it's still somewhat limited and 'unclean' for pushing it upstream.

PR #3231, opened 2015-11-11
Pods 2.6 release

sc0ttkclark:


Ramoonus:
Travis runs twice

sc0ttkclark:
@jimtrue Note for your future self doing changelog notes, new REST API integration still requires the WP REST API plugin to be installed, as register_api_field is not in WP core yet, it won't be added until WP 4.5+, though WP 4.4 has a lot of the base REST API code already.

PR #3267, opened 2015-12-10
Fixing deprecated Polylang method for versions 1.2 and above

talentedunicorn:
When using Polylang > 1.2 
__Notice:__ get_language was called incorrectly in /var/www/html/wp-content/plugins/pods/classes/PodsAPI.php on line 7208: the call to $polylang->get_language() has been deprecated in Polylang 1.2, use $polylang->model->get_language() instead.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -7205,8 +7205,8 @@ class PodsAPI {
             $current_language = pods_sanitize( pll_current_language( 'slug' ) );
 
             if ( !empty( $current_language ) ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
This should already be patched in the `2.x` branch which is where Pods 2.6 currently is in beta right now: http://pods.io/latest/

talentedunicorn:
Thanks you guys are the best :+1:

sc0ttkclark:
You're the best, I mean look at your username and your avatar! YOU ARE THE BEST :)

quasel:
love all around (hug)

Issue #3270, opened 2015-12-11
Check for CSS issues with WordPress 4.4


PodsBot:
"Check for CSS issues with WordPress 4.4" submitted via Slack by pglewis

pglewis:
![add-new](https://cloud.githubusercontent.com/assets/1973916/11757305/3acab440-a02e-11e5-99d7-bc634c5bd1bb.png)

One example from the edit Pod screen

PR #3277, opened 2015-12-14
Change markup to support css in wp 4.4

nicdford:
WP 4.4 removed h3 as a pre-styled tag. Changing tag to an h2 resolves the issue. #3270


sc0ttkclark:
We use h3 elsewhere probably, worth a look about before we merge.

sc0ttkclark:
Deadline for 2.6 is around 2pm central time, otherwise it'll go into Pods 2.6.1

Issue #3294, opened 2015-12-19
OrderBy Failing on PHP7


PodsBot:
OrderBy Failing on PHP7 submitted via Slack by jimtrue

tomherold:
The following will draw an error with php7:
```
	$forces = pods('forces');
	$params = array(
		'orderby' => 'order_alpha ASC',
		'limit' => 10,
	);
	$forces->find($params);
```

jimtrue:
In looking at the SQL, it's skipping passing the orderby field in its entirety in the PHP7 error.

sc0ttkclark:
Shouldn't this be `'orderby' => 'order_alpha.meta_value'`?

jimtrue:
When we were testing this the other day, it's not required to add .meta_value; this the one we were fighting an edge case around, thinking it was the naming of the field (he had 'order' which was obviously an issue), this seems to be something specific to PHP7 but we'd need to test and verify.

tomherold:
Yes, but it will also issue a sql error.

jimtrue:
@tomherold can you share that SQL error on PHP7 with `order_alpha.meta_value`? Might help to track down the fault domain.

sc0ttkclark:
Why PHP7? Does it work with PHP 5.x?

Hope it's not some weird regex PHP7 bug

tomherold:
Sure, will be back in 20 minutes and add the report to the issue ticket.

tomherold:
Yes, works with php 5.x...

sc0ttkclark:
WAT

jimtrue:
I know, right? You'd think this would've been caught by unit tests, right?

tomherold:
Scott,
It may be related to Regex as I had an issue when I switched to php7.

preg_match_all( '/<h([1-3])(.*)>([^<]+)<\/h[1-3]>/iU', $content, $matches, PREG_SET_ORDER ); 

I had to add the 'U' parameter to make this work again...

tomherold:
```
[...]

PR #3297, opened 2015-12-20
`Pods::find()`: Ditch the reference inside the foreach loop that builds the `orderby` parameters

pglewis:
#3294


pglewis:
This is not the fix you were looking for.

PR #3298, opened 2015-12-21
Feature/3294 php7 preg replace issue

pglewis:
#3294


pglewis:
Made it against the wrong branch.  One of them days.

PR #3299, opened 2015-12-21
#3294: Non-zero array keys here in PHP 7 cause odd behavior so just strip the keys

pglewis:
#3294


pglewis:
For your eyes, @sc0ttkclark.  See any problems with this?

sc0ttkclark:
:+1:

PR #3317, opened 2015-12-28
`field()` will return `false` for relationship fields with no relationships set.  Could resolve #3280

### LINKED ISSUE #3280: Error on extended CPT Relationship Fields queried by REST API
PHP Fatal error: Call to a member function fields() on a non-object in /home/ubuntu/wordpress/wp-content/plugins/pods/classes/PodsRESTHandlers.php on line 69

Reported here: https://github.com/pods-framework/pods/pull/3184#issuecomment-159921920

pglewis:
#3280

### DIFF
--- a/classes/PodsRESTHandlers.php
+++ b/classes/PodsRESTHandlers.php
@@ -59,58 +59,64 @@ class PodsRESTHandlers {
 		$id       = pods_v( 'id', $object );
 		$pod      = self::get_pod( $pod_name, $id );
 		if ( $pod && PodsRESTFields::field_allowed_to_extend( $field_name, $pod, 'read' ) ) {
+
 			$params     = null;
 			$field_data = $pod->fields( $field_name );
 			if ( 'pick' == pods_v( 'type', $field_data ) ) {
+
 				$output_type = pods_v( 'rest_pick_response', $field_data['options'], 'array' );
 				if ( 'array' == $output_type ) {
 
 					$related_pod = $pod->field( $field_name, array( 'output' => 'pod' ) );
-					$fields      = $related_pod->fields();
-					$fields      = array_keys( $fields );
-					if ( isset( $related_pod->pod_data['object_fields'] ) && ! empty( $related_pod->pod_data['object_fields'] ) ) {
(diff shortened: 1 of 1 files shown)


classes/PodsRESTHandlers.php
```diff
@@ -59,58 +59,64 @@ public static function get_handler( $object, $field_name, $request ) {
...
 				$output_type = pods_v( 'rest_pick_response', $field_data['options'], 'array' );
 				if ( 'array' == $output_type ) {
 
 					$related_pod = $pod->field( $field_name, array( 'output' => 'pod' ) );
-					$fields      = $related_pod->fields();
-					$fields      = array_keys( $fields );
-					if ( isset( $related_pod->pod_data['object_fields'] ) && ! empty( $related_pod->pod_data['object_fields'] ) ) {
-						$fields = array_merge( $fields, array_keys( $related_pod->pod_data['object_fields'] ) );
-					}
+					if ( is_a( $related_pod, 'Pods' ) ) {
```
sc0ttkclark: I think it's possible $related_pod here could be an array of 'Pods()' objects? or am I mistaken?

pglewis: Either that or false, if there are no related items afaik.  So this "fix" is just totally bypassing the code inside the `if`, since I don't know of a situation where a pick field is going to return a singular Pod object here.  Needs a re-think, including checking if the related object is actually exposed to the REST API.

Shelob9: @sc0ttkclark To clarify $related_pod is 1 Pods object `$related_pod = $pod->field( $field_name, array( 'output' => 'pod' ) );` I don't see anyway Pods could return an array there.

@pglewis - I don't understand the need for `checking if the related object is actually exposed to the REST API.` IF user didn't want to query form this Pod to related Pod user would not have enabled it in UI.

I think best solution is after `$fields      = $related_pod->fields();` if $fields is false, make it into an  empty array, that way fields can be added to it using or it will return empty `pods_rest_api_fields_for_relationship_response` fitler.

sc0ttkclark:
Closing in favort of #3560

PR #3325, opened 2015-12-30
Fix bug for issue number 3319

JSeifBY:
when adding a new post with a required field type 'File'; if i let the file empty the form is submitted and generate an error 'empty file' but the problem that the post has been added to the list without a file.
bug fix : iterate over pods required 'file' field type test if at least one file is selected, if not prevent submit and alert message showing 'You have same empty required files!'"


sc0ttkclark:
Looks like you created the Pull Request on the wrong branch accidentally, the first branch (on the left) should be '2.x' and the second branch (on the right) should be your branch, from your own repository that you forked to make your changes. Feel free to let us know what that branch URL is for yours and we can help you make the new PR.

PR #3327, opened 2015-12-31
Fix bug for issue number 3319 (Required files client side validation)

JSeifBY:
when adding a new post with a required field type 'File'; if i let the file empty the form is submitted and generate an error 'empty file' but the problem that the post has been added to the list without a file.

bug fix : iterate over pods required 'file' field type test if at least one file is selected, if not prevent submit and alert message showing 'You have same empty required files!'

### DIFF
--- a/ui/js/jquery.pods.js
+++ b/ui/js/jquery.pods.js
@@ -74,6 +74,17 @@
 
                     /* e.preventDefault(); */
 
+                    /**
+                     *   validate required files to be selected in client side
+                     */
+                    $('.postbox .pods-form-ui-field-type-file.pods-validate-required').each(function(){
+                        if($(this).find('.pods-files-list > li').length === 0){
+                            e.preventDefault();
+                            alert('You have same empty required files!');
+                            return false;
+                        }
+                    });
+
                     var postdata = {};
                     var field_data = {};
 


sc0ttkclark:
I like what you're doing here, but we need a couple more tweaks to the text shown and I'd like to make the messaging / alert a little less `alert()`, maybe showing a message at the top of the screen `<div id="message" class="error">....</div>` or something like that, as well as messages and/or red markings on individual fields.

JSeifBY:
Sorry for that response delay,
i have just removed the alert dialog and added an inline error messages. I was thinking about adding notification manager like https://notifyjs.com/ to notify user of occurred errors (not only for required files), for example in long scroll pages we can't find the errors in a short time.

sc0ttkclark:
@JSeifBY can you provide collaborator access to your repository?

JSeifBY:
Done!

ui/js/jquery.pods.js
```diff
@@ -122,6 +120,26 @@
...
+
+						$fieldRow.find( '.pods-validate-error-message' ).remove();
+
+						if( 0 === $fieldEl.find( '.pods-files-list > li' ).length ) {
+							valid_form = false;
+
+							var $fieldName = $fieldRow.find( 'label' ).text();
+
+							// @todo Needs i18n work
+							$fieldRow.append(' <td class="pods-validate-error-message">' + $fieldName.replace('*', '' ) + ' is required.</td>' );
```
sc0ttkclark: @JoryHogeveen Needs i18n JS inclusion here, can you add?

sc0ttkclark:
After some changes, it looks better. We need to update some of the JS here though to make sure the markup is right for 2.7

@pglewis can you check the logic of this out and confirm if markup it's using to add/remove the message is correct? Feel free to suggest changes that would be better overall for per-field validation messages.

ui/js/jquery.pods.js
JoryHogeveen: I can't edit the file, maybe new PR after this one?

sc0ttkclark:
@JSeifBY can you add access for @JoryHogeveen to your repo?

sc0ttkclark:
Also @pglewis

ui/js/jquery.pods.js
sc0ttkclark:
```suggestion
							$fieldRow.append(' <td class="pods-validate-error-message">' + PodsI18n.__( '%s is required.' ).replace( '%s', $fieldName.replace('*', '' ) ) + '</td>' );
```

ui/js/jquery.pods.js
```diff
@@ -122,6 +120,26 @@
...
+						}
+
+						$fieldRow.find( '.pods-validate-error-message' ).remove();
+
+						if( 0 === $fieldEl.find( '.pods-files-list > li' ).length ) {
+							valid_form = false;
+
+							var $fieldName = $fieldRow.find( 'label' ).text();
+
+							// @todo Needs i18n work
```
sc0ttkclark:
```suggestion
```

Issue #3083, opened 2015-07-31
Restrict File Types not working for multiple extensions


Dooza:
Not sure when this "broke" but I have a pod with a file upload field, and using "Other" as Restrict File Types. My allowed extensions are "pdf,zip,jpg", but when I try to add a file from the media library, nothing appears in the custom field.

If I remove the file type restriction it does work.

Dooza:
I tried setting the restrictions to just one file type, "pdf", and it works. It must be something to do with having multiple extensions. Maybe its the way its been escaped?

Ramoonus:
maybe it has problems loading an array like that
did you try:
"pdf", "zip", "jpg"

Dooza:
@Ramoonus 
I have tried that, but the Media Library now doesn't show anything when trying to find a file to attach.

When I use the recommended method: pdf,zip,jpg the list of files appear filtered in the Media Library, but when selected, all that happens is the Media Library closes. The chosen file doesn't appear in the list of attached files.

Ramoonus:
that sounds more like WP or javascript phenomena to me

Dooza:
There aren't any JavaScript errors occurring when this happens, but that doesn't mean it isn't. If I use just "pdf" it works, or if I leave it without restrictions it works, so my money is on it being a bug in Pods.

pglewis:
Is this still an issue?  pdf, zip, and jpg are all whitelisted mime types in `wp_get_mime_types()`, so it seems like this should work with just a normal comma separated list.

pglewis:
See also: #2031

Dooza:
I can confirm this problem still exists.

Ramoonus:
@Dooza in which version?

pglewis:
This'll be like the third time I've dug into this area, if there's a bug here I'll get it squashed.

Dooza:
Pods 2.6 on WordPress 4.4 using PHP 5.6.16

Dooza:
If I use just pdf it works, if I use pdf,jpg,zip it shows the files but when selected doesn't appear on the edit page, and if I use "pdf","jpg","zip" no files files show in the media library at all.

I am happy to help with any test code you come up with to fix this.

pglewis:
Yeah, I think that's one combination I somehow haven't tested yet: multiple whitelisted types.  A lot of the other reports (like #2031) turned out to be due to using a non-whitelisted type (commonly svg and vcf)

Dooza:
[...]

PR #3328, opened 2015-12-31
#3083: Over-escaping strikes again (file upload, restrict file types with more than one mime type)

pglewis:
#3083


pglewis:
`js_esc`ing '&&' converted them to entities

Issue #3336, opened 2016-01-09
ACT with custom table don't work with SQLite Integration


tiptronic:
I’m using with the ‘SQLite Integration’ plugin (http://dogwood.skr.jp/wordpress/sqlite-integration/), which is great, since my sites get much more portable when using SQLite (vs. MySQL).

Unfortunately  ‘Advanced Content Types’ don't work in PODS when using SQLite Integration (otherwise it works great). ACTs are never written to the database (and no table is created), which is a pity, because PODS and SQLite Integration are a VERY POWERFUL combo. The problem seems to be somewhere in the table-creation code...

I know SQLite is not _officially_ supported, but since it works just fine with PODS, it would be a huge benefit, if the problem using ACTs could be fixed.

tiptronic:
If you need some help testing, please let me know... everything else but ACTs seems to work just fine.
I can get it working, when creating a custom table manually, but it is a bit ugly since I have to mess with SQLites auto-incrementing the UniqueKey

tiptronic:
What does that mean? Is 'Enhancement Icebox' the road to death?

sc0ttkclark:
Current goal is for us to close issues and focus on what we prioritize. There are many issues marked as feature / enhancement that are not yet prioritized. For those issues (including this one), we have tagged it as "Needs Votes". We then can look over features and enhancements over time to pick out the ones with the most demand and prioritize that way. It's something I recently saw used on some other large GitHub projects, it seemed to help them prioritize and focus on what's important.

This is now the issue query we use to prioritize upcoming features and enhancements for future releases that are not yet roadmapped:

https://github.com/pods-framework/pods/issues?q=label%3A%22Needs+Votes%22+is%3Aclosed+sort%3Areactions-%2B1-desc

https://github.com/pods-framework/pods/issues?q=label%3A%22Needs+Votes%22+is%3Aclosed+sort%3Acomments-desc

tiptronic:
@sc0ttkclark thx for the clarification - makes sense!

PR #3338, opened 2016-01-11
Add SQLite testing for #3336

sc0ttkclark:
Will adjust code and continue testing to see how many issues there may be.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,9 @@ matrix:
     # MySQL 5.7
     - php: 5.6
       env: DB=mysql-5.7-dmr WP_VERSION=latest WP_MULTISITE=0
+    # SQLite
+    - php: 7.0
+      env: DB=sqlite WP_VERSION=latest WP_MULTISITE=0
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Need to figure out what additional tweaks are needed in order to properly test against SQLite, one of which means we need to pull an additional plugin during the build script:

https://wordpress.org/plugins/sqlite-integration/

Also need to move `db.php` to `/wp-content/` set `DB_FILE` / `DB_DIR` in wp-config.php per installation instructions: https://wordpress.org/plugins/sqlite-integration/installation/

pglewis:
This appears to pass already and we create ACTs in the tests, no?  Might need some digging.

sc0ttkclark:
Yeah I'm betting it's not even using SQLite, still needs that integration stuff implemented in the case of `sqlite` as the DB type (in the build .sh script)

sc0ttkclark:
Closing this, a bit too much effort necessary for it.

PR #3355, opened 2016-01-19
Added Nigerian Naira

webcreativeng:
Added the Nigerian Currency to the list


pglewis:
Sorry I didn't notice this earlier, but you should resubmit.  Base your branch from 2.x and submit your PR against 2.x as well.

PR #3371, opened 2016-01-30
travis: support for code coverage and phpcs

Ramoonus:


Ramoonus:
might have to do a single commit to the scrutinizer settings

Ramoonus:
the build of wonders: https://travis-ci.org/pods-framework/pods/jobs/105858411

Ramoonus:
second attempt
https://travis-ci.org/pods-framework/pods/jobs/105859894

pglewis:
We're stuck between a rock and a hard place: in order to get code coverage stats you need to have xdebug active, if you have xdebug active you really slow down execution.  

I personally don't think we need code coverage for the Travis stuff right now.  I think people get caught up in the metrics of the coverage more than the results, but more importantly we're currently using Travis to gate entry into 2.x and I don't think code coverage is a criterion we need to gate by.  

It's an important stat, and I'd like to work out a system, but for now if we need to peek at coverage stats we can run it manually locally where it can actually complete.  A good portion of our tests are integration tests that require a lot of the world be built and include DB access.  If we branch out with a set of actual unit tests that can be done in isolation, we can enable code coverage on those eventually.

Another option would be to only run code coverage on one instance, either HHVM or PHP7.  If we do include coverage, we certainly don't need it on every instance.  But even that is likely to slow down Travis on PRs and if we get our way there's gonna be a lotta PRs coming.

PR #3411, opened 2016-02-21
waffle.io Badge

waffle-iron:
Merge this to receive a badge indicating the number of issues in the ready column on your waffle.io board at https://waffle.io/pods-framework/pods

This was requested by a real person (user quasel) on waffle.io, we're not trying to spam you.


sc0ttkclark:
We'll want to add this in a better spot next to the other badges instead.

PR #3445, opened 2016-03-11
Update to make iFrames work in modals

nicdford:


pglewis:
Manually copied these changes in

PR #3466, opened 2016-03-27
travis-ci: remove retry

Ramoonus:


sc0ttkclark:
Was there a specific reason for removing travis_retry?

Ramoonus:
if it fails, it likely fails twice

Ramoonus:
I can also say it makes the commands Bash for Windows compatible ...

Issue #3353, opened 2016-01-19
[Fatal error] PodsInit wp 4.4.* multi and pods 2.6


global01:
I’m using **PHP 5.6.14-0+deb8 et Wordpress 4.4.1 in Multi and updated Pods to 2.6 via wp-cli**.
I now have this Fatal error (and blank pages) when I click on one of the links to administrate the multisite section /wp-admin/network/:
`
[error] 40116#0: *488544 FastCGI sent in stderr: "PHP message: PHP Fatal error: Class 'PodsInit' not found in /srv/prod/mgt-prod/wp-content/plugins/pods/includes/general.php on line 1789" while reading response header from upstream, client: xx.xx.xx.xx, server: www.xxxx.com, request: "GET /wp-admin/network/update-core.php HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "www.xxxx.com", referrer: "http://www.xxxx.com/wp-admin/"
`
I removed the plugin and did directly a wget on your latest version and the problem still appears, on my dev as well, I double checked the statics variables wich seems ok and the PodsInit Class obviously exists.

If you can help me,

Thx

pglewis:
Possibly related: #2803

jimtrue:
I sent this one over from the forums because it sounded familiar. Is there a workaround, possibly?

JoryHogeveen:
The workaround is this:

(line 81 - 88 in init.php)
`
	        if ( false !== pods_compatibility_check() ) {
			$pods_form = pods_form();
			$pods_init = pods_init();
		        /*if ( ! is_network_admin() ) {
		        }*/
	        }
`

If i'm not mistaken there is no need to do a is_network_admin() check since the plugin needs to be activated in the whole network in order to be loaded at all when on a network page.

Shall I make a patch?

See also: https://github.com/pods-framework/pods/issues/2803 and https://github.com/pods-framework/pods/issues/825

JoryHogeveen:
Fixed in https://github.com/pods-framework/pods/pull/3665

redpik:
not fixed for me.
Pods 2.7.1
WP 4.9 multisite network

global01:
Yes indeed since the last 2 versions

2017-12-15 14:19 GMT+01:00 Benjamin PONGY <notifications@github.com>:

jimtrue:
[...]

Issue #2803, opened 2015-03-19
Revisit ! is_network_admin() check in PodsInit


Shelob9:
See: https://github.com/pods-framework/pods/issues/825#issuecomment-56620588
See: https://wordpress.org/support/topic/multisite-crash-update-24?replies=15#post-6711603

#825

JoryHogeveen:
I keep having this issue on some networks and I think I found why.

This only happens when Pods is activated in the complete network by default.
I'm not sure why, but Pods gets fully loaded in the network pages with this setting and creates a fatal error because PodsInit is not found (because of "!is_network_admin").

I think the "!is_network_admin" check needs to be done way sooner so Pods doesn't initiate in the network pages because the pods_no_conflict_check still calls the PodsInit class.

Or the other sollution (my current sollution) is to allways load PodsInit...

JoryHogeveen:
I keep getting this error on a network. Somehow the pods_init() is required. For me it's a small issue because at every update I can simply change the init.php file, but just thought I'd let you know!

JoryHogeveen:
Why is this labled as "Backward Compatibility"?

It happens when Pods is activated for the whole network.
PHP versions from 5.3 up untill 7 -> WP versions 3.8 up until 4.5 -> and Pods versions for as long a I use it all produce the same error.

JoryHogeveen:
Fixed in https://github.com/pods-framework/pods/pull/3665

Issue #825, opened 2012-12-04
Pods 2.0.5.1 - Upgrade from 2.0.4.1 - Fatal error: Class 'PodsForm' not found in /pods/classes/PodsAPI.php on line 3539


iamrobreynolds:
Wordpress 3.4.2 Multisite
Pods 2.0.4.1 -> Pods 2.0.5.1

After upgrading Pods, both frontend and backend of WP would display Fatal Error and nothing else on several sites that were using Pods.

Fatal error: Class 'PodsForm' not found in /home/dpn/public_html/wp-content/plugins/pods/classes/PodsAPI.php on line 3539

JoryHogeveen:
I have got the same crash again with the 2.5.1.2 update.

Changed this:
```php
if ( false !== pods_compatibility_check() ) {
	$pods_form = pods_form();

	if ( ! is_network_admin() ) {
		$pods_init = pods_init();
	}

}
```

into this:

```php
if ( false !== pods_compatibility_check() ) {
	$pods_form = pods_form();

	$pods_init = pods_init();
		        /*if ( ! is_network_admin() ) {
		        }*/
}
```

And it was resolved again. Does anyone know why this happens?

Shelob9:
@JoryHogeveen I created a new issue for this #2803

PR #3473, opened 2016-04-11
Fix for fatal error when Pods is activated for the whole network.

JoryHogeveen:
Related:

https://github.com/pods-framework/pods/issues/825
https://github.com/pods-framework/pods/issues/2803
https://github.com/pods-framework/pods/issues/3353

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -1800,6 +1800,9 @@ function pods_no_conflict_check ( $object_type = 'post' ) {
     elseif ( 'term' == $object_type )
         $object_type = 'taxonomy';
 
+    if ( ! class_exists( 'PodsInit' ) ) 
+        pods_init();
+
     if ( !empty( PodsInit::$no_conflict ) && isset( PodsInit::$no_conflict[ $object_type ] ) && !empty( PodsInit::$no_conflict[ $object_type ] ) )
         return true;
 
@@ -1822,6 +1825,9 @@ function pods_no_conflict_on ( $object_type = 'post', $object = null ) {
         $object_type = 'post';
     elseif ( 'term' == $object_type )
         $object_type = 'taxonomy';
+
+    if ( ! class_exists( 'PodsInit' ) ) 
+        pods_init();
 
     if ( !empty( PodsInit::$no_conflict ) && isset( PodsInit::$no_conflict[ $object_type ] ) && !empty( PodsInit::$no_conflict[ $object_type ] ) )
         return true;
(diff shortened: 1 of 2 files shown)


sc0ttkclark:
I'm +1 for this but that was added for some reason that we should definitely do a little more testing on to ensure we're not fixing one bug but re-exposing a previous bug.

JoryHogeveen:
I think this was added because Pods doesn't create any network settings pages so one would guess the pods_init would not be needed there.
But if I'm corrent this makes no sense since if a plugin isn't activated for the whole network then the plugin isn't loaded on the network pages at all.
Please correct me if I'm wrong!

Anyhow, I don't how this patch could create a bug anywhere but I totally agree with testing in more cases.
For what it's worth, this patch didn't create any bugs on my networks! :-)

sc0ttkclark:
I'll just note a few test cases that I seem to recall, when someone is able to (you, one of our other users, or one of our team members) test them, we can merge if we don't encounter any issues.

Prerequisite: Have this patched code in place

* Activating Pods network-wide
* Going to network admin pages (any, including Plugins)
* Upgrading Pods plugin (this one may be tricky, not sure how best to do this until we can verify other issues aren't caused and then push out the code in a release)
* Installing and activating other plugins network-wide

sc0ttkclark:
Oh I remember another case here, if the plugin is activated on the *main site* of the network, but not network-wide, check the network admin pages there too.

JoryHogeveen:
Here is the first commit that introduced the network check:

https://github.com/pods-framework/pods/commit/a55e62f5828036320581f7749823d88eff510af3

It had something to do with content-type caching.
Maybe we can make an other change.

The post_init() is called from pods/includes/general.php on line 1789, maybe this call isn't needed when we're on a network page? If that is so, we could best leave this as is and change that file.

Thoughts?

sc0ttkclark:
I think if we can verify Activation on the main network site (not network-wide), visiting network admin pages works, and visiting network admin pages when Pods is network-activated works -- I'm happy to merge this.

JoryHogeveen:
Please check https://github.com/pods-framework/pods/commit/a55e62f5828036320581f7749823d88eff510af3 first. I'm not sure why this was added since I'm not familiar with content-type caching but I think it will mean more to you :-)

JoryHogeveen:
Hi Scott, did you manage to check https://github.com/pods-framework/pods/commit/a55e62f5828036320581f7749823d88eff510af3 ? Let me know!

sc0ttkclark:
Unfortunately it's been too long to really know, even looking at it now

JoryHogeveen:
Any news on this?

sc0ttkclark:
We just needed more testing to ensure this doesn't break in some situations.

JoryHogeveen:
Ok, for me all testing is done in the last years, I do this change manually after every update of Pods for every multisite I manage for at least 2 years now :-)

No idea what https://github.com/pods-framework/pods/commit/a55e62f5828036320581f7749823d88eff510af3 is supposed to actually fix/change.

PR #3485, opened 2016-04-22
added clear cache command

bemeall:
This time in the right branch :-)

It would be nice to be able to clear the pods cache from wp-cli.

Added a command to the pods api wp-cli command.


sc0ttkclark:
Awesome patch! Who knew we needed this, of course we did!

PR #3490, opened 2016-04-26
Support for the new Polylang versions

JoryHogeveen:
Support for the new Polylang versions
Plus some improvements.

Relative: https://github.com/pods-framework/pods/pull/3223
Also, we need to take a look at: https://github.com/pods-framework/pods/issues/1801

PR #3223, opened 2015-11-05
Update polylang compatibility

JoryHogeveen:
Use $polylang->model to get the current language (since Polylang v1.2) if available.

Issue #1801, opened 2013-10-22
Making Pods compatible with Polylang 1.2


Chouby:
Hi Scott!

I will release Polylang 1.2 during the first half of November. It will introduce a lot of changes in the internal structure. I explain why at http://polylang.wordpress.com/2013/10/01/polylang-1-2-beta-ready-for-tests/

I am aware that Pods interacts with Polylang at a rather low level so I did a grep on the Pods code and here are the results of my investigations.

in classes/PodsAPI.php at line 6652
```php
if ( !empty( $current_language ) ) {
	$current_language_t_id = (int) $polylang->get_language( $current_language )->term_id;
	$current_language_tt_id = (int) $polylang->get_language( $current_language )->term_taxonomy_id;
}
```
should be modified in
```php
if ( !empty( $current_language ) ) {
	$current_language_t_id = (int) $polylang->model->get_language( $current_language )->term_id;
	$current_language_tt_id = (int) $polylang->model->get_language( $current_language )->term_taxonomy_id;
	$current_language_tl_t_id = (int) $polylang->model->get_language( $current_language )->tl_term_id; # to be used when querying terms
	$current_language_tl_tt_id = (int) $polylang->model->get_language( $current_language )->tl_term_taxonomy_id; # to be used when querying terms
}
```
get_language has been moved from $polylang to $polylang->model. The initial code will work though as I introduced a magic method __call to transfer method calls to $polylang to $polylang->model but I guess a direct call is always better (a warning will be issued by Polylang if WP_DEBUG is set to true).

I add 2 new var for use below.

in classes/PodsAPI.php at line 6910
```php
elseif ( is_object( $polylang ) && !empty( $current_language ) && function_exists( 'pll_is_translated_taxonomy' ) && pll_is_translated_taxonomy( $taxonomy ) ) {
	$info[ 'join' ][ 'polylang_languages' ] = "
		LEFT JOIN `{$wpdb->termmeta}` AS `polylang_languages`
			ON `polylang_languages`.`term_id` = `t`.`term_id`
				AND `polylang_languages`.`meta_value` = {$current_language_t_id}
	";

	$info[ 'where' ][ 'polylang_languages' ] = "`polylang_languages`.`term_id` IS NOT NULL";
}
```
should move to
```php
elseif ( is_object( $polylang ) && !empty( $current_language ) && function_exists( 'pll_is_translated_taxonomy' ) && pll_is_translated_taxonomy( $taxonomy ) ) {
[...]

PR #3491, opened 2016-04-26
Support for the new Polylang versions

JoryHogeveen:
Sorry for the previous commit. This time in the right place.

Support for the new Polylang versions
Plus some improvements.

Relative: #3223 https://github.com/pods-framework/pods/pull/3490
Also, we need to take a look at: #1801

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -7316,19 +7316,27 @@ class PodsAPI {
         $current_language_t_id = $current_language_tt_id = 0;
 
         // WPML support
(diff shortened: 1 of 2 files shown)


sc0ttkclark:
That was fast, @pglewis would be proud

sc0ttkclark:
Thanks for your reply on #3490, yeah we need to follow up with #1801 but this is a good first step. If you want to do a PR for anything on #1801 feel free, we'll back you up where we can.

JoryHogeveen:
I believe all Polylang references are ok now: https://github.com/pods-framework/pods/search?utf8=%E2%9C%93&q=polylang

PR #3511, opened 2016-05-09
*PodsRESTHandlers::write_handler needs to be static, fixes #3510

### LINKED ISSUE #3510: PodsRESTHandlers::write_handler needs to be static
It is called statically as a callback but is not defined statically.

pcfreak30:


sc0ttkclark:
Didn't mean to merge this into `master`, but just going to leave it because I don't want to mess around with revert etc

Issue #3242, opened 2015-11-24
Pod Template issue with WPML enabled sites


Tom2527:
I have a Wordpress with both the Pods plugin and WPML plugin as I need the site to be both in English and Dutch. I checked the template option "Make 'Pod Templates' translatable" and I created a Pods Template that I translated in both languages. 

This template is shown AFTER a specific Pod is shown. Everything works fine, except that the English template is shown at all times. Never the Dutch one. 

If I rename the English template to "mytemplate_ENG" only the Dutch template is shown at all times. Both translations are linked to each other using the WPML interface.

<img width="814" alt="screen shot 2015-11-24 at 13 24 17" src="https://cloud.githubusercontent.com/assets/15998590/11366659/f1c8991a-92ae-11e5-9339-25fa790ea053.png">
<img width="1110" alt="screen_shot_2015-11-24_at_13_25_01" src="https://cloud.githubusercontent.com/assets/15998590/11366660/f1edf3d6-92ae-11e5-9bf8-19664a7026af.png">

jimtrue:
This is from this Forum Post: http://pods.io/forums/topic/wpml-and-pods-templates/

It looks like to me that the Auto Template Options isn't recognizing the translation settings/branch coming from the Add/Edit Pod Template Editor. Since a branch is created (using the same name for the Template) with the multiple translations, it's just not picking up that choice in the Auto Template Options based on the language selected. Going to have to push this one up the chain of command for troubleshooting. @quasel or @Ramoonus, do you guys have any experience with WPML and can add any pointers?

jamesgol:
I'm still going to dig into this a bit more to come up with the best solution but here is a simple fix that should not cause any issues (because it uses a filter).  Some of the other translation plugins might have the same issue
In includes/general.php:pods_by_title()
Change:
```
    if ( $page )
        return get_post( $page,  $output );
```
to
```
    if ( $page )
        return get_post( apply_filters( 'wpml_object_id', $page, $type, true ), $output );
```

jimtrue:
@jamesgol the forum poster I referenced for this issue tried your solution and it worked perfectly. Wanted to let you know ;)

JoryHogeveen:
Wasn't this issue fixed/closed with #3537 ?

PR #3526, opened 2016-05-16
Create new 'post_id' method for pods_v

jamesgol:
Related to #3242

This provides a method to allow i18n plugins to return a different post id.  There might need to be other places this filter is called or additional methods might need to be called depending on which
translation plugin is used.

### DIFF
--- a/includes/data.php
+++ b/includes/data.php
@@ -778,6 +778,22 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
 					}
 				}
 				break;
+			case 'post_id':
+				if ( is_array( $var ) ) {
+					if ( isset( $var[ 'post_id' ] ) ) {
+						$post_id = $var[ 'post_id' ];
+					}
+					if ( isset( $var[ 'post_type' ] ) ) {
+						$post_type = $var[ 'post_type' ];
+					} else {
+						$post_type = 'post';
+					}
+				} else {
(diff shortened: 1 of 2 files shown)


includes/data.php
```diff
@@ -778,6 +778,22 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
...
+					if ( isset( $var[ 'post_type' ] ) ) {
+						$post_type = $var[ 'post_type' ];
+					} else {
+						$post_type = 'post';
+					}
+				} else {
+					$post_id   = $var;
+					$post_type = 'post';
+				}
+				$output = apply_filters( 'wpml_object_id', $post_id, $post_type, true );
```
sc0ttkclark: Should this be generalized or is this specifically for WPML support? We should have our own filter here for Pods, and add WPML support just above it if necessary.

sc0ttkclark:
Should be based off of `2.x` branch

includes/data.php
jamesgol: This was specifically for WPML support.  I've not looked into the other translation plugins to see if they will require similar changes.  I was going to add some checks to only call the wpml_object_id filter if WPML is installed but figured since it was a filter there isn't any negative to it

I don't think adding an extra pods specific filter would be a bad thing, just don't know what someone would do with it.  Which is exactly the reason to add the filter :)

Issue #3520, opened 2016-05-12
Fields filter


coding-panda:
Hi Pods team

Could you please add 
_$metabox[ 'args' ][ 'group' ][ 'fields' ] = apply_filters( 'pods_meta_' . __FUNCTION__ . '_fields', $metabox[ 'args' ][ 'group' ][ 'fields' ], $id );_ 
to classes/PodsMeta.php abvoe 
_foreach ( $metabox[ 'args' ][ 'group' ][ 'fields' ] as $field )_ 
in the public function meta_post ( $post, $metabox ); ?

Just so I can filter fields to display for each page.

Thanks a lot.

jimtrue:
@coding-panda Go ahead and submit your change as a pull request against 2.x on a new branch called "feature/3520" and we can test and verify your change. We'll give you credit when we merge it in as well.

Thanks!

jimtrue:
Here's a link to our [Contributing guidelines](https://github.com/pods-framework/pods/blob/master/CONTRIBUTING.md) as well.

coding-panda:
@jimtrue I'm new to github. Would you advice how I can do it? Thanks,

coding-panda:
@jimtrue I have done it. Not sure if I have done it right but it is on the Pull Requests list now. https://github.com/pods-framework/pods/pulls

sc0ttkclark:
Merged into 2.6.6

PR #3532, opened 2016-05-20
feature/3520 - filter the pods metas to display

coding-panda:
Add $metabox[ 'args' ][ 'group' ][ 'fields' ] = apply_filters( 'pods_meta' . FUNCTION . 'fields', $metabox[ 'args' ][ 'group' ][ 'fields' ], $id ); 
to classes/PodsMeta.php abvoe foreach ( $metabox[ 'args' ][ 'group' ][ 'fields' ] as $field )  line:991 so I can filter fields to display for each page.

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -985,6 +985,8 @@ class PodsMeta {
 			self::$current_pod->fetch( $id );
 
         $pod = self::$current_pod;
+	// filter the pods metas to display
+	$metabox[ 'args' ][ 'group' ][ 'fields' ] = apply_filters( 'pods_meta_' . __FUNCTION__ . '_fields', $metabox[ 'args' ][ 'group' ][ 'fields' ], $id );
 
         foreach ( $metabox[ 'args' ][ 'group' ][ 'fields' ] as $field ) {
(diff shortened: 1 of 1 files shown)


classes/PodsMeta.php
```diff
@@ -985,6 +985,8 @@ public function meta_post ( $post, $metabox ) {
 			self::$current_pod->fetch( $id );
 
         $pod = self::$current_pod;
+	// filter the pods metas to display
+	$metabox[ 'args' ][ 'group' ][ 'fields' ] = apply_filters( 'pods_meta_' . __FUNCTION__ . '_fields', $metabox[ 'args' ][ 'group' ][ 'fields' ], $id );
```
sc0ttkclark: Can you rename this to pods_meta_post_fields instead? And be sure to pass in $post, $metabox, $pod too

Also needs phpdoc for filter if possible https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

jimtrue:
To fix #3520 @coding-panda Thank you for submitting this! If you mention the issue in the Comment or Commit block, it will reference it to your issue you're addressing.

Also for future reference, the branch you wanted to create was 'feature/3520' not 'coding-panda' ;)

Issue #3533, opened 2016-05-20
Allow multiple pod forms with the same field names


pcfreak30:
PodsForm::$form_counter needs to be used in the ID and class name possibly to prevent conflict.

PR #3535, opened 2016-05-20
Use form_counter in field name to be unique and prevent conflicts

pcfreak30:
#3533

### DIFF
--- a/classes/PodsForm.php
+++ b/classes/PodsForm.php
@@ -428,7 +428,7 @@ class PodsForm {
             if ( 0 < strlen( pods_v( 'label', $options, '' ) ) )
                 $_attributes[ 'data-label' ] = strip_tags( pods_v( 'label', $options ) );
 
-            $_attributes[ 'id' ] = 'pods-form-ui-' . $name_clean;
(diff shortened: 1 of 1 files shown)


classes/Pods.php
```diff
@@ -3441,7 +3441,7 @@ public function form ( $params = null, $label = null, $thank_you = null ) {
 		$thank_you = $params[ 'thank_you' ];
 		$fields_only = $params[ 'fields_only' ];
 
-		PodsForm::$form_counter++;
+		$form_counter = PodsForm::$form_counter ++;
```
sc0ttkclark: We try to avoid complex lines like these, each line should do one thing preferably. So ++ on one line, the $form_counter set on the next (or just use the PodsForm var down in form.php

sc0ttkclark:
Does this not require other changes to be made during the saving?

pcfreak30:
*Going to redo this and put it in PodsForm so the field name attribute isn't touched, only ID.

pcfreak30:
I have designed this in a way to be backwards compatible and only touch the ID. It only appends the counter if its a secondary+ instance.

PR #3537, opened 2016-05-20
Create new 'post_id' method for pods_v

jamesgol:
Related to #3242 and #3526

This provides a method to allow i18n plugins to return a different post id. There might need to be other places this filter is called or additional methods might need to be called depending on which
translation plugin is used.

### DIFF
--- a/includes/data.php
+++ b/includes/data.php
@@ -778,6 +778,39 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
 					}
 				}
 				break;
+			case 'post_id':
+				if ( is_array( $var ) ) {
+					if ( isset( $var[ 'post_id' ] ) ) {
+						$post_id = $var[ 'post_id' ];
+					}
+					if ( isset( $var[ 'post_type' ] ) ) {
+						$post_type = $var[ 'post_type' ];
+					} else {
+						$post_type = 'post';
+					}
+				} else {
+					$post_id   = $var;
+					$post_type = 'post';
+				}
+				if ( defined( 'ICL_SITEPRESS_VERSION' ) ) {
+					/* Only call filter if WPML is installed */
+					$post_id = apply_filters( 'wpml_object_id', $post_id, $post_type, true );
+				}
+				// Add other translation plugin specific code here
+
+				/**
+				 * Filter to override post_id
+				 *
+				 * Generally used with language translation plugins in order to return the post id of a
+				 * translated post
+				 *
+				 * @param  int $post_id The post ID of current post
+				 * @param  string $post_type Post type of current post
+				 *
+				 * @since 2.6.6
+				 */
+				$output = apply_filters( 'pods_var_post_id', $post_id, $post_type );
+				break;
 			default:
 				$output = apply_filters( 'pods_var_' . $type, $default, $var, $strict, $params );
 		}
--- a/includes/data.php
+++ b/includes/data.php
@@ -798,14 +798,13 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
 				 * translated post
 				 *
 				 * @param  int $post_id The post ID of current post
-				 * @param  mixed $default The default value to set if variable doesn't exist
 				 * @param  mixed $var The variable name, can also be a modifier for specific types
 				 * @param  bool $strict Only allow values (must not be empty)
 				 * @param  array $params Set 'casting'=>true to cast value from $default, 'allowed'=>$allowed to restrict a value to what's allowed
 				 *
 				 * @since 2.6.6
 				 */
-				$output = apply_filters( 'pods_var_post_id', $post_id, $default, $var, $strict, $params );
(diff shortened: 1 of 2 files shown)


includes/data.php
```diff
@@ -778,6 +778,39 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
...
+				 *
+				 * Generally used with language translation plugins in order to return the post id of a
+				 * translated post
+				 *
+				 * @param  int $post_id The post ID of current post
+				 * @param  string $post_type Post type of current post
+				 *
+				 * @since 2.6.6
+				 */
+				$output = apply_filters( 'pods_var_post_id', $post_id, $post_type );
```
sc0ttkclark: Brings up a good point, I wonder if we should filter all `$output` with `$type` like on the line below for defaults, and have that default line just be `$output = $default;`?

jamesgol: Hmm, that is an interesting idea.  You could do some fun things that might or might not have any practical use.  Though $type can be more than just a string (array or object) so there would have to be something like a pseudo type variable that could get set to 'array' or 'object' in those cases.

sc0ttkclark: Only run the filter `if ( is_string( $type ) ) {`, sounds like a fine solution to me.

sc0ttkclark: Actually let's leave things exactly how they are here, but you need to pass in the compatible filter vars just like `pods_var_{$type}`. I don't want to run filters on every pods_v() call because they are EVERYWHERE and that would immediately cause some speed degradation.

jamesgol: Would you want the other vars passed instead of or in addition to what I'm already sending?

I thought of another change to make that might be useful in other cases.  If var is empty it could call get_the_ID() to populate it.  So someone could call pods_v( null, 'post_id' ); and it would return the ID of the current post

pglewis:
My main concern with `pods_v` is that it is used a lot internally and thus will be call a lot within inner data loops.  A mere 1/1000 of a second adds up to a full second if you end up calling it a thousand times.  Need benchmarks.

jamesgol:
@pglewis If we added the filter that @sc0ttkclark was suggesting to be added for every call then that would be a legit issue but were not going to do that.  Even if we did it should be in a completely separate issue/PR since it's mostly unrelated.

includes/data.php
```diff
@@ -778,6 +778,45 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
...
+				 *
+				 * @param  int $post_id The post ID of current post
+				 * @param  string $post_type Post type of current post
+				 * @param  mixed $default The default value to set if variable doesn't exist
+				 * @param  bool $strict Only allow values (must not be empty)
+				 * @param  array $params Set 'casting'=>true to cast value from $default, 'allowed'=>$allowed to restrict a value to what's allowed
+				 *
+				 * @since 2.6.6
+				 */
+				$output = apply_filters( 'pods_var_post_id', $post_id, $post_type, $default, $var, $strict, $params );
```
sc0ttkclark: Just want to ensure we're using standard order of vars here for the filter, should match same order as the one below. Then I think we're finally ready for this!

`apply_filters( 'pods_var_post_id', $post_id, $post_type, $default, $var, $strict, $params );`
into
`apply_filters( 'pods_var_post_id', $post_id, $default, $var, $strict, $params );`

This ensures the first var is filtered and the default/var/strict/params are passed in the same order as below.

And further above -- but after `$post_id` gets set -- we should set `$var = array( 'post_id' => $post_id, 'post_type' => $post_type )` to normalize what `$var` is (and so it contains the `$post_type` without having to add another argument to the filter).

jamesgol: Thinking about it more, do we really even need to pass the $post_type in through the array?  We can just always call get_post_type() on the id and it should be returned.

sc0ttkclark: We don't need to pass it in however we should really normalize `$var` here imo

includes/data.php
```diff
@@ -778,6 +778,39 @@ function pods_v( $var = null, $type = 'get', $default = null, $strict = false, $
...
+				 * translated post
+				 *
+				 * @param  int $post_id The post ID of current post
+				 * @param  mixed $var The variable name, can also be a modifier for specific types
+				 * @param  bool $strict Only allow values (must not be empty)
+				 * @param  array $params Set 'casting'=>true to cast value from $default, 'allowed'=>$allowed to restrict a value to what's allowed
+				 *
+				 * @since 2.6.6
+				 */
+				$output = apply_filters( 'pods_var_post_id', $post_id, $var, $strict, $params );
```
sc0ttkclark: Since we have a value to use, we need to pass in `$default` here still, which I know is a departure from the filter below but the one below had no original value to filter in addition to a default.

sc0ttkclark: `$output = apply_filters( 'pods_var_post_id', $post_id, $default, $var, $strict, $params );`

jamesgol: heh, that is one of the iterations I had before

PR #3544, opened 2016-05-24
feature/3520 - filter the pods metas to display

coding-panda:
Add $metabox[ 'args' ][ 'group' ][ 'fields' ] = apply_filters( 'pods_meta_post_fields', $metabox[ 'args' ][ 'group' ][ 'fields' ], $id,  $post, $metabox, $pod  );
to classes/PodsMeta.php below $pod = self::$current_pod; line:987 so I can filter fields to display for each page.

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -986,6 +986,19 @@ class PodsMeta {
 
         $pod = self::$current_pod;
 
+	/**
+	 * Filter the pods metas to display
+	 *
+	 * @since 2.6.6
+	 *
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
@coding-panda can you get me collaborator access to that repository?

coding-panda:
@sc0ttkclark I has Invited you to be a collaborator. If you haven't received the invitation, please let me know. Thanks.

sc0ttkclark:
In the future, we should add similar filters to other places of output where we have direct access to the current ID/object of an item.

Like `meta_user` etc

Issue #3547, opened 2016-05-25
Add filter to PodsMeta->groups_get


jamesgol:
I wanted to be able to group up fields in the backend and didn't want to have to wait Pods 2.8.

I did a proof of concept example in a test plugin, but have to see about the best way to share the code for other people that want to add something similar.

jamesgol:
Here is a gist of how to make this happen https://gist.github.com/jamesgol/abbf8ee1428236deccd080fadadc9fca

PR #3548, opened 2016-05-25
Add filter to PodsMeta->groups_get()

jamesgol:
Allow adjusting the groups via filter

Related to #3547 

This simple change allows someone to do field grouping on their own before the official code is added.


sc0ttkclark:
Oh my, one of the few cases I didn't already have a filter for something before anyone needed it :)

PR #3553, opened 2016-05-31
Don't add space to currency names that use HTML encoding

### LINKED ISSUE #3498: British pound symbol contains space
Hi
With a currency field. I always have a space after my British pound symbol (£) then the amount

I think this line of code is the problem in the pods plugin:

file: pods/classes/fields/currency.php:310 to line 317

// Currency placement policy
// Single sign currencies: 100$, ¬£100
// Multiple sign currencies: 100 Fr, Kr 100
$currency_gap = ”;

if ( strlen( $currency_sign ) > 1 ) {
$currency_gap = ‘ ‘;
}
If its reading the British pounds symbol as ¬£ this counts as 2 characters therefore the logic is adding a gap.

jamesgol:
Resolves #3498

### DIFF
--- a/classes/fields/currency.php
+++ b/classes/fields/currency.php
@@ -312,7 +312,7 @@ class PodsField_Currency extends PodsField {
 		// Multiple sign currencies: 100 Fr, Kr 100
 		$currency_gap = '';
 
-		if ( strlen( $currency_sign ) > 1 ) {
+		if ( strlen( $currency_sign ) > 1 && 0 !== strpos( $currency_sign, '&' ) ) {
 			$currency_gap = ' ';
 		}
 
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Can merge once tests pass.

classes/fields/currency.php
```diff
@@ -312,7 +312,7 @@ public function display( $value = null, $name = null, $options = null, $pod = nu
 		// Multiple sign currencies: 100 Fr, Kr 100
 		$currency_gap = '';
 
-		if ( strlen( $currency_sign ) > 1 && 0 !== strpos( $currency_sign, '&' ) ) {
+		if ( strlen( $currency_sign ) > 1 && false === strpos( $currency_sign, '&' ) ) {
```
jamesgol: I went with checking for it in the first position because it seemed like the safest best.  I couldn't guarantee that someone might not want to add some crazy currency with a filter like 'dollars & cents' or something else crazy.

sc0ttkclark: I think we're safe, if someone has issues we can always approach that separately when the time comes.

Issue #3416, opened 2016-02-23
Error registering Pods programatically


webmandesign:
Hi guys,

**ISSUE DESCRIPTION:**

I am trying to add a Pod and some custom fields via my theme's `functions.php` file, but it throws me an error in WP dashboard: http://easycaptures.com/fs/uploaded/848/1822267647.png

What I am trying to do specifically is to add a metabox to easily set up custom fields for WordPress Pages. I want this to be automatically added via my theme once a user activates the Pods plugin. 

**HOW TO REPLICATE**

To replicate the issue you can put this into Twenty Sixteen theme `functions.php` file:

```
/**
 * Testing Pods plugin
 */
function themeslug_pods_register_custom_fields() {

    if ( function_exists( 'pods_register_type' ) ) {
        pods_register_type( 'post_type', 'themeslug_page' );
    }

} // themeslug_pods_register_custom_fields

add_action( 'init', 'themeslug_pods_register_custom_fields' );
```

As you can see, I've stripped this to pure `pods_register_type( 'post_type', 'themeslug_page' );`. Without the 3rd parameter and without additional `pods_register_field` to debug. Even this bare minimal code doesn't work and throws the error above.

I get the error on all WP admin pages - I can not use WP admin, basically.

Also I would like to note that I've tested this on WordPress multisite only without any additional plugins enabled.

Could you please have a look at the issue and resolve it?

Thanks in advance!

Regards,

Oliver

quasel:
Did a quick Test on a single install same behaviour!

No idea what's you plan is but as @sc0ttkclark already mentioned you wont have the full set of pods functionality. One workaround could be the usage of the [Pods Migration component ](https://github.com/pods-framework/pods/blob/dde459f59321827049a466a2df4d656688989790/components/Migrate-Packages/Migrate-Packages.php) guess @pcfreak30 could help with some sample code 

Or if it's only to add some Meta Boxes just use [CMB2](https://github.com/WebDevStudios/CMB2)

webmandesign:
Hi,

Indeed, just custom meta boxes is the only thing I need.

I am currently supporting Advanced Custom Fields (ACF) in my theme which works fine.

However, I've found a Pods plugin that looks great and wanted to add a support for that plugin in my theme too. It seems to support creating custom meta boxes, plus you can do a lot more with custom post types with it. That, however can not be done with ACF.

So, the ideal scenario would be: 

- If a theme user don't want to extend the site, just use ACF. 
[...]

PR #3554, opened 2016-05-31
Pass $strict = false to load_pod

jamesgol:
This will keep the "Pod not found" message from being displayed. There might be more to this issue since nothing is actually being saved to the actual Pod.

Related to #3416


sc0ttkclark:
On test pass, we can merge.

Issue #3488, opened 2016-04-24
Cannot save time as 12:00am when using date picker with field type set to time


Pezzab:
As it says on the tin. This is true of the latest update to Pods and the previous version.
If I try to set the time to 12:00am and update the page the form field goes blank.

This works the same if the time is set to 12 or 24 hour format.

Having noticed this on a test site I tried to see if I could repeat it on a live site I use this field on. Here the form not only didn't stay at 12:00am after updating it pulled in another time seemingly at random - the saved time changes each time I try and re save as 12:00am.

The test site as no other plugins installed. The live site has quite a few.

So two variations: field returns to blank each time I save as 12:00am or instead it returns an apparently random time.

Pezzab:
Just an update to this - I forgot to say that despite the field refusing to display "12:00am" after saving that the time is set to "00:00:00" anyway.

robrecord:
Not sure if this is related, but I cannot set an empty value for time - it always shows as 12am - and I can't use template tags to filter it out because [if fime_field] always returns true.

robrecord:
To fix, unless mistaken I believe you need to add '00:00:00' to lines 153, 230, 256, 336 in pods/classes/fields/time.php:

`!in_array( $value, array( '0000-00-00', '0000-00-00 00:00:00', '00:00:00' ) )`

although this doesn't help with the [if time_field] template tags - they always return true.

pglewis:
Closing as a duplicate of #4937, should be resolved in #4959

PR #3555, opened 2016-05-31
Allow midnight (00:00) as valid time

jamesgol:
Related to #3488

If 'Allow empty value' is unchecked and a value is not passed it will default to current time, but it will still accept 00:00:00 as a valid value

### DIFF
--- a/classes/fields/time.php
+++ b/classes/fields/time.php
@@ -150,7 +150,7 @@ class PodsField_Time extends PodsField {
     public function display ( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
         $format = $this->format( $options );
 
-        if ( !empty( $value ) && !in_array( $value, array( '0000-00-00', '0000-00-00 00:00:00', '00:00:00' ) ) ) {
+        if ( !empty( $value ) ) {
             $date = $this->createFromFormat( 'H:i:s', (string) $value );
             $date_local = $this->createFromFormat( $format, (string) $value );
 
--- a/classes/fields/time.php
+++ b/classes/fields/time.php
@@ -150,7 +150,7 @@ class PodsField_Time extends PodsField {
     public function display ( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
         $format = $this->format( $options );
 
-        if ( !empty( $value ) ) {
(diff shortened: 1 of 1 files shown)


jamesgol:
Just tested switching to html5 fields and on the ones marked 'Allow empty value' with it set to 12:00am it reverts to current time.  Will have to look into that sometime later.

classes/fields/time.php
```diff
@@ -150,7 +150,7 @@ public function schema ( $options = null ) {
     public function display ( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
         $format = $this->format( $options );
 
-        if ( !empty( $value ) && !in_array( $value, array( '0000-00-00', '0000-00-00 00:00:00', '00:00:00' ) ) ) {
```
sc0ttkclark: We should just take out `00:00:00` here, right? Leave the 'empty dates'?

jamesgol: I suspect the code was just copied from datetime.php initially and that came along.  It wouldn't hurt to have the empty dates in but they should never be seen by this code

jamesgol:
Not sure if it worked before but can't just pop up the picker and hit done to select midnight.  Have to slide it off a bit and come back.  Going to see if I can figure that out.

sc0ttkclark:
Okay, I'll leave this open until we figure that out (if it's possible).

sc0ttkclark:
Merge if you don't end up getting a solution for that, otherwise leave for review and I'll come back to it.

jamesgol:
Unfortunately there isn't any builtin support for just hitting done and setting it to the time.  Even checked a newer version of the js code and no luck (though we might want to upgrade to a newer version for some extra features, but thats for another issue).

Someone can type in 00:00 or 12:00am in the text box or slide the slider one step and then back to reset it.  If it ends up being an issue we can revisit the issue.

jimtrue:
I think the primary issue is that you can set the field to 'allow empty' and it never does, it always writes midnight. #4171

jimtrue:
I also just tested this on 2.x and yes, empty 'input' is always writing 12:00:00 no matter what and is treating it as 'not empty'.

Issue #3563, opened 2016-06-03
PodsAPI::duplicate_pod_item doesn't work


pcfreak30:
Since you are not excluding unique fields from saving, you are making an alias to ->save().

The following should be excluded
- ID
- post_name
- post_date
- post_date_gmt
- post_modified
- post_modified_gmt
- guid

quasel:
seems solved ;)

PR #3564, opened 2016-06-03
Exclude unique post fields from duplication

pcfreak30:
#3563

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -4201,6 +4201,18 @@ class PodsAPI {
         );
 
         foreach ( $fields as $field ) {
+	        if ( in_array( $field['name'], array(
+		        'ID',
+		        'post_name',
+		        'post_date',
+		        'post_date_gmt',
+		        'post_modified',
+		        'post_modified_gmt',
+		        'guid',
+	        ) ) ) {
+		        continue;
+	        }
+	        
             $value = $pod->field( array( 'name' => $field[ 'name' ], 'output' => 'ids' ) );
 
             if ( !empty( $value ) || ( !is_array( $value ) && 0 < strlen( $value ) ) )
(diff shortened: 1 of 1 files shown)


classes/PodsAPI.php
```diff
@@ -4201,6 +4201,18 @@ public function duplicate_pod_item ( $params ) {
         );
 
         foreach ( $fields as $field ) {
+	        if ( in_array( $field['name'], array(
+		        'ID',
+		        'post_name',
+		        'post_date',
```
sc0ttkclark: I think we can remove post_date, post_date_gmt, post_modified, post_modified_gmt from this list.

sc0ttkclark: Also I think we should probably do some `$pod->pod_data['type']` checks here, as well as generally skipping the field `if ( $pod->pod_data['field_id'] == $field['name'] )`.

pcfreak30:
You need them as I tested and WP doesn't override it if it exists. So the post dates have to be skipped or they are copied direct. As for the other checks, I don't know much about them so you would need to fork my PR to patch it.

sc0ttkclark:
@pcfreak30 can you give me write access to that repo?

pcfreak30:
@sc0ttkclark you have been invited as a collaborator.

jimtrue:
@sc0ttkclark This is still flagged for Review, is this one ready to be merged?

sc0ttkclark:
Cleaned up the logic so we can expand to cover any other cases in which certain fields need to be ignored and added a filter to future proof that further.

*Future: Add coverage for duplicating users, terms, and comments.*

Also cleaned up formatting while I was there since the method was small.

sc0ttkclark:
Just added some more coverage for terms, users, and comments.

sc0ttkclark:
Once this passes tests, we can merge.

pcfreak30:
Just realized you excluded ID from the ignore post_type array. it must be included as it will run an update on wp_insert_post/wp_update_post and not a new item. Found this bugging my code in the dev site.

Issue #3506, opened 2016-05-04
Pods relationship fields with Polylang


JoryHogeveen:
The relationship fields don't work correctly with Polylang.

Currently I have some fixes for the pick.php field but it doesn't work in all cases.
https://github.com/JoryHogeveen/pods/blob/patch-3/classes/fields/pick.php

The problem use case:

- global language filter = **en**
- post language = **nl**  _(or any other language)_
- available languages for relationships = **nl**  _(the possibly related posts are not translated yet)_

You would expect to see the relationship objects for the language "nl" since this post is in that language, but alas..
This returns no results because the global language filter is set to "en" and this filteres the results list for the relationships before we can check for translated relationship objects.

For these fields we'll need to disable the Polylang functions that filter the query results based on a language because we don't want the results from the global language filter but the results of the currently-being-edited-post language.

@Chouby , maybe you have any ideas?

JoryHogeveen:
Related: https://github.com/pods-framework/pods/issues/3448

Chouby:
Hi!

The code to filter the posts on admin side is here: https://github.com/polylang/polylang/blob/master/admin/admin-filters-post.php#L100:L128

You will see that the admin language filter is specifically de-activated to correctly filter media and links when editing a post. If I well understood, this is something similar that you want to do.

davidoster:
I am trying to figure out what is the best way to post an issue.
There is also a bbPress forum topic here: [Pods.io Forum topic](http://pods.io/forums/topic/bi-directional-relationship-not-working-anymore-problem-with-polylang/#post-194520)
There I posted:

I can confirm this also.
I am using WooCommerce and Polylang and when I try to pull a relationship from a product attribute to show on an enhanced user profile returns nothing.

WP 4.5.1
WooCommerce 2.5.5
Pods 2.6.5.2
Polylang 1.9
Polylang Theme Strings 3.0

JoryHogeveen:
Hi @davidoster ,

Could you test my version of pick.php? (make sure you have a backup of the original file)
https://github.com/JoryHogeveen/pods/blob/patch-3/classes/fields/pick.php

Let me know if this solves your issue!

davidoster:
As I mentioned on the forum...

I am afraid this version patch 3 of pick.php doesn’t work for me.
I forgot to mention one more plugin that I am using that is related to WooCommerce & Polylang
Hyyan WooCommerce Polylang Integration 0.25, so the complete list of plugins is,
WP 4.5.1
[...]

Issue #2369, opened 2014-09-26
Pods with required fields and Polylang


jmcbade:
I am using WP 4, Polylang Version 1.5.5, Pods Version 2.4.3

I have done a very basic (no other plugins added) and created a simple Pod (not advanced)

I have two fields that are required. I create a post using the pod. Then for the translation, clicking the “+” for Polylang translation. I get warning message, white box on a gray background, I think it’s from Pods, saying:

<required field> is missing

If I edit the pod and say the first field is no longer required, then the next screen shows the same message but saying the second field is missing.

If no fields are set required, then the translation version of the record appears as expected.

jtraulle:
:+1: following this too

JoryHogeveen:
@jmcbade @jtraulle 
I can't reproduce this issue with the latest release (v2.6.6). Can you please verify the latest release is working for you too?

JoryHogeveen:
Closed due to no reply.

PR #3574, opened 2016-06-09
Polylang fixes and cleanup

### LINKED ISSUE #3572: WPML and Relationship Field issues - Not copying relationships with new translations
WPML and Relationship Field issues - Not copying relationships with new translations submitted via Slack by jimtrue

sc0ttkclark:
Fixes #3572 and #3506 

#### Editing content

- [ ] Relationships show the proper posts (Relationship field: Post type) for the current language of the post or term being edited
- [ ] Relationships show the proper terms (Relationship field: Taxonomy) for the current language of the post or term being edited

#### Displaying content

`[pods]` shortcodes, Pods() objects, and get_post_meta instances should work properly.

- [ ] Proper posts shown (Pod type: Post type) with shortcode or Pods::find() for the current selected language of the site
- [ ] Proper terms shown (Pod type: Taxonomy) with shortcode or Pods::find() for the current selected language of the site
- [ ] Relationships show the proper posts (Relationship field: Post type) for the current language of the post or term being displayed
- [ ] Relationships show the proper terms (Relationship field: Taxonomy) for the current language of the post or term being displayed

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -5266,12 +5266,14 @@ class PodsAPI {
         $current_language = false;
         $load_fields = true;
 
-        // WPML support
-        if ( is_object( $sitepress ) && !$icl_adjust_id_url_filter_off )
-            $current_language = pods_sanitize( ICL_LANGUAGE_CODE );
-        // Polylang support
-        elseif ( function_exists( 'pll_current_language' ) )
-            $current_language = pll_current_language( 'slug' );
+	    // Multilingual support
+	    if ( is_object( $sitepress ) && ! $icl_adjust_id_url_filter_off ) {
+			// WPML support
+			$current_language = pods_sanitize( ICL_LANGUAGE_CODE );
+	    } elseif ( function_exists( 'pll_current_language' ) ) {
+			// Polylang support
+			$current_language = pll_current_language( 'slug' );
+	    }
 
         if ( !is_array( $params ) && !is_object( $params ) )
             $params = array( 'name' => $params, 'table_info' => false, 'fields' => true );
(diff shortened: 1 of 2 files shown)


JoryHogeveen:
Tests to do for Polylang and WPML:
All tests to be done with **one-way relationships** and **two-way** for both taxonomies and post_types:

**translated object <> translated object**
This should only return related objects with the same language

**translated object <> not-translated object**
This should return all available objects regardless of language

**translated object <> not-translated object**
This should return results as if Polylang or WPML was disabled

I've tested the above and all looks good to me exept for one bug:

When a object was previously enabled for translation there is a translation reference stored in the database. In order to fix this we need to check if the current object type is enabled for translation.

JoryHogeveen:
Patch 0d9d73a fixed the bug above.

It also enables setting the preferred languages of a non-translateable object so the user can filter the content to relate to (similar function to menu's with Polylang).

JoryHogeveen:
@sc0ttkclark 
> [pods] shortcodes, Pods() objects, and get_post_meta instances should work properly.

I think you are talking about the front-end of a website:

get_post_meta doesn't filter by language I believe (I actually never knew it did).
What if a user wants to display content of an other language? If Pods manually filters all content from post meta values this is not possible.

sc0ttkclark:
We're not filtering post meta by language, just had a note to test and confirm things are working properly everywhere Pods::find() / shortcode loops are being used

JoryHogeveen:
Did some front-end testing aswell (just to be sure) and no problems found.
Do note that I am not familiar with using the pods shortcodes and object. I all about the get_post_meta :)

sc0ttkclark:
I've got a copy of the latest WPML for anyone who wishes to test.

JoryHogeveen:
Sure! I can do some tests like I did with Polylang

sc0ttkclark:
Looks good to me!

JoryHogeveen:
Related: https://github.com/pods-framework/pods/issues/2369

Issue #3582, opened 2016-06-18
Pods::fetch doesn't properly read cache for settings when no ID is passed


pcfreak30:
PodsUI.php:2038 called fetch without an id number. PodsData.php:1769. the id is null, so it switches to slug, but then it uses a null id. result is 2 different keys holding a stale and good version of data.

PR #3583, opened 2016-06-18
Pods::fetch doesn't properly read cache for settings when no ID is passed

pcfreak30:
#3582


sc0ttkclark:
We don't have extensive slug / ID tests in our unit tests, definitely want to put this one under a bit more scrutiny as it could have unforeseen consequences.

classes/PodsData.php
```diff
@@ -1766,9 +1766,8 @@ public function fetch ( $row = null, $explicit_set = true ) {
             $mode = 'id';
             $id = pods_absint( $row );
 
-            if ( !is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) {
-                $mode = 'slug';
-                $id = $row;
+            if ( ! is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) {
```
sc0ttkclark: I think we need a `null !== $row check here && ( .... )` otherwise we'll lose the ability to do the slug lookup mode.

pcfreak30: @sc0ttkclark I have allowed you to edit the branch.

quasel: any reason this is still open? @pcfreak30 @sc0ttkclark

pcfreak30: never late than never, added a null check and fixed merge block.

classes/PodsData.php
```diff
@@ -1998,12 +1998,11 @@ public function fetch( $row = null, $explicit_set = true ) {
...
 			$id   = pods_absint( $row );
 
-			if ( ! is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) {
+			if ( null !== $row && ( ! is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) ) {
 				if ( $this->id && is_numeric( $this->id ) ) {
 					$id = $this->id;
 				} else {
-					$mode = 'slug';
-					$id   = $row;
+					$id = $this->id;
```
JoryHogeveen: This doesn't make sense. Why the if/else if both results are the same?
I think this should be reverted to the original:
```
$mode = 'slug';
$id = $row;
```

pcfreak30: It has been 3 years since I wrote this, so I cannot tell you other than the patched fixed an issue then and today.

JoryHogeveen: When we'll have to make sure this doesn't break anything else before merging.
Also, the code itself would need to update since this if/else statement doesn't make any sense.

pcfreak30: Anyone else can take over on this then.

classes/PodsData.php
```diff
@@ -1998,12 +1998,11 @@ public function fetch( $row = null, $explicit_set = true ) {
 			$mode = 'id';
 			$id   = pods_absint( $row );
 
-			if ( ! is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) {
+			if ( null !== $row && ( ! is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) ) {
 				if ( $this->id && is_numeric( $this->id ) ) {
```
sc0ttkclark:
```suggestion
				if ( 'settings' === $this->pod_data['type'] || ( $this->id && is_numeric( $this->id ) ) ) {
```

classes/PodsData.php
sc0ttkclark:
```suggestion
					$mode = 'slug';
					$id   = $row;
```

JoryHogeveen:
QA passed if @pcfreak30 confirms his issue is still resolved after the latest commit.

sc0ttkclark (approved):

PR #3589, opened 2016-06-23
Field template for file (media) fields >> Rows (default) or Tiles

JoryHogeveen:
Allows for a gallery view for images (and possibly other files in the future)

tiles template - field template option

tiles template - sortable in both directions

tiles template - tiles template for images


sc0ttkclark:
Going to ship 2.6.6 with this disabled as an option for now, but will target the release in the code already in 2.7 now (and it works great!)

PR #3588, opened 2016-06-23
New Link field with WP link modal

JoryHogeveen:
See: https://github.com/pods-framework/pods/issues/2655

Adds a link field type with options to add a link with a title and
target.
Validation similar to website field.

UI has the option to open the WordPress link modal to search for
existing content and autofill the fields for this field.

pods_field_display() will show a formatted link.

Issue #2655, opened 2015-01-27
Link field work and improvements


JoryHogeveen:
Is it possible to add a Link field to Pods?
Would be a bit similar to a Website field I think.
- URL (possible with the Link feature from wordpress itself so you can select the page to link to.
- Title
- Target

Could be separate metadata in the database or a serialized array.

**Last items for Pods 2.7** #4068 
- [x] What to do if the user changes the field type from Website to Link and vice versa?
See: https://github.com/JoryHogeveen/pods-link-field/blob/master/classes/fields/link.php#L252-L255
- [x] Add an option to set the default target value to _blank
See: JoryHogeveen/pods-link-field@5a07ddd
- [x] @sc0ttkclark Do you want to enable other storage types as well? It currently only stores as a serialized array

Shelob9:
We are not adding any new features to Pods 2.X. Once we finish the cmb2 integration for Pods 3.0 we can address new field types. This would be one be very nice to have, possibly as a specific type of loop field.

JoryHogeveen:
Is there a guide/howto on creating extra field types? I would like to try and start with this field on the 3.0 release.

sc0ttkclark:
If you can start with building the frontend for it and how it would work / look (using the same aesthetics as WP), I can meet you in the middle and help you adjust the current website field to support this field format.

JoryHogeveen:
Hi Scott, I assume you mean wireframes and a HTML version of the a link.php template (ui/fields/link.php)?

sc0ttkclark:
Yes, I'd like to see some frontend markup and styling for how you envision it looking.

JoryHogeveen:
![pods link field front-end](https://cloud.githubusercontent.com/assets/826148/6305103/18f9aba6-b929-11e4-9626-64f11f87bde7.png)

This is how I would see this working. Especially the popup feature would be neat.

HTML version for the file link.php in ui/fields:

```html
<div id="link-options">
    <p class="howto"><?php _e('Enter the destination URL') ?></p>
    <div class="alignleft">
        <label><span><?php _e('URL') ?></span><input id="url-field" type="text" name="href"></label>
    </div>
    <div class="alignleft">
        <label><span><?php _e('Title') ?></span><input id="link-title-field" type="text" name="linktitle"></label>
    </div>
    <div class="link-target">
[...]

PR #3593, opened 2016-06-23
New Link field with WP link modal

JoryHogeveen:
See: https://github.com/pods-framework/pods/issues/2655
New PR for 2.7 (old: https://github.com/pods-framework/pods/pull/3588)

Adds a link field type with options to add a link with a title and
target.
Validation similar to website field.

UI has the option to open the WordPress link modal to search for
existing content and autofill the fields for this field.

pods_field_display() will show a formatted link.

link field

link field-modal

### DIFF
--- a/classes/PodsForm.php
+++ b/classes/PodsForm.php
@@ -1310,6 +1310,7 @@ class PodsForm {
         $field_types = array(
             'text',
             'website',
+            'link',
             'phone',
             'email',
             'password',
@@ -1327,7 +1328,6 @@ class PodsForm {
             'boolean',
             'color',
             'slug',
-            'link',
         );
 
         $field_types = array_merge( $field_types, array_keys( self::$field_types ) );
--- a/classes/PodsForm.php
+++ b/classes/PodsForm.php
@@ -1326,7 +1326,8 @@ class PodsForm {
             'pick',
             'boolean',
             'color',
-            'slug'
+            'slug',
+            'link',
         );
 
         $field_types = array_merge( $field_types, array_keys( self::$field_types ) );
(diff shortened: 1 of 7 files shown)


jimtrue:
No Reviews? Just merged into 2.7?

sc0ttkclark:
I did look into the code prior to merge, are you seeing any issues?

jimtrue:
it saves properly as I can see, but it doesn't display() properly, comment from #2655 

> Never actually tested this one, sorry. Output isn't correct in a Magic Tag (ie we should just be able to go {@link_field} and it do all the necessary wrapping, etc.

>this is what it outputs: `and _blank`

>I think because it's actually built as an array, it output it as a display() list, so url, text, and target. It should actually construct the link.

JoryHogeveen:
The `display()` function returns a HTML formatted link. How was this used?

jimtrue:
In the Magic Tag format in a Pods Template {@link_field} so yep, it should've returned the display as a formatted HTML link.

jimtrue:
This is exactly what was in my Pods Template:

`<div>Link Test: [if link_test]{@link_test}[else]No Link[/if]</div>`

JoryHogeveen:
@sc0ttkclark 
https://github.com/pods-framework/pods/blob/release/2.7/classes/fields/link.php#L185-L187
Maybe there is a problem similar to the date field you've fixed? Diplay handling could be called multiple times when the data is an array.

sc0ttkclark:
Gotcha, yeah we'll need to fix that like we fixed the address field display. Check that branch

JoryHogeveen:
Maybe create a separate PR with those commits? Not sure if that is possible?

sc0ttkclark:
Created in #4233

PR #3609, opened 2016-06-28
Allow extra meta/option boxes to be added in the pod-edit screens

JoryHogeveen:
Handy to add meta/option boxes to the side of the pod edit page

### DIFF
--- a/ui/admin/setup-edit.php
+++ b/ui/admin/setup-edit.php
@@ -875,6 +875,7 @@ if ( isset( $tabs[ 'extra-fields' ] ) ) {
                 </div>
             </div>
             <!-- /#submitdiv -->
+            <?php do_action( 'pods_admin_ui_setup_edit_meta_boxes', $pod, $obj ) ?>
         </div>
     </div>
 </div>
--- a/ui/admin/setup-edit.php
+++ b/ui/admin/setup-edit.php
@@ -135,7 +135,15 @@ foreach ( $field_tab_options[ 'additional-field' ] as $field_type => $field_type
         $no_additional[] = $field_type;
 }
 
-// Make use of the WP core meta box functionality
+/**
+ * Make use of the WP core meta box functionality
+ * 
+ * Currently only context 'side' is available
+ * 
+ * @since 2.7
+ * @see https://codex.wordpress.org/Plugin_API/Action_Reference/add_meta_boxes
+ * @param array $pod The Pod object as an array
+ */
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
I wonder how much effort it would be to support meta box functions for this purpose

JoryHogeveen:
What do you mean? The form is around the whole page so these inputs should be available in the post values right?

JoryHogeveen:
Oh, you mean like regular post pages so we can use `add_meta_box()` for this?

JoryHogeveen:
Works great! :)
In this version only the 'side' context is available because of the tabs (which already have action for adding extra tabs).

JoryHogeveen:
TODO:

- [ ] Allow registration of settings with meta boxes (currently the value's aren't being saved because Pods doesn't know whem I guess)
- [ ] Tests!

JoryHogeveen:
Little but of doc info:

When using the add_meta_box function on Pods edit pages the metabox callback content should be wrapped in a element with the class `pods-submittable-fields` otherwise Pods will not save the data.

```
<div class="pods-submittable-fields">
// Metabox callback content
</div>
```

ui/admin/setup-edit.php
```diff
@@ -134,6 +134,9 @@
     if ( empty( $field_type_fields ) )
         $no_additional[] = $field_type;
 }
+
+// Make use of the WP core meta box functionality
+do_action( 'add_meta_boxes', 'pods_edit', $pod );
```
sc0ttkclark: Needs inline hook docs https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

ui/admin/setup-edit.php
```diff
@@ -875,6 +878,7 @@
                 </div>
             </div>
             <!-- /#submitdiv -->
+            <?php do_meta_boxes( 'pods_edit', 'side', $pod ); ?>
```
sc0ttkclark: Let's wrap this in a `pods-submittable-fields` div just to make things easier

JoryHogeveen: Smart thinking!

JoryHogeveen: Tested and works, see https://github.com/pods-framework/pods/pull/3609/commits/5c2be5fb768a52e426822fcf3c0b1993f1b460a9

JoryHogeveen:
Looks good for merging to me. Anyone?

Issue #1614, opened 2013-08-15
"Hide from UI" can result in empty "More Fields" section on Add/Edit page


kingIZZZY:
Issue:
If all of a pod's fields are marked "Advanced > Hide From UI", the pod's Add/Edit page shows an empty "More Fields" section.

Resolution:
Should probably not show the "More Fields" section at all in this case.

PR #3615, opened 2016-06-30
Do not display metagroup if all fields are set to hidden.

JoryHogeveen:
Fixes: https://github.com/pods-framework/pods/issues/1614

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -920,13 +920,19 @@ class PodsMeta {
                 continue;
 
             $field_found = false;
+            $group_hidden = true;
 
             foreach ( $group[ 'fields' ] as $field ) {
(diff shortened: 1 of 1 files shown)


JoryHogeveen:
Other option is to set the metagroup to hidden with the hidden_meta_boxes filter:
https://developer.wordpress.org/reference/hooks/hidden_meta_boxes/

sc0ttkclark:
Are there other areas we can/should do this?

JoryHogeveen:
Not that I know of. Post types are the only ones that use metaboxes

JoryHogeveen:
Wait with merging.
I would like to check what happens with a checkbox field since they are normally set to false (or deleted) if no post value is found.

JoryHogeveen:
Tested checkboxes aswell, looks like all is working fine!

PR #3616, opened 2016-06-30
Allow quotes in relationship custom simple fields

### LINKED ISSUE #2360: Simple Relationship (list) won't save a value with double quotes

jamesgol:
Fixes #2360
Related to #2361


sc0ttkclark:
Not sure if we have simple rel tests with items that have quotes, we should add them though before moving forward, to ensure moving parts continue to work as expected.

jamesgol:
I'll add some tests here if I get a chance this weekend sometime

sc0ttkclark:
Add the test simple relationship field here in this array:

https://github.com/pods-framework/pods/blob/feature/2360-relationship-quotes/tests/includes/testcase.php#L141

And then add the test value here in the `'data'` array:

https://github.com/pods-framework/pods/blob/feature/2360-relationship-quotes/tests/includes/testcase.php#L324

jamesgol:
Forgot how complicated the pods traversal tests are.  Spent a bunch of time on it today getting nowhere, looks like I'll have to work through and re-learn how everything gets set up in order to implement.

pglewis:
Feel free to take notes and brainstorm on directions we could possibly take to make adding tests easier.  That's a huge issue/set of issues in itself.

jamesgol:
The traversal tests won't easily work with the custom-simple fields, it looks like those tests do as much testing of the testing suite as of the Pods code.

I'm working on just a set of tests that only tests all the custom-simple fields with various parameters.

JoryHogeveen:
Is this still valid since DFV?

jamesgol:
I suspect it is partially valid.  Will have to revisit it tonight/tomorrow.

quasel:
is this still an issue? @jamesgol - been loooong night :P sorry was too tempting ^^

PR #3619, opened 2016-07-02
I18n/2.7 localize js

pglewis:
#3620


classes/PodsI18n.php
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * @package Pods
+ * @since 2.7
+ */
+class PodsI18n {
+
+	/**
+	 * @var PodsI18n
+	 */
+	static $instance = null;
```
sc0ttkclark: `private static`

classes/PodsI18n.php
```diff
@@ -0,0 +1,172 @@
...
+			self::$instance = new PodsI18n();
+		}
+
+		return self::$instance;
+	}
+
+	/**
+	 * Constructur
+	 */
+	public function __construct() {
```
sc0ttkclark: `private function`

classes/PodsI18n.php
```diff
@@ -0,0 +1,172 @@
...
+		 * Add strings to the localization
+		 * Setting the key of your string to the original (non translated) value is mandatory
+		 * Note: Existing keys in this class will overwrite the ones of this filter!
+		 * 
+		 * @since 2.7
+		 * @see create_strings()
+		 * @param array
+		 * @return array format: 'Untranslated string' => 'Translated string with use of WP translate functions'
+		 */
+		$strings_extra = apply_filters( 'pods_localized_strings', array() );
```
sc0ttkclark: Maybe pass `$this`? Not required though

classes/PodsI18n.php
```diff
@@ -0,0 +1,172 @@
...
+
+		$this->strings = array_merge( $strings_extra, $this->strings );
```
sc0ttkclark: Should we allow people to register overrides though?

classes/PodsI18n.php
```diff
@@ -0,0 +1,172 @@
...
+	 * We need to register them as normal string to convert to JS references
+	 * And we need to register the translations to attach to these references, these may not be variables!
+	 * 
+	 * @since 2.7
+	 */
+	public function create_strings() {
+
+		return array(
+
+				'%s is required.' =>
```
sc0ttkclark: This is interesting, I wonder if we should key these though, like 'text_is_required', 'field_is_required', 'add', 'add_new', 'add_new_record' etc

sc0ttkclark: Oh I see now, it's used by the JS, this sounds better then in that case to provide JS a way to do __( .. ) type calls

classes/PodsI18n.php
JoryHogeveen: array_merge should overwrite our keys with those coming from the filter.
I'd say we want the keys added with this class to be forced as defaults.

Issue #657, opened 2012-10-19
New Field Type: Address / Map


sc0ttkclark:
Supports Google Maps JS / Static maps and Geocoding; Support for Bing later on

PR & Task List #3634

# Old:
<details>

## Task List
- [x] Component
- [x] Custom Field Type
- [x] Admin Options for Component
- [x] Admin Options for Fields
- [x] Input styles for Address field group and Map
- [x] Output functions for address and maps
- [ ] UI method for when listed table item list (PodsUI)
- [x] Zip Code field that allows leading zeros (international)

## Additional tasks
- [x] Fix integration of the Google Maps API
- [x] Display field options (backend and frontend)
- [ ] Make sure code is set for easy integration of possible other map types (OpenStreetMap for example)
- [ ] Loads of testing with different configurations and switching between them
- [ ] Test with Pods templates and other functions
</details>

JoryHogeveen:
@sc0ttkclark Should this field be a separate plugin or included within Pods?
I ask this because the rep you've created is for a separate plugin.
https://github.com/pods-framework/pods-address-maps

JoryHogeveen:
Err, since when does commenting result in an assignment removal?

JoryHogeveen:
Got the code from @sc0ttkclark compatible and working with the 2.7 branche so time to look at the features etc!

- [ ] Fix integration of the Google Maps API
- [ ] Display field options (backend and frontend)
- [ ] Make sure code is set for easy integration of possible other map types (OpenStreetMap for example)
- [ ] Loads of testing with different configurations and switching between them

quasel:
Status? Map Field is one of the most asked questions these days ;)

JoryHogeveen:
Doing as much as I can. It's working but can have a lot of improvements. Plus, it's not integrated in DFV structure yet.

quasel:
Don't worry - just came up in a bunch of conversations 👍 - any idea about a timeframe? I know Scott is a busy man! ;)

sc0ttkclark:
Mostly testing I think?

JoryHogeveen:
See tasklist, more than just testing ;)

jimtrue:
@JoryHogeveen This is separate from the Address Field in #3634, correct?

JoryHogeveen:
This is the attached issue. But the todolist is maintained in the PR.

jimtrue:
[...]

PR #3623, opened 2016-07-03
2.7 New Field Type: Address / Map

JoryHogeveen:
Patched from https://github.com/pods-framework/pods-address-maps

https://github.com/pods-framework/pods/issues/657

### DIFF
--- a/classes/fields/addressmap.php
+++ b/classes/fields/addressmap.php
@@ -1,0 +1,431 @@
+<?php
+/**
+ * Class PodsField_AddressMap
+ * @package Pods\Fields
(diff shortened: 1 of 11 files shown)


JoryHogeveen:
@sc0ttkclark @pglewis 
Why not set the regular address fields as default (no map options etc) and add the map options with a component.
This way the address field is a regular field with multiple inputs and a component can extend this field with extra options for Google Maps or even order map providers.

The default address field would be quite similar to the GF address field this way (just regular fields and select boxes)
Let me know your thoughts on this.

Issue #3594, opened 2016-06-23
WooCommerce Orders (shop_order post type) not showing up in Pods 'More Fields'


Fabrunet:
I created a 'Folder' pod, with a field 'Associated order' to display linked WooCommerce orders.

I extended the shop_order post type to add a Relationship field type, related to 'Folder'.

So 'Folders' and 'Orders' are now linked

**Issue: when I edit a 'Folder', I can see the associated orders in the 'Custom fields' panel, but the 'More fields' is blank.**

-------------------------------
<img width="497" alt="screeenshot2" src="https://cloud.githubusercontent.com/assets/6525379/16316447/a97af5d4-3954-11e6-9bcf-1c029e15e6ad.png">
-------------------------------
<img width="553" alt="screenshot3" src="https://cloud.githubusercontent.com/assets/6525379/16316652/8e9a7928-3955-11e6-80e3-4dd2a27e757c.png">
-------------------------------
<img width="645" alt="screenshot" src="https://cloud.githubusercontent.com/assets/6525379/16316448/a98e19f2-3954-11e6-8bea-ee7d64f0453f.png">
-------------------------------
I have the latest versions of Wordpress (4.5.3), Pods (2.6.5.2) and WooCommerce (2.6.1). Local environnement, MAMP.

jimtrue:
I **think** this has to do with Custom Statuses on posts not showing up in relationship fields, since Orders wouldn't exactly be 'published'.

Fabrunet:
@jimtrue you were right. I manually edited the status for 'publish' in the database and the order is now showing up in relationship field.

Is there a simple way to force Pods to consider other post statuses?

ValmyRasp:
Hi,

I have same version and i have also empy field list.

![image](https://cloud.githubusercontent.com/assets/6216822/16317592/25c4c07a-398c-11e6-89bd-3e5d3183666b.png)

![image](https://cloud.githubusercontent.com/assets/6216822/16317648/4ee3d388-398c-11e6-9c86-90bcdf0861ba.png)

jimtrue:
@Fabrunet Nope, there's not. That's why I figured this might be the issue. You confirmed my suspicions, so this will go straight to a bug/compatibility issue for handling. @sc0ttkclark This one sounds familiar (the need for managing multiple post statuses for relationship fields). Wasn't this similar to the #3451 issue?

JoryHogeveen:
This filter should work for this:

[...]

PR #3626, opened 2016-07-05
Allow post_status filter to be set for related post_type objects in the edit field UI

JoryHogeveen:
https://github.com/pods-framework/pods/issues/3594


JoryHogeveen:
@sc0ttkclark Shall I patch this to 2.7 aswell?

PR #3633, opened 2016-07-06
Added bulk action export

sc0ttkclark:
Added bulk action export "Export Pods" to export pods data from manage UI

Reopened from #3603 after branch rename


sc0ttkclark:
Oops, didn't realize they both did the same thing, closing in favor of #3662

PR #3634, opened 2016-07-06
New Field Type: Address / Map

sc0ttkclark:
Patched from https://github.com/pods-framework/pods-address-maps
#657

Bringing over PR from #3623 
## Address field type
- [x] Name field type "address" (type) / "Address" (label)
- [x] Display in frontend with HTML markup
- [x] Custom HTML markup possible with magic tags ( `{{address_tag}}` )
- [x] Option to format in schema.org microdata
- [ ] Extend the province/region options. @JoryHogeveen (related: #4116)
  - [ ] Add an option to set this as a select box instead of a text input. (currently only supports US states)
  - [ ] Add a callback function that looks for a list file with existing provinces/region options by country (parameter) + add a filter to change/extend this list.
  - [ ] Add a small AJAX script that autofills the available options based on the above callback.

## Maps component
- [x] "Maps" component name/label
- [x] Hook into field type options to add Maps-specific options (PR #3711 is required)

**Backend UI** 
- [x] Base (currently loads the map with the correct address
- [x] Hidden lat/lng fields when `address` or `free-form` display is selected
- [x] Store lat/lng values after call to geocoder (even if these fields are not visible) to decrease number of api calls
- [x] Determine where API call should and shouldn't be done

**Frontend UI**
- [x] Display address field as a map on front end with use of the providers
- [x] I'd say it's important that this plugin returns an array instead of the display value when using `get_post_meta`, possible workaround for this?
- [x] Fix: Array storage triggers `display()` multiple times
- [ ] Option to set width/height @JoryHogeveen
- [ ] Option to use a Pods Template for the infowindows / marker-tooltip. @JoryHogeveen

**API**
- [ ] API methods for sorting objects based on GEO - [wp ticket](https://wordpress.org/support/topic/field-type-google-map-location/#post-10608913)
- [ ] Map display options for object relationships or a list of objects @JoryHogeveen
  - [x] Make base front-template work with multiple values (basics) @JoryHogeveen
  - [ ] Option to order by distance (straight line can be done without API) - [example](https://stackoverflow.com/questions/10053358/measuring-the-distance-between-two-coordinates-in-php)

#### Google Maps API handler (priority)
- [x] Default geocode calls
- [ ] Check "Client ID" feature (API for business) https://developers.google.com/maps/premium/overview#client-id

**Display**
- [x] Load the map in backend
- [x] Show markers
- [x] Update fields when moving the marker
- [x] Update GEO data when changing the field values (freeform & address input type)
- [x] Show infowindows
- [x] Display address values or custom input data in infowindow
- [x] Custom markers (field type `file` not working in the field edit tab @sc0ttkclark )
- [ ] Map style (interactive map / static image)
- [x] Map type (terrain, satellite, etc)
- [x] Front-end display (based on back-end field display, but then with limited functions etc)
- [x] Option to disable scrolling @JoryHogeveen

# General
- [x] Determine storage structure @sc0ttkclark 
- [x] Determine validation @sc0ttkclark 
- [ ] Fix todo's @sc0ttkclark 
  - fields/address.php
    - [ ] L202 - Storage types
    - [ ] L286 - Validate each returned value for variable type and content (sanitizing)
    - [ ] L414 - check pregreplace, maybe this can be done better (nl2br not working)
  - Maps.php (component)
    - [ ] L87 (assets) - Allways load required front end assets: Maybe as an option?
    - [ ] L217 - Put part of the general maps config in the maps provider class? This could be different between providers.
    - [ ] L311 - Put maps field options in the maps provider class? This could be different between providers.
    - [ ] L329 - Autocorrect is not implemented yet (maybe old?)
    - [ ] L409 - 'display_type' is only available for field type 'address'
    - [ ] L535 (validation) - Validate based on address type ( lat / lon, address fields)
  - [ ] front/map-google.php + field/map-google.php - Check field type: support other field types for maps?
- [x] Delete old/unused files
- [ ] Tests!

# Future tasks

**Structure**
- [ ] MV-field structure (+ maps logic in a JS object) @pglewis 
- [ ] Global maps JS file instead of inline within the field/front files

**Other API integrations**
- [ ] Bing Maps API handler (needs research + contributor!)
- [ ] OpenStreetMap API handler (needs research + contributor!)
- [ ] Leaflet API handler (needs research + contributor!) http://leafletjs.com/index.html


JoryHogeveen:
This field is not ready for full review yet, maybe remove that label.

I'll put the same question here for clarity:

@sc0ttkclark @pglewis 
Why not set the regular address fields as default (no map options etc) and add the map options with a component.
This way the address field is a regular field with multiple inputs and a component can extend this field with extra options for Google Maps or even order map providers.

The default address field would be quite similar to the GF address field this way (just regular fields and select boxes)
Let me know your thoughts on this.

sc0ttkclark:
Still needs review of PR and that question

JoryHogeveen:
**Note:**
Old files that are to be deleted since they are now (or will be) part of the Maps component
- classes / fields / addressmap.php
- ui / css / pods-address-maps.css
- ui / fields / lat-long.php
- ui / fields / map.php
- ui / front / address-map.php
- ui / front / address.php (is in fields)
- ui / front / lat-long.php
- ui / front / maps.php
- ui / js / pods-address-maps.js

classes/fields/address.php
```diff
@@ -0,0 +1,508 @@
...
+				'depends-on' => array(
+					self::$type . '_address_region_input' => 'pick',
+					self::$type . '_address_region' => true,
+					self::$type . '_type' => 'address'
+				),
+				'default'    => 'long',
+				'type'       => 'pick',
+				'data'       => array(
+					'long' => __( 'Full name', 'pods' ),
+					'short' => __( 'Stage / Province code', 'pods' )
```
JoryHogeveen: Typo: Stage -> State

JoryHogeveen:
Found an issue that needs to be addressed:

When a field returns a array from serialised stored, in this case it gets looped through in PodsForm::display() and causes the front-end logic to break the value.

@sc0ttkclark I see there is a $tableless_field_types check for this. Can you give me some info on what you exactly consider a "tableless field type"?

JoryHogeveen:
@sc0ttkclark 
About the data storage. Currently it's all stored serialized under one metakey. Do you want to support multiple metakeys for the different data as well?

sc0ttkclark:
@JoryHogeveen Just fixed the conflicts, can you come back around on the tasklist in this PR and update it with whatever's not specifically called out? If there are `@todo` markers that we need to take care of, can you add them as tasklist items here?

JoryHogeveen:
@sc0ttkclark 
Todolist stayed the same. I've added the todo's I saw in the code to the list.
I think the main topic for now is a structure discussion, both in PHP and JS it's important that the component is flexible to work with for other map providers.

jimtrue:
@JoryHogeveen is this going to include the GeoMashup (Archive of all fields with the 'map' field or related items, etc. being displayed on an overview map). I saw you mention it in #3982 but didn't see it linked through

JoryHogeveen:
@jimtrue That is one of the todolist items yes.

Jen340922:
Has this been released yet?

tiptronic:
Is there any progress or info on this?

bwdgraphics:
Hi. I'm looking for help importing from a CSV into this PODs "address" field. I see the fields within the field for the data I want to import (Addr 1, Addr 2, City, State, Zip) but not sure how to map the import to put the data into those fields. Have tried a few ways, but no luck so far. How would I go about doing this?

jimtrue:
@bwdgraphics I responded over on the WordPress.org forums; what are you using to perform the import with?

bwdgraphics:
Yes, that's how I arrived here, actually. We can close that other thread. WP All Import is working fantastically, but when I try to map all those individual fields to the one "address" field for the pods-maps plugin/add-on, no data is received. Is there a simple solution that I'm missing?

jimtrue:
Sounds like it might be a 'structure' question (ie how to combine those fields into an array to be supported by the Map Field. @JoryHogeveen might be able to give some insight on this one. I'll close the other ticket: https://wordpress.org/support/topic/importing-to-address-field/#post-11139966

JoryHogeveen:
@bwdgraphics 
Please see: http://www.wpallimport.com/documentation/custom-fields/serialized-fields/

Also, I think it's better to create a new ticket for topics like this. Let's keep this thread purely for development purposes.

JoryHogeveen:
@tiptronic Since I couldn't wait for this to be actually finished and integrated into Pods I created a plugin to work with (actually 2 plugins, the address field and the maps component).
Keep in mind that if you are not comfortable with PHP (coding in general) I'm not sure it will be suitable.
Repo: https://github.com/JoryHogeveen/pods-address-maps

tiptronic:
@JoryHogeveen I am not really a PHP Pro, but I will definitely have a look at it asap, as I can definitively use this here and there. Thanks for sharing!

oligoform:
we run into issues with this in pods 2.8. This helps a little 
https://github.com/JoryHogeveen/pods-address-maps/commit/1af43f9d56a8e96adfef14f9b41689c9d3554608  but the geolocation doesn't work anymore.
maybe we could use /tribe-common/src/Tribe/Utils/Coordinates_Provider.php for this in pods 2.8? But how? I cannot choose it from the "shipped" pods fields.

PR #3637, opened 2016-07-06
Allow extra meta/option boxes to be added in the pod-edit screens

sc0ttkclark:
Handy to add meta/option boxes to the side of the pod edit page

Bringing over PR from #3609

### DIFF
--- a/ui/admin/setup-edit.php
+++ b/ui/admin/setup-edit.php
@@ -146,8 +146,10 @@ foreach ( $field_tab_options[ 'additional-field' ] as $field_type => $field_type
  */
 do_action( 'pods_add_meta_boxes', '_pods_pod', $pod );
 
+$pod_post = get_post( $pod['id'] );
+
 /** This action is documented in wp-admin/edit-form-advanced.php */
-do_action( 'add_meta_boxes', '_pods_pod', get_post( $pod['id'] ) );
+do_action( 'add_meta_boxes', $pod_post->post_type, $pod_post );
 ?>
 <div class="wrap pods-admin">
 <div id="icon-pods" class="icon32"><br /></div>
--- a/ui/admin/setup-edit.php
+++ b/ui/admin/setup-edit.php
@@ -144,7 +144,7 @@ foreach ( $field_tab_options[ 'additional-field' ] as $field_type => $field_type
  * @see https://codex.wordpress.org/Plugin_API/Action_Reference/add_meta_boxes
  * @param array $pod The Pod object as an array
  */
-do_action( 'add_meta_boxes', 'pods_edit', $pod );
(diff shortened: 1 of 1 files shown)


ui/admin/setup-edit.php
```diff
@@ -875,6 +886,9 @@
                 </div>
             </div>
             <!-- /#submitdiv -->
+            <div class="pods-submittable-fields">
+            <?php do_meta_boxes( 'pods_edit', 'side', $pod ); ?>
```
sc0ttkclark: Let's use `_pods_pod-edit-fields` for this and the above meta boxes action

ui/admin/setup-edit.php
```diff
@@ -134,6 +134,17 @@
...
+/**
+ * Make use of the WP core meta box functionality
+ * 
+ * Currently only context 'side' is available
+ * 
+ * @since 2.7
+ * @see https://codex.wordpress.org/Plugin_API/Action_Reference/add_meta_boxes
+ * @param array $pod The Pod object as an array
+ */
+do_action( 'add_meta_boxes', 'pods_edit', $pod );
```
sc0ttkclark: Let's use `_pods_pod-edit-fields` for this

JoryHogeveen: Seems strange to me since this metabox is for the whole pod, not just the fields

sc0ttkclark: Because in one of the next feature releases we'll have groups editor AND fields editor (within a group). Let's define that now as edit-fields, then the other can be edit-groups, and we can continue etc

sc0ttkclark: Also, I know we want to match WP but because $pod is not a WP_Post, we ought to run this as the action name `pods_add_meta_boxes` instead.

sc0ttkclark:
As long as this tests correctly and shows meta boxes that you added, I'm good with this for merge.

JoryHogeveen:
I'll test with your changes.
Why use `_pods_pod-edit-fields` instead of `_pods_pod-edit` ??

sc0ttkclark:
_pods_pod is the post type, we'll eventually want to make our pod editor a little more core-friendly so we can make use of meta boxes for real in our pod editor. As for the identifier being 'edit-fields' instead of just 'edit', we plan on expanding to support group editing soon so we'll want to differentiate between the multiple edit screens for this purpose.

JoryHogeveen:
Ok. (since the i18n component is dependent on this) This is not planned for 2.7 I guess?

sc0ttkclark:
Let's just use `_pods_pod` and we can use `_pods_group` in the other edit screen that's group-centric.

sc0ttkclark:
Ready to merge once you confirm it's working

JoryHogeveen:
Ok, tested aswell. Merge is fine with me!

Issue #12, opened 2011-09-14
PICK - Support for WP Objects (CPT, CT, Comments)


sc0ttkclark:
Post Types exist in wp_posts.post_type, where as Taxonomies exist in wp_terms_taxonomy.taxonomy

We should use one set of code for Post Types in general, and one set of code for Taxonomies in general - It doesn't make much sense to add this as additional code - it should replace the existing Categories / Posts / Pages code.

A slight change in the way the relationship is stored will be required though, we should discuss further

Scott Kingsley Clark - 10 Jan 2010, 2:33pm  
i think it may be best to combine some of our existing post types like Post / Page as well as Taxonomies like Categories to simplify the PICK relationships and code necessary, by making it generalized it will cover all Post Types and Taxonomies

Matt - 13 Jan 2010, 10:07am 
Agreed. Then the user could use the "PICK filter" column to manually choose the post_type(s)...

post_type IN ('page', 'post', 'something_else')

Scott Kingsley Clark - 13 Jan 2010, 12:51pm 
Well.. it should be more in the User Interface, so they could select which post_type (like an attribute), perhaps we should save this for when we redo the UI for the Setup tabs

Scott Kingsley Clark - 3 Sep 2010, 12:45am  
With all the new changes, to relate going forward we will store CPT and Taxonomies (along with other stuff in the 'object' field. Examples:
- object = post (pickval = empty)
- object = page (pickval = empty)
- object = cpt (pickval = cptname)
- object = tag (pickval = empty)
- object = category (pickval = empty)
- object = taxonomy (pickval = taxonomyname)
- object = comment (pickval = empty)
- object = user (pickval = empty)
- object = pod (pickval = podname)

All existing Category pickval data goes into the options JSON

Scott Kingsley Clark - 3 Sep 2010, 12:47am  
The reasoning behind the separation of Pod objects and CPT objects are that there can be completely separate CPT built outside of Pods, which shouldn't cause an additional query to check if it's a Pods-based CPT or not in use

Scott Kingsley Clark - 3 Sep 2010, 12:49am  
I'm not partial to storing post / page in the CPT object pickval, same with tag / category in the taxonomy object pickval. That could simplify things a lot possibly.

PR #3644, opened 2016-07-07
Add PODS_DIR to directories that are checked by field_loader()

### LINKED ISSUE #3643: PodsForm::field_loader doesn't work with unit tests
With the default setup of the unit test suite the field_loader() function won't properly load the fields files.

PR incoming

jamesgol:
Resolves #3643

### DIFF
--- a/classes/PodsForm.php
+++ b/classes/PodsForm.php
@@ -1208,6 +1208,7 @@ class PodsForm {
         $plugins_dir = realpath( WP_PLUGIN_DIR );
         $muplugins_dir = realpath( WPMU_PLUGIN_DIR );
         $abspath_dir = realpath( ABSPATH );
+        $pods_dir = realpath( PODS_DIR );
 
         if ( !class_exists( $class_name ) ) {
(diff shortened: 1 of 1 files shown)


jamesgol:
```
svn: OPTIONS of 'https://develop.svn.wordpress.org/tags/4.5.3/tests/phpunit/includes': could not connect to server (https://develop.svn.wordpress.org)
```
Is there a way to kick of the tests again later without pushing another commit?

jamesgol:
Pods_Unit_Tests\Test_Traversal::test_field_base with data set #12 is failing because Pods->index() is returning the data in the field.
```
	/**
	 * Return the item name
	 *
	 * @return string
	 * @since 2.0
	 */
	public function index () {
		return $this->field( $this->data->field_index );
	}
```
The test is expecting the display output, so I'm suspecting that the test and the index() function are bugged.

This is the test that is failing:
```
			$this->assertEquals( $data[ 'data' ][ $data[ 'field_index' ] ], $p->index(), sprintf( 'Item index not as expected (%s) [%s]', $data[ 'field_index' ], $variant_id ) );
```

PR #3646, opened 2016-07-07
Improve field alignment on setting pages

JoryHogeveen:
I can't find any actual need of the `margin-left` property. It only breaks alignment.

Changed `max-width` to `25em` to match with the default WP setting fields css class "regular-text" as seen in the general settings.

This:
testsetting wp pods development wordpress

Becomes this:
testsetting fixed wp pods development wordpress


JoryHogeveen:
https://github.com/pods-framework/pods/pull/3649

PR #3691, opened 2016-07-20
Fix WPML compatibility

sc0ttkclark:


sc0ttkclark:
@srdjan-jcc is this officially all good to go now?

srdjan-jcc:
@sc0ttkclark It's ready now.
Thank you.

sc0ttkclark:
Thanks!

classes/PodsAPI.php
```diff
@@ -7628,8 +7640,12 @@ public function get_table_info ( $object_type, $object, $name = null, $pod = nul
 			    'tt.taxonomy' => '`tt`.`' . $info[ 'field_type' ] . '` = "' . $taxonomy . '"'
 		    );
 
+            /*
+             * @todo wpml-comp WPML API call for is_translated_taxononomy
+             * @todo wpml-comp Check if WPML filters can be applied afterwards
+             */
 		    // WPML Support
-		    if ( is_object( $sitepress ) && !empty( $current_language ) && $sitepress->is_translated_taxonomy( $taxonomy ) && !$icl_adjust_id_url_filter_off ) {
+		    if ( is_object( $sitepress ) && !empty( $current_language ) && $sitepress->is_translated_taxonomy( $taxonomy ) && apply_filters( 'wpml_setting', true, 'auto_adjust_ids' ) ) {
```
sc0ttkclark: Is there not a taxonomy version of `wpml_is_translated_post_type` filter here?

classes/PodsAPI.php
```diff
@@ -8343,6 +8361,11 @@ public static function get_current_language() {
 				} elseif ( isset( $current_screen->base ) && ( $current_screen->base == 'term' || $current_screen->base == 'edit-tags' ) ) {
 					// @todo MAYBE: Similar function like get_post_type for taxonomies so we don't need to check for $_GET['taxonomy']
 					if ( ! empty( $_GET['taxonomy'] ) ) {
+					    /*
+					     * @todo wpml-comp API call for taxonomy needed!
+					     * Suggested API call:
+					     * add_filter( 'wpml_is_translated_taxonomy', $_GET['taxonomy'], 10, 2 );
```
sc0ttkclark: It sounds like we should be using apply_filters() here, right?

classes/PodsAPI.php
srdjan-jcc: Not yet :( I opened API call request for WPML.

classes/PodsAPI.php
srdjan-jcc: I added that as reminder. Should be switched to filter when available.

classes/PodsAPI.php
sc0ttkclark: Well we can only hope :)

So this works as it is for now?

Issue #3695, opened 2016-07-21
New MV based Pick field as full drop-in replacement


pglewis:
PR: #3697  

We cannot easily implement the new pick field in parallel with the old one due to the special treatment we give to 'pick' throughout core.  The next major step is for the re-built pick field to fully replace the old one.

Add New Pod:
- [x] Track down and fix up issues with depends-on/excludes-on for select lists

Fix Pod Edit screen:
- [x] Find what's snagging in the JS `collapsible` method when clicking edit on a field
- [x]  Rework `pods_sister_field` JS function
- [x] Selected attribute not being set on initialization for some lists (field type is one culprit)
- [x] Track down slow initial load when expanding the collapsible field row

pglewis:
Now that we can create and edit Pods, this is ready to merge back in.

PR #3696, opened 2016-07-21
Feature/#3695 pick replacement

pglewis:
Issue: #3695


pglewis:
Made it against the wrong branch

PR #3702, opened 2016-07-30
travis-ci: test with PHP 7.1

Ramoonus:


.travis.yml
```diff
@@ -36,6 +36,7 @@ cache:
 
 before_script:
     - if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then phpenv config-rm xdebug.ini; fi
+    - if [[ $TRAVIS_PHP_VERSION != '7.1' ]]; then phpenv config-rm xdebug.ini; fi
```
pglewis: Because 7.1 is not hhvm, the line above will pass and cause a failure before we get here.  

Try: `if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]] && [[ $TRAVIS_PHP_VERSION != '7.1' ]]; then phpenv config-rm xdebug.ini; fi`

PR #3704, opened 2016-07-30
travis-ci: mysql 5.7

Ramoonus:


pglewis:
What are we doing here?  I don't know of a single issue that has turned up so far due to MySQL versions so any effort to expand the testing in that direction is probably just extra technical debt.  I would like to see MariaDB testing in place at some point, but I think picking one "definitive" version of MySQL in the matrix is enough right now.

pglewis:
I mean, we could merge this... it has been commented out anyway.  But I'm of the opinion that we just remove it altogether and continue to stick with the default, lacking an existing pain point to be addressed.

Ramoonus:
MariaDB is slightly more tricky to implement than MySQL

Issue #3708, opened 2016-08-05
Currency field doesn’t allow the entry of negative numbers


PodsBot:
Currency field doesn’t allow the entry of negative numbers submitted via Slack by jimtrue

pcfreak30:
After fixing and testing this, I found `5.00-` converts to `5.00` and same for `5.0-0`. @sc0ttkclark this an issue?

PR #3709, opened 2016-08-05
*Allow negative numbers

pcfreak30:
#3708

### DIFF
--- a/ui/fields/currency.php
+++ b/ui/fields/currency.php
@@ -44,7 +44,7 @@ $currency_sign = PodsField_Currency::$currencies[ $currency ];
             ?>]$/.test( $( this ).val() ) ) {
                 var newval = $( this )
                     .val()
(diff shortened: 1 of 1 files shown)


ui/fields/currency.php
```diff
@@ -44,7 +44,7 @@
             ?>]$/.test( $( this ).val() ) ) {
                 var newval = $( this )
                     .val()
-                    .replace( /[^0-9\<?php echo esc_js( $currency_sign ); ?>\<?php
+                    .replace( /[^0-9-\\<?php echo esc_js( $currency_sign ); ?>\<?php
```
sc0ttkclark: Should that be a `\-`?

pcfreak30: No, I tested it in js console as I had the same thought. It doesn't have a/0 before so not interpreted as a range character.

PR #3720, opened 2016-08-13
Properly update bidirectional taggable relations

### LINKED ISSUE #3477: relationship with taggable creates the post, but not relationships
if i have post type A and post type B with a bi-directional relationship.
And I am using autocomplete and I create a relationship to a new post form type A to B. and save A.
The new pod in B gets created but the relationship is missing.

caleb:
Fixes #3477

When editing an object's bidirectional taggable relation, and you add a related object that doesn't exist, the reverse relation is now correctly set up.

I just learned about Pods today so this fix could be completely off base, but it works for me.

The problem was that the `save` method for the `pick` field was being passed the original array of ids instead of the new variable that contains the id of any newly added objects.

(I'm using Pods 2.6.6 and Wordpress 4.5.3)

Cheers!

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -3759,7 +3759,7 @@ class PodsAPI {
                         $this->save_relationships( $params->id, $value_ids, $pod, $fields[ $field ] );
 
                     // Run save function for field type (where needed)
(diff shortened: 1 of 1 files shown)


jamesgol:
I just spent a bit troubleshooting this issue and was about to submit a PR for it when I decided to search for issues and see if there was an existing issue to reference.  Turns out there was, and this PR which is exactly the same as I was going to submit.  I can confirm that this fixes the issue and does not affect anything else I've seen (because in all other cases $values is the same as $value_ids unless there are new items that are tagged)

Issue #3707, opened 2016-08-04
Bug: Field option multiple dependencies not working properly across tabs in edit Pod UI


JoryHogeveen:
See https://github.com/pods-framework/pods-seo/pull/6

**Use case:**

If you put two dependencies for a field in the tab "Advanced" that is dependent on a setting in the tab "Basic", this is not working nicely.
When opening the field these dependencies are not processed. Only when the actual dependency field is edited will the dependency functionality work.

Found this problem in the patch for Pods SEO where XML Sitemaps checkbox (Advanced tab) is only avaiable for the Field Type "_file_" (Basic tab) with File Types set to "_images_" (Additional Field Options tab).

When opening a different field type and going directly to the Advanced tab the XML Sitemaps option is visible. Only after clicking on the Field Type select box in the Basic tab is this changed according to dependencies.

**Use case 2:**
When adding a dependency that doesn't exist it validates as true instead of false. It seems more logical to me to make this false.

Example (also Pods SEO)
```php
'depends-on' => array(
    // This is a field type specific option (only file fields) 
    // It is considered true if it doesn't exists for a field
    'file_type' => 'images',

    // This is a Field Type dependency, available in all field types (works as expected)
    'type' => 'file',
),
```

JoryHogeveen:
Did some extra research on this.

The issue happens because in this specific case the `file_type` is `images` by default. This means that the `file_type` dependency would be valid, but the `type` dependency isn't.

So if the field isn't a `file` field, this value is the default value and for some reason the dependency script thinks it's valid even though the first dependency (field type) is not `file`.

I'd say if the option has two dependencies, both of them must be valid, even if they reside in different tabs.

I'm having some trouble getting through the whole dependency script in `jquery.pods.js` (no documentation..). If anyone can shed some light here that would be great.

JoryHogeveen:
Fixed in #3723

PR #3723, opened 2016-08-15
Validate field option dependencies based on location within in tabs

JoryHogeveen:
Check if this element is a child from an 'Advanced Field Options' group.
If so, set the value to empty if this is not the current field type group.

Fixes dependency compatibility of https://github.com/pods-framework/pods/issues/3707 and https://github.com/pods-framework/pods-seo/pull/6

### DIFF
--- a/ui/js/jquery.pods.js
+++ b/ui/js/jquery.pods.js
@@ -844,6 +844,27 @@
 					$field = $el,
 					val = $el.val();
 
+                /**
+                 * Check if this element is a child from an 'advanced field options' group.
+                 * If so, set the value to empty if this is not the current field type group
+                 * Fixes dependency compatibility
+                 *
+                 * @todo Validate & improve this
+                 */
+                // Are we in the "Fields" tab?
+                if ( $current.parents('#pods-manage-fields').length ) {
+                    // And are we also in the "Additional Field Options" tab?
+                    if ( $el.parents('.pods-additional-field-options').length ) {
+                        // Get this field's type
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Needs secondary confirmation that it fixes the issue

JoryHogeveen:
@jimtrue One test case is installing https://github.com/pods-framework/pods-seo/pull/6 and verify that the XML sitemap options are ONLY available for fie `file` type set to `images`.
On all other occasions and fields it should say: 
"_This field does not currently support WP SEO XML Sitemap integration._"

We also need to double-verify that it's not breaking any other dependency functionalities.

jimtrue:
Is this message ""This field does not currently support WP SEO XML Sitemap integration." coming from pods-seo? 

Also, what other dependency functionalities?

JoryHogeveen:
Yes, it's from pods-seo.

We'll, what I did is check the dependency settings in the code and check if they all still work correctly in the UI. There are a lot, don't know if this is documented anywhere though...

jimtrue:
@sc0ttkclark I'm happy to check this one, but I don't usually hunt down the dependencies in the code (that's outside my area of expertise); do we have this 'self-documented' in the code?

JoryHogeveen:
@sc0ttkclark @jimtrue 
Any news on testing progress?

JoryHogeveen:
@sc0ttkclark 
With this one merged all is ready for pods-framework/pods-seo#6 aswell

Issue #3736, opened 2016-08-23
Labels not working correctly with ACT's


JoryHogeveen:
Related: https://github.com/pods-framework/pods/issues/3735

Just like with Post Types and Taxonomies we need to revisit all label handling for ACT's

sc0ttkclark:
Fixed via #3737

Issue #3735, opened 2016-08-22
Text label with problem editing


saulolago:
At some time I have noticed this problem. When we try to edit the Pods of the label text does not work. See the picture: http://www.awesomescreenshot.com/image/1503105/c5b49fd945226b27d22a4e5c29943628

JoryHogeveen:
@saulolago 
That is strange, just tested this and it works like expected.
Have you tries clearing you cache?

saulolago:
Yes, I cleared the cache but still does not work @JoryHogeveen

JoryHogeveen:
Can you try two things?

- Is the same problem happening on the other custom post types?
- If you disable the admin modifications, does it work then?

saulolago:
- It is going on with any type pods created
- Deactivated and activated again but not resolved

**Important**: I am using Advanced Content Type

JoryHogeveen:
@saulolago 
Ah, that was it. Same problem here. Will dive into this a.s.a.p.

JoryHogeveen:
Fixed in #3737

PR #3737, opened 2016-08-23
Redo ACT label handling

JoryHogeveen:
See https://github.com/pods-framework/pods/issues/3736 and https://github.com/pods-framework/pods/issues/3735

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -1038,7 +1038,7 @@ class PodsAdmin {
                     'help' =>__( 'help', 'pods' ),
                     'type' => 'text',
                     'default' => '',
-                    'object_type' => array( 'taxonomy' )
+                    'object_type' => array( 'taxonomy', 'pod' )
                 ),
                 'label_duplicate' => array(
                     'label' => __( 'Duplicate', 'pods' ),
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -1181,21 +1181,21 @@ class PodsAdmin {
                     'help' =>__( 'help', 'pods' ),
                     'type' => 'text',
                     'default' => '',
-                    'object_type' => array( 'post_type', 'pod' )
+                    'object_type' => array( 'post_type' )
                 ),
                 'label_insert_into_item' => array(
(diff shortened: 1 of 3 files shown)


JoryHogeveen:
@sc0ttkclark 
There are also a lot of labels within the UI templates. What shall we do with those?

Plus, (for the i18n component). Since there is nog "register" filter for ACT's, how about adding one so the labels can be filtered?

classes/PodsAdmin.php
```diff
@@ -1038,26 +1037,88 @@ public function admin_setup_edit_options ( $pod ) {
                     'default' => '',
                     'object_type' => array( 'taxonomy' )
                 ),
-                // @todo Why was label_view added previously? Can't find it in WP
```
JoryHogeveen: This was a ACT label only

classes/PodsAdmin.php
```diff
@@ -1017,13 +1017,12 @@ public function admin_setup_edit_options ( $pod ) {
                     'default' => '',
                     'object_type' => array( 'taxonomy' )
                 ),
-                // @todo Why was label_edit added previously? Can't find it in WP
```
JoryHogeveen: This was a ACT label only

JoryHogeveen:
@sc0ttkclark 
There are 3 locations I placed a filter. I'd say it would be better to do this at an earlier stage, but I'm not sure if caching is possible for languages that aren't done with Polylang or WPML.

`pods_advanced_content_type_pod_data` and `pods_advanced_content_type_pod_data_**POD_NAME**`

Please let me know if you would like to keep it as it is now or if you want to relocate this filters to an earlier stage. Also let me know where, I was thinking about `PodsMeta` right after loading the Pod objects.

JoryHogeveen:
@sc0ttkclark 
Could you reply on my last comment?

> Please let me know if you would like to keep it as it is now or if you want to relocate this filters to an earlier stage. Also let me know where, I was thinking about PodsMeta right after loading the Pod objects.

sc0ttkclark:
@JoryHogeveen i don't think the filter like `pods_advanced_content_type_pod_data` is going to be the right way to go, we should wait for the magic of PodsObject in 2.8 to enable something like that.

JoryHogeveen:
In that case please make a reminder or something :)

Issue #3738, opened 2016-08-23
Add description field to file upload input


pcfreak30:

JoryHogeveen:
Nice, +1 on this!

sc0ttkclark:
Closing this as won't fix, we've redone the file field type in Pods 2.7 and it offers much better flexibility to do things in a better workflow.

PR #3739, opened 2016-08-23
Add description field to file upload input

pcfreak30:
#3738


sc0ttkclark:
Closing this as won't fix, we've redone the file field type in Pods 2.7 and it offers much better flexibility to do things in a better workflow.

PR #3768, opened 2016-09-05
Keep Plupload instances in `windows.pods_uploader`

### LINKED ISSUE #3763: Expose Plupload objects
Right now, Plupload fields are initialized inside of a closure, and the variable pointing the instance isn't available to any other code. This would be fine, except I haven't been able to figure out (after probably 4+ hours of searching) how to get the instance using the Plupload or Moxie APIs :scream_cat: It should be possible to get the Plupload instance using the jQuery plugin, but the jQuery Plupload plugin doesn't seem to be available.

Having access to this Plupload instance would make it possible to bind event listeners to the file uploader, and would make it possible to write custom JS code to fix issues such as #1377. I personally plan to use it to add some JS that automatically submits a Pods form after a file is uploaded.

thirdender:
This change fixes #3763 by making Plupload instances available to other code by storing them in `windows.pods_uploader[id]`.

It also removes `<?php echo esc_js( pods_js_name( $attributes[ 'id' ] ) ); ?>` from the variable names used in this scope, as these variables are only available in this closure.

### DIFF
--- a/ui/fields/plupload.php
+++ b/ui/fields/plupload.php
@@ -205,17 +205,23 @@ else
             } );
         } );
 
-        var pods_uploader_<?php echo esc_js( pods_js_name( $attributes[ 'id' ] ) ); ?> = new plupload.Uploader( <?php echo json_encode( $plupload_init ); ?> ),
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Needs follow up for Pods 2.7 compatibility

Issue #3777, opened 2016-09-08
Fix sc0ttkclark's fail :P


pcfreak30:
See 07463f9ca9efaa30136f84318e3c41d20b3e05dc

quasel:
really? shit happens wait until you get cought! ^^

pcfreak30:
@quasel I usually see a glaring error in phpstorm before committing :P. Anyways fun's over, moving on..

PR #3778, opened 2016-09-08
*Fix sc0ttkclark's fail

pcfreak30:
#3777


sc0ttkclark:
I don't always fail, but when I do, there's @pcfreak30

jimtrue:
In the future, can we please actually describe what you're fixing? It makes it more efficient for the person writing the change log.

pcfreak30:
@jimtrue it was a syntax error with a block "}". Can't be very descriptive about it.

PR #3785, opened 2016-09-14
Add pods_ui_get_params hook

telwing:
Allows query parameters to be filtered before the data is fetched and displayed on the admin side. The main purpose of this hook is to enable default ordering of advanced content types.

Usage:

``` php
function set_pod_default_orderby($params, $pod) {
  if ($pod == 'pod_slug') {
    $params['orderby']['field_name'] = 'ASC';
  }

  return $params;
}
add_filter('pods_ui_get_params', 'set_pod_default_orderby', 10, 2);
```


sc0ttkclark:
Sorry for the confusion on this next request, can you also use `apply_filters( 'pods_ui_get_params', ... )` instead of do_hook? Just pass in a third argument into the filter which is `$this`.

We're working on switching everything over, so it'll be great to have new hooks added the new (correct) way.

PR #3800, opened 2016-09-30
Patch the responsive markup WP edit.php UI to Pods UI

JoryHogeveen:
Adds the required classes and attributes to form nice responsive view for PodsUI pages.
- Adds the `column-primary` class to the first field (ignores a possible first checkbox field)
- Adds the `data-colname` attributes
- Adds our own `column-secondary` that can be passed with a field when calling PodsUI. A secondary field will always be show so the user won't need to toggle
- Reformatted the code to work with arrays of classes instead of strings. (Easier to manage)
# Edit Pods screen

**Before:**
image

**After:**
image
# Manage roles screen

**Before:**
image

**After:**
image

### DIFF
--- a/classes/PodsUI.php
+++ b/classes/PodsUI.php
@@ -3644,7 +3644,7 @@ class PodsUI {
                                         <input type="hidden" name="order[]" value="<?php echo esc_attr( $row[ $this->sql[ 'field_id' ] ] ); ?>" />
                                         <?php
                                     }
-                                    ?><button type="button" class="toggle-row"><span class="screen-reader-text">Show more details</span></button>
+                                    ?><button type="button" class="toggle-row"><span class="screen-reader-text"><?php esc_html_e( 'Show more details', 'pods' ); ?></span></button>
                 </td>
 <?php
                                 }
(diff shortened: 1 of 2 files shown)


JoryHogeveen:
This also fixes responsive layouts for ACT's

JoryHogeveen:
Example for `column-secondary` (i18n Component):
image

classes/PodsUI.php
```diff
@@ -3270,19 +3270,33 @@ public function table ( $reorder = false ) {
 
                             $width = '';
 
+                            $column_classes = array(
+                                'manage-column',
+                                'column-' . esc_attr( $id )
```
sc0ttkclark: `esc_attr()` no longer needed here as it's called further down below.

sc0ttkclark:
Can you provide an example of how to use this by implementing this for the Roles component?

JoryHogeveen:
What to you mean? The `column-secondary` class?

sc0ttkclark:
That's right, just implementing the class, or maybe we should add an option for it in the manage fields.

JoryHogeveen:
Ah ok sure:

Roles.php Line 81-113
```php
$ui = array(
    'component' => $component,
    'data' => $roles,
    'total' => count( $roles ),
    'total_found' => count( $roles ),
    'items' => 'Roles',
    'item' => 'Role',
    'fields' => array(
        'manage' => array(
            'label' => array( 'label' => __( 'Label', 'pods' ) ),
            'name' => array( 'label' => __( 'Name', 'pods' ) ),
            'capabilities' => array( 
                'label' => __( 'Capabilities', 'pods' ),
                'classes' => array( 'column-secondary' ) // <<< The custom classes
            ),
            'users' => array(
                'label' => __( 'Users', 'pods' ),
                'type' => 'text',
                'options' => array(
                    'text_allow_html' => 1,
                    'text_allowed_html_tags' => ''
                )
            )
        )
    ),
    /* etc. */
);
```

This would result in a mobile view like this:
image

classes/PodsUI.php
JoryHogeveen: Check!

JoryHogeveen:
> or maybe we should add an option for it in the manage fields
These are only available to the ACT types I guess?

JoryHogeveen:
By default it always shows the title for ACT's (like WP does). Not sure if this should be a real field attribute for two reasons:

- Differs from WP default while users expect the default (ACT behave like default WP objects..)
- Can give confusion since this is not available for non-ACT types while all other field options are

I think it's also better to change the key to `overview_column_classes` or something so it's clear what this key is for.
Any thoughts?

JoryHogeveen:
commit #b7705a5:

Before:
image

After:
image

Issue #3828, opened 2016-10-25
PHP7 Compatibility issues multiple files in Pods Core


PodsBot:
PHP7 Compatibility issues multiple files in Pods Core submitted via Slack by jimtrue

Referenced from this Forum post:
https://wordpress.org/support/topic/php7-error-messages/
- [x] classes/PodsData.php - Extension 'mysql_' is deprecated since PHP 5.5 and removed since PHP 7.0 - use mysqli instead. - Lines 2141, 2145, 2199, 2203
- [x] components/Builder/Modules issues in multiple files - Use of deprecated PHP4 style class constructor is not supported since PHP 7 (field/PodsBuilderModuleField.php, line 21, form/PodsBuilderModuleForm.php, line 21, list/PodsBuilderModuleList.php, line 21, single/PodsBuilderModuleSingle.php, line 21, view/PodsBuilderModuleView.php, line 21)
- [x] components/Markdown.php - Use of deprecated PHP4 style class constructor is not supported since PHP 7, line 230
- [x] deprecated/deprecated.php - Extension 'mysql_' is deprecated since PHP 5.5 and removed since PHP 7.0 - use mysqli instead. - Lines 89, 91, 93, 101

jimtrue:
From this plugin checker: PHP linter: https://wordpress.org/plugins/php-compatibility-checker/

```
FILE: /home/www/wp-content/plugins/pods/classes/PodsData.php
---------------------------------------------------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
---------------------------------------------------------------------------------------------------------------
 2141 | ERROR | Extension 'mysql_' is deprecated since PHP 5.5 and removed since PHP 7.0 - use mysqli instead.
 2145 | ERROR | Extension 'mysql_' is deprecated since PHP 5.5 and removed since PHP 7.0 - use mysqli instead.
 2199 | ERROR | Extension 'mysql_' is deprecated since PHP 5.5 and removed since PHP 7.0 - use mysqli instead.
 2203 | ERROR | Extension 'mysql_' is deprecated since PHP 5.5 and removed since PHP 7.0 - use mysqli instead.
---------------------------------------------------------------------------------------------------------------


FILE: /home/www/wp-content/plugins/pods/components/Builder/modules/field/PodsBuilderModuleField.php
---------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------
[...]

PR #3830, opened 2016-10-27
Fix PHP7 Compatibility issues

JoryHogeveen:
See #3828 
- [x] Constructors
- [x] PodsData.php
- [x] deprecated/deprecated.php

### DIFF
--- a/classes/PodsData.php
+++ b/classes/PodsData.php
@@ -2138,7 +2138,7 @@ class PodsData {
             $wpdb->term_relationships
         );
 
-        $showTables = $wpdb->get_results( 'SHOW TABLES in ' . DB_NAME, ARRAY_N );
+        $showTables = $wpdb->get_results( 'SHOW TABLES in ' . DB_NAME, ARRAY_A );
 
         $finalTables = array();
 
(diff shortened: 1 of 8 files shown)


JoryHogeveen:
@sc0ttkclark Please check deprecated.php.

> One problem with the pod_query function in deprecated.php. $wpdb->dbh is a protected property and not allowed. How did this even work at all?!

What would you like to do with this function? Is it even used anywhere?

classes/PodsData.php
```diff
@@ -2196,11 +2196,13 @@ public static function get_table_columns ( $table ) {
...
     public static function get_column_data ( $column_name, $table ) {
-        $describe_data = mysql_query( 'DESCRIBE ' . $table );
+	    global $wpdb;
+
+        $describe_data = $wpdb->get_results( 'DESCRIBE ' . $table, ARRAY_A );
 
         $column_data = array();
 
-        while ( $column_row = mysql_fetch_assoc( $describe_data ) ) {
+        foreach ( $describe_data as $column_row ) {
```
JoryHogeveen: This could be simplified by just using `$describe_data` since the contents will be the same when using `$wpdb`.

classes/PodsData.php
```diff
@@ -2196,11 +2196,13 @@ public static function get_table_columns ( $table ) {
...
             $column_data[] = $column_row;
         }
```
JoryHogeveen: This function could be simplified by just using `$describe_data` instead of creating a new var (`$column_data`) since the contents will be the same when using `$wpdb`.

sc0ttkclark:
If there's a way to refactor pod_query to act like it does now but have it work, great. Otherwise I'm okay with throwing doing it wrong notices and map the query directly to pods_query() because we can't do anything further at that point.

PR #3836, opened 2016-10-27
Fix slash problem with LIKE statements

### LINKED ISSUE #3823: Too Many Backslashes "\'s" Added When Using Quotes in Search Field
I'm having an issue with search filters in pods. In a listing of custom pods (businesses), I've added the filter to search the post_title. The search works great until you add a quote to the search field. 

For example, searching for a business with the title:
Ray's

Turns into:
Ray\\\'s

The search results return the following error. Please help.

Database Error; SQL: SELECT DISTINCT `t`.\* FROM `wp_posts` AS `t` WHERE ( ( `t`.`post_status` IN ( "publish" ) ) AND ( `t`.`post_type` = "businesses" ) ) AND ( `t`.`post_title` LIKE '%Ray\\\'s%' ) ORDER BY `t`.`menu_order`, `t`.`post_title`, `t`.`post_date` LIMIT 0, 15; Response: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 's%' ) ORDER BY `t`.`menu_order' at line 7

JoryHogeveen:
Fixes #3823 

About sanitizing `$wpdb->esc_like()` input:

> This needs to be done **after** using $wpdb->esc_like(), to ensure correct and secure slashing of the string.

https://codex.wordpress.org/Class_Reference/wpdb/esc_like

### DIFF
--- a/includes/data.php
+++ b/includes/data.php
@@ -111,7 +111,7 @@ function pods_sanitize_like( $input ) {
 	}
 	else {
 		global $wpdb;
-		$input = wp_unslash( $input );
+		$input = pods_unslash( $input );
 
 		if ( pods_version_check( 'wp', '4.0' ) ) {
 			$output = pods_sanitize( $wpdb->esc_like( $input ) );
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Needs further review

PR #3847, opened 2016-11-04
Add Magic Tag reference for Templates

sc0ttkclark:
We should add brief information about the magic tags on the Template screen.

### DIFF
--- a/components/Templates/class-pods_templates.php
+++ b/components/Templates/class-pods_templates.php
@@ -212,6 +212,13 @@ class Pods_Templates_Frontier {
 			'slug' => 'pod_reference',
 			'groups' => array()
 		) );
+		add_meta_box( 'pod_reference', __( 'Magic Tag Reference', 'pods' ), array(
+			$this,
+			'render_metaboxes_custom'
+		), '_pods_template', 'side', 'default', array(
+			'slug' => 'tag_reference',
+			'groups' => array()
+		) );
 
 	}
 


sc0ttkclark:
Need content added to https://github.com/pods-framework/pods/blob/feature/magic-tag-reference-for-templates/components/Templates/includes/element-tag_reference.php

sc0ttkclark:
Pending some work by Jim to add some content in that file.

jimtrue:
In reference to this, should we point them to our documentation and where exactly on the screen would this show up? (ie how much room do we have for content?)

jimtrue:
ie, if it's in the 'help' viewer, we could put a good document up there, but don't know if that's the preferred method going forward for stuff like this. It makes sense to me, but not sure if that feature is being 'phased' out.

sc0ttkclark:
This is in a new meta box, it would appear to the right side as well I believe.

jimtrue:
Well, thinking forward for something like this, ideally in a UX, if it's in a meta box, it should actually 'populate' the template into the Template Editor, ie 

tag-meta-box

And selecting one of those would automatically push the template into the editor with the cursor where you'd want to put your IF or EACH field (preferably also allowing you to auto-complete, which doesn't work at present, because it wouldn't have an @ in front of it.

jimtrue:
@sc0ttkclark What format does this need to be in? Markdown? I want to link out to our doc pages (since they won't be moving for those Template Tags).

sc0ttkclark:
Format on this should be in HTML

components/Templates/class-pods_templates.php
```diff
@@ -221,6 +221,15 @@ public function add_metaboxes( $slug, $post = false ) {
 				'groups' => array(),
 			)
 		);
+		add_meta_box(
+			'pod_reference', __( 'Magic Tag Reference', 'pods' ), array(
```
tr1b0t: * :no_entry_sign: Only one argument is allowed per line in a multi-line function call
* :no_entry_sign: Only one argument is allowed per line in a multi-line function call

<sub>hash: 1c0d3fc2ebfa87e21348bfb2133f0c3f</sub>

components/Templates/class-pods_templates.php
```diff
@@ -221,6 +221,15 @@ public function add_metaboxes( $slug, $post = false ) {
...
+				$this,
+				'render_metaboxes_custom'
```
tr1b0t: * :no_entry_sign: Comma required after last value in array declaration
* :no_entry_sign: Each array item in a multi-line array declaration must end in a comma

<sub>hash: fa322500090b9ad482beacd9020a39fa</sub>

components/Templates/class-pods_templates.php
```diff
@@ -221,6 +221,15 @@ public function add_metaboxes( $slug, $post = false ) {
...
+			), '_pods_template', 'side', 'default', array(
```
tr1b0t: * :no_entry_sign: Only one argument is allowed per line in a multi-line function call
* :no_entry_sign: Only one argument is allowed per line in a multi-line function call
* :no_entry_sign: Only one argument is allowed per line in a multi-line function call
* :no_entry_sign: Only one argument is allowed per line in a multi-line function call

<sub>hash: ba6f6be3ab04cb9a717e310cbe508868</sub>

components/Templates/class-pods_templates.php
```diff
@@ -221,6 +221,15 @@ public function add_metaboxes( $slug, $post = false ) {
...
+				'slug' => 'tag_reference',
```
tr1b0t: :warning: Array double arrow not aligned correctly; expected 3 space(s) between "'slug'" and double arrow, but found 1.

<sub>hash: 4dcc1798e03a9e42e644225566a6026f</sub>

components/Templates/class-pods_templates.php
```diff
@@ -221,6 +221,15 @@ public function add_metaboxes( $slug, $post = false ) {
...
+				'groups' => array()
```
tr1b0t: * :no_entry_sign: Each line in an array declaration must end in a comma
* :no_entry_sign: Each array item in a multi-line array declaration must end in a comma

<sub>hash: 4ff5747177efdf40a6f983794bc3282a</sub>

PR #3803, opened 2016-10-01
oEmbed field: Fix some notices for PHP and JS

JoryHogeveen:
Title says it all.
Mostly JS/PHP linting stuff. One warning ($css_id not defined)

PR #3850, opened 2016-11-05
Make field labels translatable

### LINKED ISSUE #3849: Make field type labels translatable
Currently the field labels are fixed, wrap all labels in __() to enable i18n.

JoryHogeveen:
Fixes #3849

**Todo for 2.7:**
- [ ] Make `Link` field label translatable
- [x] Make `oEmbed` field label translatable #3803 
- [x] Make `Address` field label translatable #3634

### DIFF
--- a/classes/fields/taxonomy.php
+++ b/classes/fields/taxonomy.php
@@ -12,6 +12,7 @@ class PodsField_Taxonomy extends PodsField_Pick {
      * @since 2.0
      */
     public function __construct () {
+	    parent::__construct();
 		// this field type just maps to the relationship field
     }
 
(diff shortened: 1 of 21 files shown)


classes/fields/taxonomy.php
```diff
@@ -12,6 +12,7 @@ class PodsField_Taxonomy extends PodsField_Pick {
      * @since 2.0
      */
     public function __construct () {
+	    parent::__construct();
```
sc0ttkclark: Shouldn't there be a Taxonomy label for this?

JoryHogeveen: > this field type just maps to the relationship field

No real use I guess since everywhere in Pods this field is called a relationship field. 
From the looks of it this file is there for backwards compatibility?

sc0ttkclark: We'll leave it since we're not doing that anywhere else (like Avatar) since we're not calling the properties via `static::$property` and only using `self::$property` in the various places.

PR #3853, opened 2016-11-06
Make new Custom Taxonomies Hierarchical by default

### LINKED ISSUE #3846: Make default for new Custom Taxonomies be Hierarchical
Currently new taxonomies are tag-like, we should use category-like for the default instead.

JoryHogeveen:
Fixes #3846

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -1295,7 +1295,7 @@ class PodsAdmin {
                     'label' => __( 'Hierarchical', 'pods' ),
                     'help' => __( 'help', 'pods' ),
                     'type' => 'boolean',
-                    'default' => false,
+                    'default' => true,
                     'dependency' => true,
                     'boolean_yes_label' => ''
                 ),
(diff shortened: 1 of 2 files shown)


classes/PodsAdmin.php
```diff
@@ -1295,7 +1295,7 @@ public function admin_setup_edit_options ( $pod ) {
                     'label' => __( 'Hierarchical', 'pods' ),
                     'help' => __( 'help', 'pods' ),
                     'type' => 'boolean',
-                    'default' => false,
```
sc0ttkclark: Not just here but also in the create new pod function in PodsAPI.

sc0ttkclark:
Needs some adjustment to move forward.

classes/PodsAdmin.php
JoryHogeveen: I've put it in save_pod(). Please test this as well!

JoryHogeveen:
@sc0ttkclark should that be in save pod? I wrote it in add_pod at first but then noticed that all saves pass through the save_pod so thought this would be a better location.

sc0ttkclark:
In the future we will be moving some of the defaults from add_pod to save_pod, we can move it there at that time. I just didn't want it to get left behind with the defaults that we use in add_pod explicitly.

Issue #3580, opened 2016-06-18
Revision history for pod fields


bustapaladin:
Pretty much what these fine folk here are asking:
 https://wordpress.org/support/topic/add-revision-history-to-custom-fields?replies=2

But then for the pods plugin and the fields it creates. So if someone edits a custom field on a custom post type it tracks the changes somewhere.

Also be able to clear the history and set the limit of how much is stored ,etc..

quasel:
see https://de.wordpress.org/plugins/wp-post-meta-revisions/ maybe that's all you need ;)

bustapaladin:
Does it support custom post types?


From reading this I'm not sure: https://wordpress.org/support/topic/custom-post-types-235?replies=1

How would I now use it for pods plugin?
Like this? 
```
function add_meta_keys_to_revision( $keys ) {
    $keys[] = 'uname', 'suspision', 'other_name';
    return $keys;
}
add_filter( 'wp_post_revision_meta_keys', 'add_meta_keys_to_revision' );
```

JimSRush:
if anyone is looking for a way to achieve this, I've made a [gist](https://gist.github.com/JimSRush/4edb2dc1253d1d7b5ba1e02a6f61db56) that might help someone. Good luck!

jaycollier:
Hi @JimSRush and others on this thread.

I added the gist mentioned above to our functions.php and custom field changes are still not triggering revisions (although title, author, and excerpt do).

I have also tried these plugins, no luck:

https://wordpress.org/plugins/wp-revisions-control/

https://wordpress.org/plugins/wp-post-meta-revisions/

JimSRush:
@jaycollier I'd suggest you start by looking at what's happening in the `$_REQUEST` object - can you see your custom fields in there? They're prefixed by `pods_meta..`, so the gist may be failing at the `isset` step, where it checks to see if they're set.

Also, another thing you could try is double checking that this function is working properly, since this is the bit where it adds the list of fields to check changes against - if it's not working, it won't trigger a revision.
```
function my_revision_fields( $fields ) {
	// return the pod story type and extract the fields
	$story = pods('story');
	foreach ($story->fields as $key => $value) {
		$fields[$key] = $value['label'];
	}
	return $fields;
}
add_filter( '_wp_post_revision_fields', 'my_revision_fields' );
```

In this case, our pod name was 'story' so you'll need to change it to whatever the name of your pod is - apologies I didn't make this clear. I'll update the gist.

[...]

PR #3857, opened 2016-11-07
*Add support for Post Meta Revisions

pcfreak30:
#3580 

This integrates the plugin referenced in #3580 to automatically whitelist all fields.

### DIFF
--- a/classes/PodsInit.php
+++ b/classes/PodsInit.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @package Pods
  */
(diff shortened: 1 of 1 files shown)


JoryHogeveen:
Maybe add this as an advanced field option instead of enabling this for all fields by default?

pcfreak30:
I thought about that but I really did not see the benefit to it.

JoryHogeveen:
When you work with a lot of metadata a lot less is stored this way. This can be a big thing for larger sites/networks.

sc0ttkclark:
We also need to move the bulk of this code into the PodsMeta class. Will have to think about how best to lay out the field options and how to restrict it to Post-only Meta-only.

quasel:
Maybe to look at with 2.8 and the pods object stuff?

sc0ttkclark:
Will have more time for this soon, but the main challenge here is that it's so heavy to go fetch all the Pods involved. I'll have to build a better way to do this without the performance/memory hit.

PR #3876, opened 2016-12-09
Accidentally added a non-printable character to settings-tools.php

szepeviktor:
After `$api->cache_flush_pods();`


szepeviktor:
Excuse me.

PR #3907, opened 2016-12-30
css markup

Ramoonus:
force correct markup


sc0ttkclark:
Need to first verify this won't block any other PRs that change CSS (since this reformats and could force people to redo that manually)

JoryHogeveen:
Why spaces as indentation instead of tabs?

sc0ttkclark:
We can reformat that easily, preference is tab.

Issue #3905, opened 2016-12-30
WP Gallery display options for multiple images relationship fields


PodsBot:
WP Gallery display options for multiple images relationship fields submitted via Slack by keraweb

Output field options:
- [x] Display as WP Gallery
- [x] Gallery image size
- [x] Gallery images per row (columns)
- [x] Gallery image link to
- [x] Gallery random sort order
- [ ] More?

Field class modifications:
- [x] Edit `display()` method to conditionally return a WP gallery instead of the default.

More info:
https://codex.wordpress.org/The_WordPress_Gallery
https://codex.wordpress.org/Gallery_Shortcode
WP function: `gallery_shortcode()` (`wp-includes/media.php`)

PR #3910, opened 2017-01-02
WP Gallery display options for image fields

JoryHogeveen:
See #3905

### DIFF
--- a/classes/fields/file.php
+++ b/classes/fields/file.php
@@ -197,7 +197,8 @@ class PodsField_File extends PodsField {
                 'default' => __( 'Add File', 'pods' ),
                 'type' => 'text'
             ),
-			/* WP GALLERY OUTPUT */
+
+            /* WP GALLERY OUTPUT */
             self::$type . '_wp_gallery_output' => array(
 	            'label' => __( 'Output as a WP Gallery', 'pods' ),
 	            'help' => sprintf(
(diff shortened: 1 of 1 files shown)


jimtrue:
@JoryHogeveen @sc0ttkclark @pglewis Did this get merged into 2.7? Some of the fields additions are in there but the primary option for 'Output this as a WP Gallery' isn't in the 2.7 branch I'm testing on.

jimtrue:
Further comment, I'm getting 'erroneous' issues on one where it's available as an option AFTER adding the new field, but on a pod where the field already existed and was set for Gallery, the options aren't available. It's also doing some weirdness with Auto Templates. I'm going to do a new test on these both with creating brand new, and with taking existing and shifting them over (and yes, I've cleared cache).

sc0ttkclark:
Can you test on latest 2.x stable to confirm the issues don't exist there too? Just so we can rule that out?

sc0ttkclark:
I see the field options in 2.7 too

jimtrue:
yeah I'm testing out a stable clean install. It was so weird it didn't make any sense.

jimtrue:
But 3 different versions:

2.x with fields created and configured and then switching to 2.7
2.7 new addition in 2.7 only
2.x upgrading to 2.7 and switching back to 2.x (same as 1, but making sure that if they were created and upgraded, they still work correctly in 2.x)

jimtrue:
I'm not seeing the oddness where the field wasn't available for selection, but what is happening is the Auto Templates aren't selectable (or it's not saving the selection, only defaulting to the first template in the list) upon saving and that's what was causing the issue I was seeing.

JoryHogeveen:
So it's an auto template issue instead of WP Gallery options? Maybe it has something to do with the new DFV structure?

jimtrue:
Let's keep discussion of this over in #4143 I noticed it because my pods had fields and relationships and were using the gallery feature, but it's definitely an Auto Template issue

PR #3938, opened 2017-01-17
Properly clear cache before running post-save actions in PodsAPI::save_pod_item

### LINKED ISSUE #3917: pods_api_post_save_pod_item returns old field values
I have a feature which, when a post is saved, automatically sets the title of a post. It uses the values of various fields for that record.

E.g let's say the Pod (let's call it OriginalPod) has a couple of fields:

FieldA: Relationship field to some other Pod
FieldB: Number

After save of a record in OriginalPod any called to `$original_pod_just_saved_record->field()` and also `$original_pod_just_saved_record->display()` displays the data saved for the record _before_ the save function. 

Essentially, this means that my feature which automatically updates a records name, after save, needs to be executed/triggered by 2 saves to the record:

* First save - saves the new values into the record
* Second save - updates the title using the values saved (but not available) during first save.

Expected behaviour is that when retrieving the values for `$original_pod_just_saved_record` in the `pods_api_post_save_pod_item` action hook, the values returned in `$original_pod_just_saved_record->field()` & `$original_pod_just_saved_record->display()` should be the same as what was saved for the fields.
i.e - Only a single save should be necessary of my auto-title-update feature.

I have a hunch this may be similar/related to #3801

sc0ttkclark:
Fixes #3917

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -3784,6 +3784,28 @@ class PodsAPI {
         if ( !$no_conflict )
             pods_no_conflict_off( $pod[ 'type' ] );
 
+        // Clear cache
+        pods_cache_clear( $params->id, 'pods_items_' . $pod[ 'name' ] );
+
+		if ( $params->clear_slug_cache && !empty( $pod[ 'field_slug' ] ) ) {
+			$slug = pods( $pod[ 'name' ], $params->id )->field( $pod[ 'field_slug' ] );
+
+			if ( 0 < strlen( $slug ) ) {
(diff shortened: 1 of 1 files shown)


lougreenwood:
@sc0ttkclark - I can confirm this fixes the issue for us, as reported in #3917 - thanks 👍

lougreenwood:
I think this has regressed in 2.x branch, seeing old behaviour has returned and some of our project's tests are failing again because old data is being passed to the hook action callback.

lougreenwood:
False alarm... ignore that, seems it is working correctly, the bug is in our code.

PR #3942, opened 2017-01-18
composer: composer/installers 1.2

Ramoonus:


pglewis:
Latest appears to be 1.3.1, is there a reason we've gone with 1.2.*?

Ramoonus:
Gemnasium says 1.2
currently the composer aims at 1.0

PR #3946, opened 2017-01-19
Add new singular capabilities for taxonomies (WP 4.7)

### LINKED ISSUE #3896: New capabilities for individual taxonomy in 4.7
https://make.wordpress.org/core/2016/10/28/fine-grained-capabilities-for-taxonomy-terms-in-4-7/

Plural capabilities only appear here, I think: 

https://github.com/pods-framework/pods/blob/f9ace/classes/PodsInit.php#L663-L668

JoryHogeveen:
Quick fixes #3896 

For Pods 2.7 (or later) I think we should revisit the capability handling for Pods.
Since WP 4.7 there are two default capabilities for taxonomies: `category` & `tag`, just like with posts (`post` & `page`).
I'll create a new ticket for this (not a high priority issue).

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -2331,16 +2331,18 @@ class PodsAdmin {
                     $capabilities[] = 'delete_' . $capability_type;
 
                     if ( 1 == pods_var( 'capability_type_extra', $pod[ 'options' ], 1 ) ) {
-                        $capabilities[] = 'read_private_' . $capability_type . 's';
-                        $capabilities[] = 'edit_' . $capability_type . 's';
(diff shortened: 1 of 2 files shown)


classes/PodsAdmin.php
```diff
@@ -2331,29 +2331,37 @@ public function admin_capabilities ( $capabilities ) {
                     $capabilities[] = 'delete_' . $capability_type;
 
                     if ( 1 == pods_var( 'capability_type_extra', $pod[ 'options' ], 1 ) ) {
```
JoryHogeveen: Changes post_type code to be similar as what I've done for taxonomies. Consistency FTW!

PR #3955, opened 2017-01-21
codeclimate array fix

Ramoonus:


pglewis:
This is YAML, not JSON... I don't think there needs to be commas at all?

Ramoonus:
in that case i have to remove the comma`s on every line

Ramoonus:
changed

pglewis:
Needs indentation as well.  Tabs not allowed, lets go with 2 spaces to match the stuff below it.

PR #3990, opened 2017-02-07
Allow pod_exists() to work if passed a post_type name

jamesgol:
get_posts() returns an array with the results even if you are only asking for 1 post per page.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -5324,6 +5324,7 @@ class PodsAPI {
                     'post_type' => '_pods_pod',
                     'posts_per_page' => 1
                 ) );
+
                 if ( is_array( $pod ) ) {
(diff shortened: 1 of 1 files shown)


jamesgol:
For some additional context.  If you would call is_pod( $post ); on a WordPress post object (that was actually a post_type of a Pod) it would return false.

classes/PodsAPI.php
```diff
@@ -5324,6 +5324,9 @@ public function pod_exists ( $params, $type = null ) {
                     'post_type' => '_pods_pod',
                     'posts_per_page' => 1
                 ) );
+                if ( is_array( $pod ) ) {
```
sc0ttkclark: I'd prefer that there are blank lines between conditional block sets.

PR #3992, opened 2017-02-09
Only allow admins access to the 'pq_loadpod' AJAX action

jamesgol:
Currently any logged in user can execute this AJAX action which would allow them to return a list of the names of all of the elements on any Pod.

Not a major issue, but should be protected anyways.

### DIFF
--- a/components/Templates/includes/functions-pod_reference.php
+++ b/components/Templates/includes/functions-pod_reference.php
@@ -3,6 +3,9 @@
 add_action( 'wp_ajax_pq_loadpod', 'pq_loadpod' );
 
 function pq_loadpod($podname = false) {
+	if ( !pods_is_admin() ) {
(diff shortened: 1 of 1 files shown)


components/Templates/includes/functions-pod_reference.php
```diff
@@ -3,6 +3,9 @@
 add_action( 'wp_ajax_pq_loadpod', 'pq_loadpod' );
 
 function pq_loadpod($podname = false) {
+	if ( !pods_is_admin() ) {
+		pods_error( __( 'Unauthorized request', 'pods' ) );
+	}
```
sc0ttkclark: Please add a blank line. Sorry to nit pick :)

sc0ttkclark: Nevermind

PR #4000, opened 2017-02-16
Tame the template reference recursion

### LINKED ISSUE #3370: Pods Template ‘Pod Reference’ is Over traversing relationships resulting in ‘stalled’ Template Editor
Pods Template ‘Pod Reference’ is Over traversing relationships resulting in ‘stalled’ Template Editor submitted via Slack by jimtrue

This is one of those 'sometimes' Edge Case, but definitely an issue where the Pods Reference of Fields the Template Editor creates is 'over traversed' to the point of having 4 to 5 or more levels of traversal.

This results in an 'unusable' Pods Reference because no one could possibly want to go that deep into traversal, but also fields like this:

related_tutorials.related_faqs.related_videos.related_faq.related_code.contributors.user_url

In the above Tutorials has a relationship into Faqs and Faqs also has a Relationship into Videos and into Code, but that doesn't means we need to be able to walk across all of those.

jamesgol:
Fixes #3370, #3992
This cleans up the code quite a bit.  Still not perfect and the ordering might need to be adjusted a bit but it will keep everything from exploding.

Related to #3370

Also includes #3992 since it covers the same functions

Replaces #3999 which was a PR from my own fork

### DIFF
--- a/components/Templates/includes/functions-pod_reference.php
+++ b/components/Templates/includes/functions-pod_reference.php
@@ -3,31 +3,19 @@
 add_action( 'wp_ajax_pq_loadpod', 'pq_loadpod' );
 
 function pq_loadpod($podname = false) {
+	if ( !pods_is_admin() ) {
+		pods_error( __( 'Unauthorized request', 'pods' ) );
+	}
 	if(!empty($_POST['pod_reference']['pod'])){
 		$podname = $_POST['pod_reference']['pod'];
 	}
 	if(!empty($_POST['pod'])){
 		$podname = $_POST['pod'];
-
 	}
-	$fields = array('No reference Pod selected');
+	$fields = array( __( 'No reference Pod selected', 'pods' ) );
+
 	if(!empty($podname)){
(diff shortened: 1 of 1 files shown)


jimtrue:
I think 4000 is more auspicious anyway. I can confirm this fixes #3370

components/Templates/includes/functions-pod_reference.php
```diff
@@ -3,31 +3,19 @@
 add_action( 'wp_ajax_pq_loadpod', 'pq_loadpod' );
 
 function pq_loadpod($podname = false) {
+	if ( !pods_is_admin() ) {
+		pods_error( __( 'Unauthorized request', 'pods' ) );
+	}
```
sc0ttkclark: These lines are in a separate PR already, right? Also, please add a blank line

sc0ttkclark: Nevermind, I see the code below has no blank lines so no worries here.

Issue #4012, opened 2017-03-01
Passing array to `esc_js()` due to currency sign changes


pglewis:
https://github.com/pods-framework/pods/blob/9743465/ui/fields/currency.php#L48&ts=4

pglewis:
We're emitting the result directly into the js code, so this results in a Javascript error.

PR #4016, opened 2017-03-02
Fix for #4012

pglewis:
One more time... with feelin'.


pglewis:
@JoryHogeveen : Any other spots that might have been missed?

Issue #4017, opened 2017-03-02
Enable drag-drop for select2 v4.0


pglewis:
PR #4029

Part of 2.7: #3364

PR #4029, opened 2017-03-07
Feature/4017 select2 drag drop

pglewis:
Issue: #4017


pglewis:
Might be ready to merge in.  The specified order is not being preserved on reload but reordered items do appear to be saved in the correct order.  List view is showing the same behavior, so I currently suspect the mis-ordering is happening on the PHP side when assembling the data.

pglewis:
The ordering not being preserved is a display issue; custom drag-drop ordering is saved properly.  This one is ready to go.

Issue #4036, opened 2017-03-08
Move i18n logic to PodsI18n


JoryHogeveen:
Put all language logic in PodsI18n.

- [x] Move `get_current_language()` from PodsAPI (with reference).

PR #4037, opened 2017-03-08
Move & refactor i18n logic

JoryHogeveen:
Issue: #4036 

- [x] PodsI18n is now a true singleton class
- [x] Moved and enhanced `get_current_language()`
- `get_current_language()` now returns the current locale code
- `get_current_language_data()` returns the language object (Polylang or WPML)


classes/PodsI18n.php
```diff
@@ -196,4 +205,240 @@ private static function default_strings() {
...
+	 * @since 2.6.6
+	 * @since 2.7 Moved to this class from PodsAPI
+	 *
+	 * @param array $args (optional) {
+	 *     @type bool $redo Rerun logic?
+	 * }
+	 *
+	 * @return array
+	 */
+	public static function get_current_language( $args = array() ) {
```
pglewis: We have `$current_language` and a `$current_language_data` members, this method returns the language data.  Should the method reflect that with the name `get_current_language_data()`?

classes/PodsI18n.php
```diff
@@ -196,4 +205,240 @@ private static function default_strings() {
...
+
+		$args = wp_parse_args( $args, array(
+			'redo' => false,
+		) );
+
+		if ( ! $args['redo'] && null !== self::$current_language_data ) {
```
pglewis: Do we have something already somewhat standardized for cache-busting or semi-related to use as the key name?  We don't seem to use 'redo' anywhere else.   @sc0ttkclark ?

classes/PodsAPI.php
```diff
@@ -8331,220 +8331,13 @@ public function process_form ( $params, $obj = null, $fields = null, $thank_you
 	 * Get current language information from Multilingual plugins
 	 *
 	 * @since 2.6.6
+	 * @since 2.7 Logic moved to PodsI18n
 	 *
 	 * @return array
 	 */
 	public static function get_current_language() {
```
pglewis: Do we need to leave a proxy around in `PodsAPI`?  I don't see any references to `PodsAPI::get_current_language()` outside of itself so seems like this should just be removed and all calls made directly into `PodsI18n`

pglewis:
Also, any reason we can't keep `$current_language` and `$current_language_data` private?  I'd prefer to leave any access to those through public methods.  

And what about `init()`?  `get_instance()` is needed from the outside, but I think we can get by with `init()` being private?

classes/PodsI18n.php
JoryHogeveen: Sounds good to me, @sc0ttkclark  ??

classes/PodsAPI.php
JoryHogeveen: Can indeed be removed, @sc0ttkclark agreed?

JoryHogeveen:
`init()` can (and IMO should) be private indeed. At the moment both functions do the same...
If we make `init()` private we can also remove the `is_object()` checks and return.

classes/PodsAPI.php
JoryHogeveen: Oh, I remember. The reason I left it there is that it also calls `pods_i18n()` which includes the class if it doesn't exists yet.

JoryHogeveen:
To continue on this.

I'd refactor `ini()` to a private `__construct()`. This way this class is a true singleton class.

classes/PodsI18n.php
JoryHogeveen: Maybe leave `get_current_language()` as it is and add `get_current_locale()`.
And reflect this to the properties as well:
`$current_language` => `$current_locale`
`$current_language_data` => `$current_language`

JoryHogeveen:
4a09cc9
"Make language functions public" should be "Make language functions non-static"

classes/PodsI18n.php
JoryHogeveen: Note for Scott. And the winner is: `refresh`

PR #4047, opened 2017-03-10
Add support for 'find' output type in field()

### LINKED ISSUE #4046: Support field() single Pods object as output type for related pod objects
Support field() Pods set as output type for related pod objects submitted via Slack by sc0ttkclark

- [x] Add new conditional when doing find(), do not return data() directly `if ( 'find' === $output )` https://github.com/pods-framework/pods/blob/2.x/classes/Pods.php#L1311 `$item_data = array(); $data = $related_obj;`

sc0ttkclark:
Fixes #4046


FLHerne:
Tested (I asked about this on Slack, got handed a patch). After the array_values fix it seems to be working perfectly for me.

FLHerne:
Sorry, there is a problem:

Unlike 'pods' or 'array' mode, this doesn't preserve the relationship order in a multiselect.

That makes `pods('question', <n>).fetch(['name' => 'answers', 'output' => 'find'])` equivalent to `pods('answer').find(['where' => 'question.id = <n>'])`, and doesn't work for my use-case.

sc0ttkclark:
@FLHerne can you clarify --

1. Are you talking about the relationship order as in the order of which they values were selected in the form?

2. Is this the only problem you've found?

FLHerne:
Yes, to both.

What I'm using now instead:
```
$relevant = pods('question', $id)->field('answers', ['output' => 'ids']);
if ( count($relevant) ) {
    $joined = join(', ', $relevant);
    $params['where'] = 'id in ('.$joined.')');
    $params['orderby'] = 'FIELD(id, '.$joined.')';
}
else {
    $params['where'] = 'FALSE'; // No matching answers
}
$newpod = pods('answers')->find($params);
```

sc0ttkclark:
I'll work on that order by clause next then, thanks for that.

sc0ttkclark:
@FLHerne this is ready for testing again when you're able to.

sc0ttkclark:
Any update @FLHerne, I'd love to get this out in Pods 2.6.9 coming in the following few days.

sc0ttkclark:
@pglewis did this get wiped after a merge conflict? I don't see the code in the release/2.7 branch for some reason

pglewis:
I don't recall anything like that but wouldn't swear 100% that it didn't happen.

pglewis:
Track revisions for that section, maybe?

Issue #4039, opened 2017-03-08
2.7/feature Add New Option to Enable/Disable


PodsBot:
2.7/feature Add New Option to Enable/Disable submitted via Slack by jimtrue

Part of 2.7: #3364
PR: #4054

- [x] Make a list of other objects that shouldn't show the option (no add new) (holding on #4080)
- [x] Add JS unit test for add new option (@pglewis)
- [x] Option

pglewis:
Needs to have more types excluded still

pglewis:
@sc0ttkclark: I'm not sure of the best way to either include a whitelist or exclude a blacklist of objects that won't allow add new (and thus shouldn't bother showing the option).  Otherwise, I think this one is good to go pending a look over the PR.

sc0ttkclark:
Code looks good here

pglewis:
I don't think we have any way to differentiate ACTs from Pods' Settings Pages by name alone, but that can be a separate issue rather than hold this one up.

PR #4054, opened 2017-03-15
Option to disable "Add New" for pick fields

pglewis:
Issue: #4039

### DIFF
--- a/ui/js/pods-dfv/pods-dfv.min.js
+++ b/ui/js/pods-dfv/pods-dfv.min.js
@@ -1144,7 +1144,7 @@ var PodsDFV = (function () {
   		this.showChildView('list', list);
 
   		// Show Add New?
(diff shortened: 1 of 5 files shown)


ui/js/pods-dfv/pods-dfv.min.js
```diff
@@ -1145,7 +1145,7 @@ var PodsDFV = (function () {
   		this.showChildView('list', list);
 
   		// Show Add New?
-  		if (this.fieldConfig.get('iframe_src') !== '') {
+  		if (this.fieldConfig.get('iframe_src') !== '' && this.fieldConfig.get('pick_allow_add_new') == 1) {
```
sc0ttkclark: We should usually strive to keep comparisons as yoda conditions. But I'm okay if this goes in because the condition before it isn't.

Issue #4056, opened 2017-03-15
I18n items + check and fix all translation strings in Pods for 2.7


PodsBot:
Main 2.7 issue: #3364 

- [x] PR: #4057
- [x] #4032 `__( 'Select from existing' )`

TODO
- [ ] Class files
- [ ] Function files
- [ ] Field type UI
- [ ] UI -> Admin
- [ ] Components
  - [ ] Component comment info, not sure how...

pglewis:
The bulk of what we need to look at is on the JS side but this is a generic bucket for any i18n things we need to cover for 2.7.

JoryHogeveen:
Okidoki!
I think it's better to wait with files that are in progress on other PR's to prevent merge conflicts or missing items later on, agreed?
PHP side is easy and with the PodsI18n class the JS side is now easy as well :).

JoryHogeveen:
Reopening for PHP translations

pglewis:
Where are we sitting on this one?  We're at the final issue sweep for RC.

JoryHogeveen:
I think most of it is done, it just isn't checked yet

Issue #4055, opened 2017-03-15
Add New [item] Label from Related Post Labels for #3364


PodsBot:
Add New [item] Label from Related Post Labels for #3364 submitted via Slack by jimtrue

jimtrue:
Related to #3364 

Label for button should default to 'Item' or just "_blank" but otherwise should pull Related Record(s) Plural Label for Add New [Books] (users, etc.)

pglewis:
Maybe should also apply to "Select From Existing" if we stick with a labelled button as the UI element.  (#4019)

sc0ttkclark:
This is partially done, we can move the adding of customized labels into a future release.

PR #4057, opened 2017-03-15
Pods DFV i18n

JoryHogeveen:
Issue: #4056 

I18n in other PR's
- New string in #4032: 'Select from existing' - commit: 3fa214f

### DIFF
--- a/ui/js/pods-dfv/_src/core/iframe-frame.js
+++ b/ui/js/pods-dfv/_src/core/iframe-frame.js
@@ -1,4 +1,4 @@
-/*global jQuery, _, Backbone, Marionette, wp */
+/*global jQuery, _, Backbone, Marionette, wp, PodsI18n */
 
 /**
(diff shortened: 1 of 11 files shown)


ui/js/pods-dfv/_src/core/iframe-frame.js
```diff
@@ -33,7 +33,7 @@ export const IframeFrame = wp.media.view.Frame.extend( {
 	},
 
 	initState: function () {
-		const title = this.options.title || 'Add New Record';
+		const title = this.options.title || PodsI18n.__( 'Add New Record' );
```
JoryHogeveen: Should "Record" be replaced with the object name? (CPT or Tax singular label)

pglewis: That's it's own issue, #4055

Issue #3446, opened 2016-03-13
Ehancement: Activate Pods Templates on Pods Activation for User Experience


PodsBot:
PR: #4060

Ehancement: Activate Pods Templates on Pods Activation for User Experience submitted via Slack by jimtrue

It's fairly easy to miss this for new & beginning users of Pods. This would be a quality of life improvement. If a developer or anyone else didn't need Auto Templates or Pods Templates, they could very easily disable this component.

jimtrue:
@sc0ttkclark How hard do you think this one would be to implement? When we're getting folks on-ramped with Pods, telling them to go and do this extra step is a bad bit of UX. If it was already there and activated, they'd be able to get rolling with Pods quickly.

pglewis:
I'm culling through 2.7 milestones on a mission, so we need to decide if this happens or punt.  I don't think it's difficult, not sure if there are repercussions of changing a default behavior.

jimtrue:
there are no repercussions as the goal is this would only happen on initial installs. After that, if someone wanted to disable Templates, it should recognize they were disabled. Of course, I don't know if we have a setting for the initial 'stuff' but this particular component should be enabled by default as it just leads to confusion and reminding people to 'do that step'

pglewis:
What about multi-site?  Should we exclude multi-site installs as being outside the target audience?

sc0ttkclark:
Done, also works for multisite

jimtrue:
How are you testing for this @sc0ttkclark 'cause this isn't working for me. I've got a local installation brand new fresh and clean, install and activate Pods and no Auto Enabled Templates.

sc0ttkclark:
Will look again at install/upgrade procedure to see where this might not be working as expected.

PR #4059, opened 2017-03-17
Make templates component active after an initial installation or re-installation

### LINKED ISSUE #3346: Ability to add filters inline in the Pods edit UI
It would be handy to have the ability to edit simple filters and actions from within the Pods edit screen.

For example, in a field, after the REST tab you could have a Filters tab. You could have a text area to enter PHP code for each filter type. Alternatively you could have a drop down of filters and choose to add or remove edit boxes.

This would be handy because often I only need a very small amount of code in a callback, for example to update a post title. If it was integrated into the Pod editing, I could export the whole Pod and import it without worrying about other files.

Alternatively, you could have a separate screen for editing all filters and actions in one place. It makes a complete solution without resorting to editing theme functions or creating a plugin.

pglewis:
Dyslexia

### DIFF
--- a/classes/PodsInit.php
+++ b/classes/PodsInit.php
@@ -79,6 +79,13 @@ class PodsInit {
 	static $upgrade_needed = false;
 
 	/**
+	 * Whether setup() was called (activations for new installation or re-installation only)
+	 *
+	 * @var bool
+	 */
+	private static $did_setup = false;
+
+	/**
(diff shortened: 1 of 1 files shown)


jimtrue:
Nope doesn't work; see note in #3446 

>How are you testing for this @sc0ttkclark 'cause this isn't working for me. I've got a local installation brand new fresh and clean, install and activate Pods and no Auto Enabled Templates.

PR #4060, opened 2017-03-17
Make templates component active after an initial installation or reinstallation

pglewis:
Issue: #3446


jimtrue:
@pglewis Just cloned this down and switched to this branch on a brand new installation of WordPress. Upon activating Pods in the plugin list, Templates Component was not enabled. :(
Should I test this another way?

pglewis:
I'll have to setup and test a fresh, never been installed site.  Worked perfectly on a reinstall using the same method.

jimtrue:
Let me test that. On a reinstall Pods would already be in the 'database', right? just deactivated?

sc0ttkclark:
Templates is not enabled by default on new installs, it was auto-enabled from 1.x upgrades only. Otherwise it's manual.

pglewis:
This is a feature request to make them auto-enabled on new installs only.  This is a present for Jim because it's a repeated support headache.

pglewis:
Now with less "not working"

sc0ttkclark:
This forces template component to always be enabled/included.

In the 1.x > 2.0 upgrade code, we toggle the component on after updating successfully:
https://github.com/pods-framework/pods/blob/2.x/sql/upgrade/PodsUpgrade_2_0_0.php#L925-L927

I think we could make use of similar code during the PodsInit install at the end of the empty $pods_version check, which indicates a fresh install:
https://github.com/pods-framework/pods/blob/2.x/classes/PodsInit.php#L1116

jimtrue:
Was this merged into 2.x? I see the 2.7 branch merge, but not the 2.x one.

jimtrue:
Also, as a note, I'm on 2.7 and it didn't work on a database reset on first activation of Pods.

PR #4061, opened 2017-03-17
Hidden fields don't save Default Values

### LINKED ISSUE #3996: Hidden fields don't save Default Values
1) Created custom post type
2) Added new text field, set default value and hid from UI

When adding a new post, html source shows hidden field without a value attribute

Post meta record in database does contain the default_value

3) Changed field to no longer hide from UI

When adding a new post, the field does contain the default value.

Dave

pglewis:
Fixes #3996


pglewis:
This is intentionally against 2.x; shouldn't have to wait until 2.7.

PR #4084, opened 2017-03-24
Fix in_array() var position

sc0ttkclark:


jimtrue:
Tested with user in #support on Slack who was receiving the error so had the proper test case. Confirmed merge from that review.

Issue #4087, opened 2017-03-25
Merge Pods Template improvements into 2.x, as well


pglewis:
These need to make the next release

PR #4088, opened 2017-03-25
Feature/4087 merge templates improvments into 2.x

pglewis:
#4087


pglewis:
@jamesgol: I think I got all the related commits that went into 2.7 but not 2.x.  Let me know if there are any I missed that should be included or if there are any probs with merging these pieces.

Issue #4105, opened 2017-04-09
Custom date format input for date and date/time fields


PodsBot:
Custom date format input for date and date/time fields submitted via Slack by keraweb

sc0ttkclark:
Added as part of #4106

Issue #2208, opened 2014-05-30
Feature Request: Date Format


asap247:
Hi,

Simple feature request:  Can you make the following date format for the pods date select field.

Month Year

That's all.

szepeviktor:
Is it possible to default to WP's data format?

jimtrue:
Related #4105

jimtrue:
Why was this particular issue closed? An easier to use date filters format input screen would be a very good thing.

jimtrue:
Even WordPress Core has it:

![screenshot 2017-04-10 06 34 37](https://cloud.githubusercontent.com/assets/3052875/24857887/d5fc0498-1db7-11e7-883d-8413cd528342.png)

JoryHogeveen:
@jimtrue I'm already working on this. One problem I'm running into is that the backend UI (jQuery datepicker) doesn't handles formats differently and might nog support all PHP date() options.

WP only has this for front-end output (so PHP only).

jimtrue:
Weren't we looking for some better datepickers anyway? I know we use HTML5 as an option (which I think turns off the jQuery datepicker)

jimtrue:
I'd also like to confirm that by default the date output that we kick out matches the WP General Settings one.

sc0ttkclark:
Added this ability as part of #4106

PR #4106, opened 2017-04-10
Enhance Date/Time fields - Class inheritance & WP defaults + custom options

### LINKED ISSUE #3829: [if] Template Tags do not recognize empty date fields as empty
[if] Template Tags do not recognize empty date fields as empty submitted via Slack by jimtrue

JoryHogeveen:
Issue: #4105 
Related: #2208 , #4107

- [x] Use `datetime` as main class and extend it for `date` and `time` fields.
- [x] Create a converter script for JS formats (no need to have this logic in the form field file)
- [x] Option to use WP Default format in both backend and front end
- [x] Option to create your own format
  - _Be aware that jQuery UI doesn't support all formatting that PHP does!_
- [x] Backwards compat for existing PHP and JS filters
- [x] Optional custom format input for jQueryUI formats
  - Custom JS (field input) format needs to be compatible with `strtotime()` after conversion to PHP date() format.
- [x] Fix #3829
- [x] Tests: functionality + backwards compatibility @jimtrue 
- [x] Convert file indentation to Tabs once PR is ready @JoryHogeveen

**Feedback request:**
- [x] Review option names
- [x] Review help texts
- [x] (Date/Time input only, no format validation possible currently) Maybe add format validation? @sc0ttkclark 
  - Can be done with `strtotime()` or `DateTime::createFromFormat()`
- [x] (Do both) _Maybe only allow custom formats for front-end?_ @jimtrue @sc0ttkclark 
  - Currently the code doesn't work when adding any kind of text or non-date characters because using the datepicker it returns a formatted string instead of a timecode.
- [x] DateTime field: Allow custom input for date/time combined instead of separate? @sc0ttkclark @jimtrue (NOPE)


JoryHogeveen:
@sc0ttkclark Tests are failing though

sc0ttkclark:
I've merged latest into the branch from 2.7, let's see if tests are passing now or if there are issues to resolve. I can't imagine what would cause any of the stuff to fail here from the changelog, so it might have been something in 2.7 from what i can tell.

sc0ttkclark:
Yeah looks like we do have some issues with testing the date/datetime/time fields. How does this test out when using normally right now? Notice any issues with saving data or getting it from field() or display()?

JoryHogeveen:
Nope, didn't see any errors. I also have no idea what these tests do that are failing..

sc0ttkclark:
They are using the old field options, so it's possible there's a back-compat issue here

sc0ttkclark:
* We need to test this on a site that has Pods 2.6.8 on it and has an existing pod with an existing date field added to it.
* We'll want a shortcode to output the field value on a demo page.
* Then we upgrade Pods to this feature https://github.com/pods-framework/pods/archive/feature/4105-custom-datetime-options.zip
* Without modifying any pods or fields, go to the demo page and then confirm the date is showing up the same as when running Pods 2.6.8

JoryHogeveen:
@jimtrue 
When you have the time, can you do some tests? (See above)

jimtrue:
@JoryHogeveen I'll do my best to get that test done today or tomorrow. It would help to have a list of date formats to test against. I don't do that much with them myself, so we should probably keep a string of them handy to validate against.

JoryHogeveen:
@jimtrue Main thing is the upgrade from the current stable version. Everything should keep working the same.
Second thing is indeed making sure the existing and new options work. Custom strings have limitations but after saving a post Pods will generate an error is the config isn't correct.

jimtrue:
@JoryHogeveen @sc0ttkclark Per the notes above, those tests passed.

Ready to test modifying existing fields and date/time strings

jimtrue:
Those were using default date/time configurations, though

jimtrue:
@JoryHogeveen Instead of the PHP Date & Time Formatting, should we maybe reference them to the WordPress Codex one? https://codex.wordpress.org/Formatting_Date_and_Time (and/or both?)

jimtrue:
"Predefined format" does that mean the format that was initially in the date/time, date or time input field? Some help text on that one would help:

"Predefined format" - Former version of Pods format, please choose WordPress Default or Custom?

jimtrue:
Sorry Predefined would be 'select from a list of available formats' got it. WordPress default would mean it's defined under the General, Settings for the WordPress Installation and Custom would mean it would be defined using the PHP schema from the docs for just this field.

jimtrue:
@JoryHogeveen @sc0ttkclark I think we should set the DEFAULT field format for date, date/time and time on NEW fields to WordPress default. That would ensure the most consistency with what they did the work on.

jimtrue:
Nevermind! @JoryHogeveen already did that ;)

jimtrue:
@JoryHogeveen @sc0ttkclark Everything looks really good from my tests. I tend to test your custom string to php thingie, but that's the only thing outstanding. Flipping branches didn't break anything and the new field options look really good. Review my help text suggestions above, but other than that this should be good to go.

JoryHogeveen:
> Instead of the PHP Date & Time Formatting, should we maybe reference them to the WordPress Codex one? https://codex.wordpress.org/Formatting_Date_and_Time (and/or both?)
Not sure, the documentation info is the same and I think most (novice) users will user WP default or a predefined format.

> Predefined would be 'select from a list of available formats'
Yep! Maybe a label change?

> WordPress default would mean it's defined under the General, Settings for the WordPress Installation
Yep! Maybe add a link to that setting page in s help tooltip somewhere? Or do you think this is clear enough?

Let me know if you find anything else! Great to hear all migration tests worked at your end as well.
Now that I think of it. Maybe also test ACT's? Since they fully rely on Pods fields for all data. @sc0ttkclark could you take a look at that? I think you know ACT's the best "under the hood".

JoryHogeveen:
Also, Travis tests are still failing for some reason, I get the feeling it has something to do with ACT's..

jimtrue:
yeah I'm terribly inexperienced with the ACTs.

This is what I'd do on the Help Text just for Date Field Format (since there's nothing there right now) and likewise on Time:

> **WordPress Default** is the format used in [Settings, General]([site-url]/wp-admin/options-general.php) under 'Date Format'. **Predefined Format** will allow you to select from a list of commonly used date formats. Custom will allow you to enter your own using the PHP Date/Time Strings. (link).

JoryHogeveen:
@sc0ttkclark Want me to do a full indentation refactor? These files had a big refactor so why not..

Issue #4020, opened 2017-03-02
Prototype view updates from JSON for 2.7 add new


pglewis:
Related: #3309

Part of #3364

pglewis:
No PR involved, this is just an example of returning the data and refreshing the view.  

https://github.com/pods-framework/pods/commit/0a02fbc

sc0ttkclark:
I've got everything in there for posts, terms, users, and pod forms. The only thing remaining there is to handle the JS saving process that terms have which is unique to any of the others. In the future there will be more JS things in core so it'll have to be less than hacky.

That's all that's left though.

sc0ttkclark:
Edit links can be used for modals too now, just need to flesh out onChildviewEditClick in _src/pick/pick.js and add the event handler on the edit links in that modal.

sc0ttkclark:
Needs PR and merge into 2.7

sc0ttkclark:
https://github.com/pods-framework/pods/tree/feature/4020-view-update-prototype

Issue #4019, opened 2017-03-02
Clean up "list view" for relationships (select from existing, dashicons, etc)


pglewis:
PR: #4032 
Part of 2.7: #3364

#### General 
- [x] support translation - 3fa214f
- [ ] Maybe add an option in the field setting to set a icon overwrite (dashicon)?
- [x] Use dashicons
- [x] Set the icon to the object type (of possible from the registered object)
- [x] UI element to initiate a selection list
- [x] selection list that shows/hides appropriately

#### CSS Tweaks
- [x] List view vertical alignment
- [x] Icon styles

#### Up for discussion
- [ ] Put buttons `select from existing` and `add new` side-by-side
- [ ] Auto-update button `add new` to the add new label of that object type (if available)

pglewis:
A rudimentary solution is in place: "Select from existing" toggles between list view and checkbox view (which allows selecting new records).  Would need some CSS love and possibly fixed sized containers so things don't move around so much, but it is usable now.

jimtrue:
Comment on the Select from Existing, when you flip back and forth between the 'selected' view and the Select from Existing, it's a little jarring because the line heights/icon heights are not the same. I think it'd be smartest to make the icons all the same height/width ratio, line-heights consistent for the draggable icon, the icon for the post-type/featured image, and the 3 'action ones' to the right 'edit or delete'; I'd switch that from a 'link' icon to a 'view' icon if that's opening a link to the front-end version of the post. I was checking but there seems to be some inconsistency, with some 25 x 32 blocks and a 40 x 32 block. 
![screenshot 2017-03-15 19 17 43](https://cloud.githubusercontent.com/assets/3052875/23974891/1d991c4a-09b4-11e7-9f89-25454980243e.png)

pglewis:
There's also a brainstorming idea to add a new option for list view that allows you to choose the "select from existing" UI.

PR #4108, opened 2017-04-10
Work on modal communication from add new

sc0ttkclark:
Issue: #4020 
Works on add new AND edit, we should probably try opening edit links in the modal too.

Later:

- [ ] Add new terms needs extra JS work to integrate with


JoryHogeveen:
@sc0ttkclark Please also check #4019 to prevent conflict issues.

PR #4118, opened 2017-04-17
Add support for Comment traversal in field and find

sc0ttkclark:
Now we can traverse post comments in field and find (only supported by Post types)


sc0ttkclark:
Example: `comments.comment_content LIKE "%some content%"`

Works the same as taxonomy traversal: Example: `category.name LIKE "%some term%"`

PR #4138, opened 2017-04-27
Modal Add/Edit for ACTs

### LINKED ISSUE #4137: Modal Add/Edit for ACTs
Main 2.7 issue: #3364 
PR: #4138

- [x] ACT saves are ajaxed, so we should just directly return the JSON rather than an HTML script tag

pglewis:
Fixes #4137


pglewis:
That was not the only fix needed for this issue

PR #4139, opened 2017-04-27
Handle modal add/edit for ACT ajaxed requests

### LINKED ISSUE #4137: Modal Add/Edit for ACTs

pglewis:
Fixes #4137


pglewis:
_This_ one should fix the modal editing needs review.

PR #4164, opened 2017-05-10
CSS: Metabox Form field borders + DFV list form field css changes

saltnpixels:
Added lines between pod fields. Fixed scrollable View lists borders.


saltnpixels:
I added line spacers between fields:
screenshot 2017-05-10 10 46 55

Then I noticed an issue with scrolling view lists and borders and worked on fixing it:
screenshot 2017-05-10 11 39 33

PR #4165, opened 2017-05-10
Correct usage of single and double quotes

quasel:
Pod Reference in the Template admin does output {$prefix} in Plain Text due to incorrect ' usage!


pglewis:
Do you know if this is pre-existing in 2.x?  If so, we should fix it there and merge it up to 2.7.

PR #4183, opened 2017-05-15
Follow REST API meta key visibility for exported objects

### LINKED ISSUE #3919: REST API does not honour field visibility when indirectly querying a Pod via a related Pod
If I have 2 pods, linked by a relationship field.

If one Pod has global read-write access to it's fields, and the second, related Pod selectively does not enable read access for various fields over the WP JSON REST API. If the linking relationship field returns Full Field with 1 or more levels of depth, fields which have not had read-access enabled will be returned.

Expected behaviour is that if a field is not explicitly enabled/exposed as readable via the JSON REST API should not be returned when querying other pod related to this field.

To be clear, this is only an issue when querying another Pod related to the Pod with the selective, non-readable fields - read/write access works fine when accessing a Pod directly.

sc0ttkclark:
Fixes #3919

### DIFF
--- a/classes/Pods.php
+++ b/classes/Pods.php
@@ -3014,7 +3014,8 @@ class Pods implements Iterator {
 			'id' => $id,
 			'fields' => null,
 			'depth' => 2,
-			'flatten' => false
+			'flatten' => false,
+			'context' => null,
 		);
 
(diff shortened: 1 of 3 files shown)


sc0ttkclark:
cc: @lougreenwood

Issue #4188, opened 2017-05-16
Enhance PodsI18n JS object with basic sprintf and _n functionality


JoryHogeveen:
PR: #4189

PR #4189, opened 2017-05-16
PodsI18n JS sprintf and _n support

JoryHogeveen:
issue: #4188 

**Covers `%s`, `%1$s`, `%d` and `%1$d`.**
_Note: Doesnt take multiple arguments but an array of arguments (in order)._

```js
PodsI18n._s( 'Test %s string', [ 'yay!' ] )
```
> Test yay! string

```js
PodsI18n._s( 'Test %s string number %d', [ 'yay!', 2 ] )
```
> Test yay! string 2

```js
PodsI18n._s( 'Test %1$s string %2$s', [ 'yay!', 'woohoo!' ] )
```
> Test yay! string woohoo!

```js
PodsI18n._s( 'Test %2$s string %1$s number %1$d', [ 'yay!', 3.1, 'whoohoo!' ] )
```
> Test whoohoo! string yay! number 3.1

**Does not work:**
- Other sprintf functionality (http://php.net/manual/en/function.sprintf.php)
- Multiple `%s` (also not recommended, nor WP coding standards!)


JoryHogeveen:
@pglewis I chose `_s` as the method name because it's short. I'll leave it up to you if you want it to be `__sprintf` for example.

pglewis:
I think I'll see if I can give you some tests so we can safely refactor with confidence.

pglewis:
I brought aboard a `sprintf` implementation with significant adoption and their own test suite rather than roll our own and re-invent the wheel.

Issue #4207, opened 2017-05-17
Currency/Number Field is not Blank but 0.00 (display) so Fails in [if][/if] Template


PodsBot:
PR: #4209

Currency Field is not Blank but 0.00 (display) so failed in [if][/if] Template submitted via Slack by jimtrue

jimtrue:
Per our discussion @JoryHogeveen and referenced in this forum post:
http://pods.io/forums/topic/if-not-equal-to-statement-in-pod-template/

JoryHogeveen:
Same problem with the Number field.

jimtrue:
Don't know if we have a separate issue for number, but I'd suggest tackling them together

PR #4209, opened 2017-05-17
Number/Currency field enhancements

JoryHogeveen:
Related: #4207 

Similar to what we've done with the date/time fields.

- [x] Late static binding
- [x] Currency field extends Number field (remove duplicate code)
- [x] Put thousands/dot formatting logic in separate method and re-use it
- [x] `value()` method to prevent returning values that should be empty
- [x] Also refactors old `pods_var()` calls in the number field
- [ ] Test and review
- [ ] Convert file indentation to Tabs once PR is ready @JoryHogeveen

### DIFF
--- a/classes/fields/number.php
+++ b/classes/fields/number.php
@@ -162,31 +162,6 @@ class PodsField_Number extends PodsField {
 	}
 
 	/**
-	 * Change the value of the field
-	 *
-	 * @param mixed|null  $value
-	 * @param string|null $name
-	 * @param array|null  $options
-	 * @param array|null  $pod
-	 * @param int|null    $id
-	 *
-	 * @return mixed|null|string
-	 *
-	 * @since 2.7
-	 */
-	public function value( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
-
-		if ( pods_v( static::$type . '_allow_empty', $options, 1 ) &&
-		     in_array( rtrim( $value, '0' ), array( '0.', '0,' ) )
-		) {
-			$value = '';
-		}
-
-		return $value;
-
-	}
-
-	/**
 	 * Define the current field's preparation for sprintf
 	 *
 	 * @param array $options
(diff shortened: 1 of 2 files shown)


classes/fields/number.php
```diff
@@ -137,68 +137,76 @@ public function options () {
...
+	 * @param array $options
+	 *
+	 * @return string
+	 * @since 2.0
+	 */
+	public function schema( $options = null ) {
 
-        if ( $length < 1 || 64 < $length )
-            $length = 64;
+		$length = (int) pods_v( static::$type . '_max_length', $options, 12, true );
```
sc0ttkclark: Thank you for the pods_v stuff here, it helps.

We call pods_var in so many places, it has an extra sanitization call that's not needed in many cases. So converting to pods_v will help us long term too.

sc0ttkclark:
> Store handling (pre_save()) for empty values. Return 0.00 or null/''?

If we break backcompat, we don't want that. What do you see here for potential backcompat issues?

JoryHogeveen:
> What do you see here for potential backcompat issues?

The fix is already in `value()` so even existing "empty-but-non-empty" values should return an empty string with that fix.
As for backwards compat. I'n not sure how ACT's handle this kind of data. I know we can safely store an empty string in metadata, but what about ACT's tables with fixed data types?

sc0ttkclark:
ACTs / table-based pods will all store the values as 0 or 0.00 or whatever the decimal format is for the table column.

jimtrue:
Number fields, 0, is being recognized as empty by `[if number_field][/if]`: Yay!
Currency fields, 0.00, are still not recognized as empty by `[if currency_field][/if]`: Sad.

JoryHogeveen:
@jimtrue 
Since the results are different. What happens if the Number field decimals setting is set to more than 0 (without soft formating enabled)

jimtrue:
@JoryHogeveen Once you add decimals, the number no longer represents as empty. As noted, this might be in the `[if field]` handling and might have nothing to do with the number itself, if it passes 'empty' check on 0.00

Issue #4107, opened 2017-04-10
Update jQuery Timepicker


PodsBot:
Update jQuery Timepicker submitted via Slack by keraweb

JoryHogeveen:
https://github.com/trentrichardson/jQuery-Timepicker-Addon/releases

sc0ttkclark:
@JoryHogeveen was this solved as part of #4106?

JoryHogeveen:
@sc0ttkclark 
Nope, see here: #4224

pglewis:
Should be closed via #4224 merge?

JoryHogeveen:
Yep!

Issue #3429, opened 2016-03-02
Please add support for date/time picker localization


S8nsick66:
I find no way get the date picker/time picker in another language than English.

Translations are available though, just needs to be implemented in Pods:
https://jqueryui.com/datepicker/#localization

Pods 2.6.1
WP 4.4.1

pierluigiabramo:
Solution? Thanks!

jimtrue:
@pierluigiabramo no workaround as of yet; this is considered an enhancement.

pierluigiabramo:
Thank's Jim, so expected in 2.6.x ? or ?

jimtrue:
@pierluigiabramo you'll see that label 'needs contributor'. That means we have no one on the Core Team that can tackle that one right away and it's been out there (untranslated) since we added the DatePicker. Features/Enhancements unfortunately don't get as much weight in our current roadmap and we often have to rely on external contributors to pick up and help in those areas. We have a very specific roadmap for the next few releases which are taking extensive coding cycles. We hit the hard-hitting bugs and what's on the roadmap first and if we can get the cycles, it might go into one of the later releases. It won't be on 2.6.x or 2.7 unless we can find someone willing to jump on this one.

JoryHogeveen:
Hi @pierluigiabramo ,

I'm busy with a l10n component for Pods and have included this enhancement. Will keep you up to date!

JoryHogeveen:
@jimtrue @sc0ttkclark  I can also build this within the Pods core if you would prefer this. At this point I'm getting the translation data from the jquery-ui github:
https://rawgit.com/jquery/jquery-ui/master/ui/i18n/datepicker-LOCALE.js

JoryHogeveen:
First dev for the component can be found here: https://github.com/JoryHogeveen/pods/tree/component_i18n

Based on 2.x tree. Added files:
/components/I18n/I18n.php

JoryHogeveen:
https://make.wordpress.org/core/2016/07/06/jquery-ui-datepicker-localization-in-4-6/

JoryHogeveen:
@sc0ttkclark: @pglewis 
This is in the i18n component (planned for 2.7) at the moment. Want to keep it there or move it to the date/time field itself? 
I'd say the field is more logical plus we can patch it over to the next minor release aswell with a check for WP 4.6.

JoryHogeveen:
Diplicate of #1363
@sc0ttkclark same for this one.

Issue #1363, opened 2013-05-08
jquery Date calendar : not beeing translated


YammYcoding:
Can I have this calendar translated to current wp language as well ? If it's supposed to work, it's not working for me. ( in any of my pod installs )

If it still needs the translation typing, please tell me where is the code for it, that I get it done for you...

[]s

JoryHogeveen:
Related: https://github.com/pods-framework/pods/issues/3429

JoryHogeveen:
Fixed in #3641
@sc0ttkclark We might want to put this in Pods core instead of a component. You can close this issue if you think the component is fine.

pavelevap:
Datepicker (no time) localization is part of WordPress from version 4.6: https://core.trac.wordpress.org/ticket/29420

JoryHogeveen:
Hi @pavelevap 
Ah ok! @sc0ttkclark Shall I remove it from the

PR #4224, opened 2017-05-18
Update jQuery Timepicker to v1.6.3

JoryHogeveen:
Issue: #4107 
Info: http://trentrichardson.com/examples/timepicker/

- [x] Update Timepicker plugin (basic)
- [x] Full update of the library + sliderAccess and i18n addons.
- [x] Fix localization issue for time/date pickers
  - Related: #3429, #1363
  -  [x] WP localization since 4.6 working now!
- [ ] Testing.

I already had this running locally for a while, didn't find any issues.


sc0ttkclark:
Let's try local for now

JoryHogeveen:
I'll try to implement WP's solution (WP 4.6+) so we can ditch all datepicker files in general.

JoryHogeveen:
Woot! it works. I don't realy like the nl_NL implementation from WP but that is not our problem. We should keep WP solutions as default imo.

JoryHogeveen:
Alright. This one is ready for testing/review. @jimtrue @sc0ttkclark

classes/PodsInit.php
```diff
@@ -287,17 +287,28 @@ public function register_assets() {
 		wp_register_script( 'pods-codemirror-mode-html', PODS_URL . 'ui/js/codemirror/mode/htmlmixed/htmlmixed.js', array( 'pods-codemirror' ), '4.8', true );
 		wp_register_script( 'pods-codemirror-mode-css', PODS_URL . 'ui/js/codemirror/mode/css/css.js', array( 'pods-codemirror' ), '4.8', true );
 
+		if ( ! wp_script_is( 'jquery-ui-slideraccess', 'registered' ) ) {
+			// No need to add dependencies. All managed by jquery-ui-timepicker.
+			wp_register_script( 'jquery-ui-slideraccess', PODS_URL . 'ui/js/timepicker/jquery-ui-sliderAccess.js', array(), '0.3' );
+		}
+		// Localization enqueued within the fields display files.
```
sc0ttkclark: Is this block needed?

JoryHogeveen: Not anymore since we enqueue the localization files separately. This is one big file with all of them, but without autoloading the defaults.
I left it here for reference but it can safely be removed.

classes/fields/datetime.php
```diff
@@ -831,4 +831,96 @@ public function convert_format( $source_format, $args = array() ) {
...
+			     function_exists( 'wp_localize_jquery_ui_datepicker' )
+			) {
+				wp_localize_jquery_ui_datepicker();
+				$done[] = 'date';
+			}
+
+			if ( in_array( 'time', $types, true ) && ! in_array( 'time', $done, true ) ) {
+
+				// Local files.
+				if ( ! file_exists( PODS_DIR . 'ui/js/timepicker/i18n/jquery-ui-timepicker-' . $locale . '.js' ) ) {
```
JoryHogeveen: @sc0ttkclark Because of this file_existence check we might even want to remove the whole `get_locale_jquery_ui_i18n()` method. Not sure how this would impact performance though..

JoryHogeveen:
@sc0ttkclark  Let me know if you see other improvements, otherwise you can merge (optionally with the i18n comment block removed, your call).

sc0ttkclark:
If @jimtrue or @pglewis signs off on this after testing, let's merge

PR #4259, opened 2017-05-24
Handle duplicating differently for merging of array data

### LINKED ISSUE #4003: Additional CSS Classes not copied when Pod duplicated
When a Pod is duplicated using Pods Admin, the values of field 'Additional CSS Classes' in tab 'Advanced' are not copied to the newly created Pod.

sc0ttkclark:
Fixes #4003


pglewis:
This one is still against 2.x at the moment.  

This nearly works, and I don't understand why it doesn't work 100%, but there is definitely at least one case where options aren't getting saved from 'Additional Field Options'.  I'll gather more details next time I circle back around.

pglewis:
Minimal package: https://gist.github.com/pglewis/e886ee29b0a43fbf7dcbebde6deed996

- Has allow shortcode and allow html enabled
- Has the class "additional" for Additional CSS classes

Duplicate works on this branch, all the options above are copied.  

Once you duplicate it'll take you to the edit form for the new duplicated pod.  Changes to allow shortcode and/or html enabled are not saved.  Does not happen on 2.x.

pglewis:
Also, I'm now wondering if this is because it's against 2.x.  I need to to double-check this in 2.x, not just 2.7.

pglewis:
Still awaiting my review, but needs to be against 2.7

sc0ttkclark:
Ready for CR

pglewis:
> Once you duplicate it'll take you to the edit form for the new duplicated pod. Changes to allow shortcode and/or html enabled are not saved. Does not happen on 2.x.

This was introduced in the PR somehow.  I cannot uncheck either of those two options as they always revert to being checked after saving in the duplicated pod.  Doesn't happen on the original Pod that was duplicated, clearing cache has no effect, doesn't happen on 2.7.

PR #4265, opened 2017-05-26
travis-ci: remove hhvm

Ramoonus:
https://make.wordpress.org/core/2017/05/25/hhvm-no-longer-part-of-wordpress-cores-testing-infrastructure/


sc0ttkclark:
Was just thinking about doing the same thing from reading that yesterday.

PR #4275, opened 2017-05-31
Remove CodeClimate integration

Ramoonus:


JoryHogeveen:
No CodeClimate?

Ramoonus:
it doesn`t work automatically

JoryHogeveen:
Why not? We only need to create an account and it will.

Ramoonus:
if Scott could try to active it on the CC site

Ramoonus:
you don`t need the files to have testing run

PR #4282, opened 2017-06-02
Fix term update when sending just term id and name

### LINKED ISSUE #4281: Term name not being updated with form
Hello,

I’m having one issue trying to update the name of an existing term.

I’m creating a form to update the name of my custom taxonomy like this:

```
$queried_object = get_queried_object();
$componente = pods('icc_componente', $queried_object->term_id);
$fields = array('name' => array('label' => 'Nombre:'));
$formhtml =  $componente->form($fields);
echo $formhtml;
```

The form is showing properly but, unfortunately, when pressing save button the changes in the name are not persisted.

I’ve debugged a little bit and I’ve arrived at this piece of code:

```
if ( empty( $term_data['term_id'] ) ) {
  $term_name = $term_data['name'];
  unset( $term_data['name'] );
  $term_data['term_id'] = wp_insert_term( $term_name, $taxonomy, $term_data );
} elseif ( 2 < count( $term_data ) ) {
  $term_data['term_id'] = wp_update_term( $term_data['term_id'], $taxonomy, $term_data );
}
```

I’m updating only one field: name and $term_data contains only 2 keys: ‘name’ and ‘term_id’. Therefore condition count($term_data)>2 is not met and my term is not updated.

If I change my form creation to:

```
  $fields = array('name' => array('label' => 'Nombre:'),
    'slug' => array('hidden' => true));
  $formhtml =  $componente->form($fields);
```
then $term_data contains three keys (‘name’, ‘term_id’ and ‘slug’), the condition is met and my term is updated.

I wonder if this is a bug or maybe there’s a different way to create a form with this purpose. Your thought will be much appreciated.

Thanks.

David

sc0ttkclark:
Fixes #4281


sc0ttkclark:
Ready for CR

Issue #4313, opened 2017-06-17
Saving multi-file upload fields causes Array to string conversion in classes/PodsAPI.php on line 3754


mgratch:
First reported: https://wordpress.org/support/topic/error-when-using-multiple-images/
I was able to reproduce locally on a fresh local VVV install.
nothing but pods installed.
1 post type added (artwork)
1 field added – file (slide show images)
only setting change was to set it to multiple files

I ensured that define('WP_DEBUG', true);
Added 4 images to slide_show_images
debug log displayed the notice

sc0ttkclark:
Fixed via #4314

PR #4314, opened 2017-06-17
Fix: multi-file fields save causes array to string conversion notice.

mgratch:
#4313 
I added a function `array_filter_walker` and a few extra steps to ensure backwards compatibility.  This may be unnecessary but since the multidimensional array is generated by the name properties in the fields HTML it seemed prudent not to affect that for the sake of making this a quick fix.  Everything should perform as expected now, regardless of the array structure.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -8627,7 +8627,7 @@ class PodsAPI {
     }
 
 	/**
-	 * Remove multidimensional array values that are equal to false.
+	 * Filter an array of arrays without causing PHP notices/warnings.
 	 *
 	 * @param array $values
 	 *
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
@pglewis Can you review this PR with the changes I've made?

classes/PodsAPI.php
```diff
@@ -8619,4 +8625,39 @@ public function __call ( $name, $args ) {
...
+	 * Filter an array of arrays without causing PHP notices/warnings.
+	 *
+	 * @param array $values
+	 *
+	 * @return array
+	 *
+	 * @since 2.6.10
+	 */
+	private function array_filter_walker( $values = array() ) {
+
```
JoryHogeveen: Maybe verify that the it's an array before running `foreach`?

sc0ttkclark: That sounds good, but it's not verifying since it only runs on an array where it's called, would be open to adding a check but not sure it's entirely necessary since it's private

JoryHogeveen: Did a simple typecast. That should be enough.

pglewis:
Shouldn't we be doing the PR against 2.7?

pglewis:
Ah never mind, I understand there will be a patch going out.

classes/PodsAPI.php
```diff
@@ -8619,4 +8625,41 @@ public function __call ( $name, $args ) {
...
+			if ( is_object( $v ) ) {
+				// Skip objects
+				continue;
+			} elseif ( is_array( $v ) ){
+				if ( empty( $v ) ) {
+					// Filter values with empty arrays
+					unset( $values[ $k ] );
+				}
+			} else {
+				$check = (bool) $v;
```
pglewis: `$check` is unused here.  The `!$v ` test will cast to Boolean anyway.

pglewis:
Thumbs up from me

Issue #4307, opened 2017-06-13
In Pods templates, [each taxonomy_term]{@slug}[/each] is rendering wonkily.


regnskygge:
Here is the template in question:
![template](https://user-images.githubusercontent.com/996593/27093203-72e9798e-502b-11e7-814c-217dc7613030.png)

In line 2, the focus is on` [each library_type] {@slug}[/each]`. `library_type` is a related, multiselect taxonomy and my desire is to list the slug(s) of the related taxonomy terms as classes. 

In Pods 2.6.8 this worked swimmingly...
![render-of-268](https://user-images.githubusercontent.com/996593/27093221-874b25a8-502b-11e7-88c7-c6bc30c749b7.png)

In Pods 2.6.9 it renders oddly...
![render-of-269](https://user-images.githubusercontent.com/996593/27093240-95cf91ea-502b-11e7-92c6-0324a6823177.png)

The salient `<div>` is highlighted in both images with all things equal except the odd rendering.

regnskygge:
I reverted back to 2.6.8 and everything seems to work again.

jimtrue:
@regnskygge We have a fix for you if you're willing to test it:

1) Copy your existing /pods/ plugin from /wp-content/plugins/pods/ into a /wp-content/plugins/backups/ directory (you'll probably have to make this one) (you can do this on your CPanel on your webhost or through FTP/SFTP.
2) Download the following version of Pods: https://github.com/pods-framework/pods/issues/4324
3) Install through the Plugin, Add New, Upload screen, or unzip the file above into your /wp-content/plugins/ directory. WordPress will recognize it as 'Pods'

Test! If it works, let us know back here. If it doesn't, let us know how it's still broken.

Thanks!

Issue #4308, opened 2017-06-13
Upgrade to 2.6.9 Broke IF functionality in Templates, outputting shortcodes


jimtrue:
This is from this forum post: http://pods.io/forums/topic/update-to-2-6-9-if-clauses-in-templates-broken/

I just updated to pods_2.6.9, but had to roll back after realizing it did no longer evaluate [IF] clauses in pods templates. Here’s one line that used to work under 2.6.8, but doesn’t under 2.6.9:
```
<img class="logo" src="[if video_thumbnail]{@video_thumbnail}[else]{@post_thumbnail_url.thumbnail}[/if]">
```
`video_thumbnail` is a custom field within the pod.

jimtrue:
After communicating, provided me with the inspector values of what was outputting:

Thanks for looking into this, much appreciated! I’ve got 2 examples for you with different field types. Let me know if you need anything else! I might stick with 2.6.8 for now: I’ve got IF clauses dotted all over my website, replacing them would be pretty tedious.

**field type:**
`video_thumbnail` is of type `File/Image/Video`.

**template:**
```
<img class="logo" src="[if video_thumbnail]{@video_thumbnail}[else]{@post_thumbnail_url.thumbnail}[/if]">
```

**2.6.8 output:**
`<img class="logo" src="http://path/to/thumbnail.jpg">`

**2.6.9 output:**
```
<img class="logo" src="[pod_if_field id=" 2217"="" pod="video" field="video_thumbnail" index="{_index}" seq="0" ]e0b2awrlb190ahvtym5hawx9w2vsc2vde0bwb3n0x3rodw1ibmfpbf91cmwudgh1bwjuywlsfq="=[/pod_if_field]&quot;">
```

**field type:**
`race_date_confirmed` is of type `Yes / No`.

**template:**
```
<p style="display:[if race_date_confirmed]none[else]inline[/if]"> (unconfirmed)</p>
```
**2.6.8 output:**
`<p style="display:inline"> (unconfirmed)</p>`

**2.6.9 output:**
```
<p style="display:[pod_if_field id=" 643"="" pod="race" field="race_date_confirmed" index="{_index}" seq="0" ]bm9uzvtlbhnlxwlubgluzq="=[/pod_if_field]&quot;"> (unconfirmed)</p>
```

Asking client to submit full text of template to this issue.

mstummer:
[...]

Issue #4303, opened 2017-06-12
Shortcodes within conditional (if/else) statements not parsed since pods 2.6.9


oldrup:
Hi folks

I have a website using pods and the avada theme at https://sortering.sonfor.dk/materiale/el-tandboerste/

In the pod template, I use conditional statements to display different icons depending on the pod. Sample snippet:

[if genbrugsbeholder]
  [fusion_content_box title="Genbrugsbeholder" icon="fa-trash" circlecolor="#5ebc61" /]
[else]

That worked perfectly in Pods 2.6.8, but since 2.6.9, the shortcode within the conditional statement is not begin passed, and the shortcode is displayed on the front end.

Shortcodes **outside** the conditional statement is passed just fine, showing, in this case the icon.

If I remove the conditional statements, thus always displaying the icon, it works fine - so the shortcode itself is working fine.

I downgraded to Pods 2.6.8, and the template worked perfectly again.

Any ideas?

jimtrue:
@oldrup I had a similar issue that was reported on our pods.io forums last week. They also down-revved to 2.6.8 and it fixed the issue. 

Can you do me a great favor? Could you upload the entire code in your template and not just the piece you provided above? I have a feeling this has something to do with also including post_content in the template itself. Also, it'd be really helpful to see the error you received in your templates after upgrading (ie what was output in the inspector before and after).

Thanks!

jimtrue:
@jamesgol I think this might have to do with your short-circuiting the Pods Shortcode, but I want to get you more ammunition (more details) to help troubleshoot it. We've had a long standing issue with [if][/if] blocks and [each][/each] blocks with single-select images and single-select relationships and I think this is all related.

oldrup:
Hi Jim! Thanks for digging in to this. I'm very happy about my pods solution, and sad that it breaks with the latest update. I may be "holding it wrong" ;). I'll attach the template as a txt file, screenshot of results before and after pods 2.6.9 update. Should a login to the development-site be useful to you, I'd be happy to create one.
[pods_template.txt](https://github.com/pods-framework/pods/files/1070819/pods_template.txt)
![before-pods-2 6 9](https://user-images.githubusercontent.com/17141205/27073821-76d288de-5025-11e7-8024-0eef30b2fb26.PNG)
![after-pods-2 6 9](https://user-images.githubusercontent.com/17141205/27073898-bd4bdfcc-5025-11e7-9faa-ecff9fdabeaf.PNG)

[...]

PR #4335, opened 2017-07-06
Fix for issues when Pods shortcodes are embedded in HTML

### LINKED ISSUE #4324: Templates inside HTML broken after Updating Pods
```
<div class="free-it">
    <div class="free-th">
        <a target="blank" href="[if url]
{@url}
[else]
{@permalink}
[/if]"><img src="{@image._src}"></a>
    </div>
    <div class="free-ct"><a target="blank" href="[if url]
{@url}
[else]
{@permalink}
[/if]"><span>{@post_title}</span></a></div>
</div>
```

**Becomes this on the front end:**

```
<a target="blank" href="[pod_if_field id=" 56428"=""  field="url" index="{_index}" seq="2" ]dqp7qhvybh0ncltlbhnlxq0ke0bwzxjtywxpbmt9dqo="[/pod_if_field]
```

pglewis:
Fixes #4324

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -1010,7 +1010,11 @@ function pods_do_shortcode( $content, $shortcodes ) {
 		add_filter( 'pre_do_shortcode_tag', $temp_shortcode_filter, 10, 4 );
 	}
 
-	$content = do_shortcode( $content );
+	// Build Shortcode regex pattern just for the shortcodes we want
+	$pattern = get_shortcode_regex();
+
+	// Call shortcode callbacks just for the shortcodes we want
+	$content = preg_replace_callback( "/$pattern/s", 'do_shortcode_tag', $content );
 
 	if ( isset( $temp_shortcode_filter ) ) {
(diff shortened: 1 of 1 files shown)


pglewis:
@jamesgol: Can I get a sanity check that this change won't re-break the things you fixed?

jamesgol:
I'm away on vacation, but looking at the code it won't break what was previously fixed but it looks like a band aid for the way the [if], [else], etc. tags are implemented.

You might also be able to just do do_shortcode( $content, true );

pglewis:
> You might also be able to just do do_shortcode( $content, true );

Nah, that'll force it to always escape out the shortcode as entities, giving the same result.  We really can't have it call through to `do_shortcodes_in_html_tags()` with any of our shortcodes still potentially embedded inside HTML, which will happen fairly often in templates.  `wp_kses_attr_parse` will choke and there really doesn't seem to be a kosher way to sneak around it.  This might be the least amount of duct-tape to do it.

jimtrue:
Fixes #4307, #4308 and #4324. #4303 is still outstanding for resolution.

PR #4362, opened 2017-07-24
Optimize REST API integration code and save handling

### LINKED ISSUE #4358: REST API integration - update_callback is cool, rest_insert_$type is better
Currently all REST API fields that we allow to be updated all trigger the `update_callback` via `PodsRESTHandlers::write_handler()`. This is great, since it does what we're after, except it has to call `Pods::save()` over and over, which isn't the *right* way for our purposes.

We should use the `rest_insert_{$type}` actions which are used after insert/update calls, with us then looping through additional fields that have our specialized argument added to it (like `update_callback` but maybe something namespaced with `pods_`).

## General

- [x] Update registered REST API fields with special Pods argument to know what is supposed to be writeable

## Integration hooks

- [x] `rest_insert_{post_type}`
- [x] `rest_insert_{taxonomy}`
- [x] `rest_insert_user`
- [x] `rest_insert_attachment`
- [x] `rest_insert_comment`

sc0ttkclark:
Fixes #4358

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,3 @@ before_script:
 
 script:
     - vendor/bin/phpunit
-
-branches:
-    only:
-        - master
-        - 2.x
-        - release/3.0
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -15,6 +15,7 @@ module.exports = function ( grunt ) {
 		'!.gitignore',
 		'!.gitmodules',
 		'!.travis.yml',
+		'!CODEOWNERS',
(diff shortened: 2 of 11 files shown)


sc0ttkclark:
Holding merge on this until we get confirmation from @lougreenwood

pglewis:
You might want to consider cherry picking #4360 into 2.x to fix the 5.3 failure there.

lougreenwood:
@sc0ttkclark All good - this PR fixes the issue - thanks 👍

Issue #4341, opened 2017-07-10
VersionPress 4.0 support


Ramoonus:
https://blog.versionpress.net/2017/07/versionpress-4-0-beta/

borekb:
Cool to see this ticket! Some previous discussion:

- https://github.com/versionpress/versionpress/issues/1036#issuecomment-217214337
- https://github.com/versionpress/versionpress/issues/1036#issuecomment-217214337

sc0ttkclark:
I think a good first goal is to ensure versioning the Pods configurations themselves works correctly. A main issue with versioning the configs themselves are dealing with table schema changes, field renaming (meta keys won't match, etc), field deletions (meta values will still be there, table columns/values will still be there).

Ramoonus:
it will be tough, but a huge functionality improvement

lougreenwood:
FYI - I'm already storing dumps of my entire WP dev installation in git and dumping changes into the dump in branches. So may be able to give some insight into handling Pods config changes via git...

pavelevap:
I am using Pods for my projects, so I started to play with VersionPress support.

- Pods configuration simply works, post types and postmeta are supported by default, only entities have to be checked. I tested adding several fields and I did not notice any problem.
- I also added support for `podsrel` table and related entities to start tracking data and it seems that it also works, but I will test also renaming fields, etc (PR incoming).

![image](https://user-images.githubusercontent.com/3008007/28626145-818267a2-721e-11e7-87c4-b16ee7df09f2.png)

- Shortcode support has to be added and also some advanced components will be challenging (especially custom table storage).

Some related questions:
- Relationships and postmeta are deleted and then added (not updated) by default?
- For relationship field with name `test-relation-field` there are 2 similar postmeta with ID values `_pods_test-relation-field` (serialized value) and `test-relation-field` (simple value). I am not sure about the second one (there is no prefix), is this postmeta important?
- I tried to rename some field and it seems to me that it can be renamed only once? Postmeta `old_name` is changed, but when I change it once again, there is no relation to previous name (and that is why mentioned postmeta `test-relation-field` will not work correctly)?

lougreenwood:
Nice work @pavelevap! 👍

PR #4367, opened 2017-07-26
Support for VersionPress

pavelevap:
Related: #4341 

Work in progress...

### DIFF
--- a/.versionpress/schema.yml
+++ b/.versionpress/schema.yml
@@ -1,0 +1,15 @@
+pods_relationship:
+  table: podsrel
+  id: id
+  references:
+    pod_id: post
+    field_id: post
+    item_id: post
+    related_pod_id: post
+    related_field_id: post
+    related_item_id: post
+
+postmeta:
+  value-references:
+    meta_key@meta_value:
+      _pods_*[/\d+/]: post
--- a/.versionpress/actions.yml
+++ b/.versionpress/actions.yml
@@ -1,0 +1,7 @@
+pods_relationship:
+  actions:
+    create: Created Pods relationship
+    update:
+      message: Updated Pods relationship
+      priority: 12
+    delete: Deleted Pods relationship


.versionpress/schema.yml
```diff
@@ -0,0 +1,15 @@
+pods_relationship:
+  table: podsrel
+  id: id
+  references:
+    pod_id: post
+    field_id: post
+    item_id: post
```
sc0ttkclark: The complexity of what Pods does makes this a little trickier, the item ID could be related to a post, a term, a user, etc depending on the pod the field was added on.

.versionpress/schema.yml
```diff
@@ -0,0 +1,15 @@
...
+    related_pod_id: post
+    related_field_id: post
+    related_item_id: post
```
sc0ttkclark: The complexity of what Pods does makes this a little trickier, the item ID could be related to a post, a term, a user, etc depending on the object the relationship is related to.

.versionpress/schema.yml
```diff
@@ -0,0 +1,15 @@
...
+
+postmeta:
+  value-references:
+    meta_key@meta_value:
+      _pods_*[/\d+/]: post
```
sc0ttkclark: The complexity of what Pods does makes this a little trickier, the item ID could be related to a post, a term, a user, etc depending on the object the relationship is related to.

.versionpress/schema.yml
pavelevap: Right, thank you! I did not realize that, I quickly started with extended post. We will have to use custom function, similar example is used in VP core for `_menu_item_object_id` ([here](https://github.com/versionpress/versionpress/blob/d3d0c984dff174d70ca67555a0b0b9ca325958f3/plugins/versionpress/.versionpress/schema.yml#L30)). Pods API will probably have similar function to get item type for postmeta value? I will try to look at it.

Ramoonus:
travis hangs on PHP 5.3
not really an issue imo

sc0ttkclark:
More work is needed here, just noting the link for their docs https://docs.versionpress.net/en/developer/plugin-support/

sc0ttkclark:
We can revisit this in the future if VersionPress ends up becoming active again

PR #4373, opened 2017-07-27
Simplify logic for frontier [if] shortcodes

jamesgol:
Complicated issue with nested shortcodes that fall under non-pods shortcodes (ie: inside a builder shortcode).  The frontier output isn't available to the typical WordPress do_shortcodes since
it is obfuscated in base64.

Also eliminates some code duplication, seeing that same array of tags all those times was driving me crazy.

Related to #4303

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,3 @@ before_script:
 
 script:
     - vendor/bin/phpunit
-
-branches:
-    only:
-        - master
-        - 2.x
-        - release/3.0
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -15,6 +15,7 @@ module.exports = function ( grunt ) {
 		'!.gitignore',
 		'!.gitmodules',
 		'!.travis.yml',
+		'!CODEOWNERS',
 		'!composer.json',
 		'!composer.lock',
 		'!CONTRIBUTING.md',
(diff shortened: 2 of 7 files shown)


jamesgol:
I'm really not sure if the PODS_SHORTCODE_ALLOW_SUB_SHORTCODES check is desired since it doesn't jive exactly with existing behavior but does fit the docs.

In my test I have the following:
```add_shortcode( 'jmgtest', 'jmgtest_short' );
function jmgtest_short( $a = null, $content = null ) {

	$output = do_shortcode( $content );

	return $output;
}

add_shortcode( 'jmginside', 'jmginside_short' );
function jmginside_short( $a, $content ) {
	$output = '<br>JMG shortcode text';
	return $output;
}
```
Pretty basic, the 'jmgtest' shortcode simulates what a builder shortcode would do.

So in the following test:
```
[jmgtest]
[if whatever]
[jmginside]
[/if]
[/jmgtest]
```
On current 2.x the '[jmginside]' code will not be executed but will be displayed raw.

In this test:
```
[if whatever]
[jmginside]
[/if]
```
On current 2.x the '[jmginside]' code is executed and the output is displayed.

jimtrue:
@jamesgol So the issue is with if/else/if, not if/if ?

jamesgol:
I'm away for the week on vacation at the Oregon coast.  I'll get back on this when I return.

jamesgol:
If this goes in the changes in #4335 can be safely reverted.

pglewis:
BTW: this was originally made against 2.x.  I naively tried to just change the base to 2.7 through GitHub, which obviously didn't work.  Not sure of the best way to iron that out.

jamesgol:
Closing in favor of #4423

PR #4382, opened 2017-08-01
Modal event listeners

### LINKED ISSUE #4380: 2.7 Editing relationship adds entries to other relationships on same page
in 2.7.0-b-2 if there are multiple relationships to the same post_type, making multiple edits to these relationships, seems to add entries to these relationships.

I have a pod with 3 relationships to the same post_type (Events). Each of these events or dates are setup as a single selection, list view, with allow add new.

you can edit one of the relationships (date_1) update, and when you return to the pod, everything looks fine.

Before updating the pod, edit a second relationship (date_2), save the changes, and when you return to the pod
The relationship that was edited (date_2), looks fine, only lists 1 item.
However the first relationship that was edited (date_1), now looks like there are 2 items selected, the original (date_1) and the relationship just edited (date_2).
It looks like an additional input statement, etc was inserted in the first relationship that was edited.
The one relationship that was not edited has not changed.

Edit the third relationship (date_3) and the date_3 relationship looks fine, only 1 item listed.
However date_1 now has 3 items (Date_1, Date_2, and date3).
Date_2 now has 2 items (date_2 and date_3).

clicking update to update the post saves everything as expected and everything is back to only 1 item displayed for each date field.

image

pglewis:
Fixes #4380 
~- [ ] Find out why selection limits on add new are not being respected and fix it~ (pre-existing)
- [x] Event listeners should only respond once
- [x] Trap cancel events and disable all modal event listeners


pglewis:
Still not quite ready, the latest commit introduces a glitch where the item limit (1 in my case) is not detected on add new for some reason.

pglewis:
Looks like the selection limit issue is pre-existing, that happens on 2.7 for me now so it needs its own issue.  This one is ready to test.

Issue #4388, opened 2017-08-03
Relationship field "List view" format not adding posts when selected


rob:
I'm using the new "List view" format for the Relationship field in the Pods 2.x branch. I set it to be related to "Posts (post)."

**Issue:** When searching for posts and clicking enter, or clicking a post manually, it doesn't add that post to the list. It doesn't do anything. However, sometimes it _does_ add the post to the list. I would say usually it doesn't, however.

Don't see any JavaScript errors in the console or anything.

Video showing issue, where it both adds and doesn't add posts:

![record](https://user-images.githubusercontent.com/503/28945693-88d6c75c-7875-11e7-896d-2a7c567c5799.gif)

Any thoughts?

sc0ttkclark:
Thanks for posting your issue and describing it very well, we really appreciate the effort you put to provide the video. We'll take a look and see what's up.

pglewis:
I wasn't aware that list view had even been put into the 2.x branch.  Can you test this on the 2.7 branch?  That's the only one where the list view work is actively going on.

pglewis:
Actually, I'm guessing "2.x" was a typo... I don't get anything for list view on the 2.x branch.  

I've tested this under 2.7 and haven't been able to reproduce it yet.  

I'm also seeing entity escapes ("Ireland's") and raw markup (<em>...</em>) in your screen cap.  This does not happen for me in the 2.7 branch.  I'm starting to wonder if you have the older version of select2's javascript cached.  We should be cache-busting via version number and Wordpress's enqueueing, I think, but it wouldn't hurt to make sure it's not keeping an old local copy of select2 for some reason.

rob:
Hmm, that might be it. I upgraded via GitHub Updater from the stable 2.6
version installed. I'll check tomorrow and let you know. Thanks!

Rob

On Aug 3, 2017 22:37, "PG Lewis" <notifications@github.com> wrote:

Actually, I'm guessing "2.x" was a typo... I don't get anything for list
view on the 2.x branch.

I've tested this under 2.7 and haven't been able to reproduce it yet.

I'm also seeing entity escapes ("Ireland's") and raw markup (*...*) in your
screen cap. This does not happen for me in the 2.7 branch. I'm starting to
wonder if you have the older version of select2's javascript cached. We
should be cache-busting via version number and Wordpress's enqueueing, I
think, but it wouldn't hurt to make sure it's not keeping an old local copy
of select2 for some reason.

—
[...]

PR #4405, opened 2017-08-08
AJAX mode pick fields 

pglewis:
Fixes: #4388 

- [x] Return full data for the client side models when getting ajax results
- [x] Update the partial collection on the client side after an ajax call
- [x] Test regular autocomplete in ajax mode
- [x] Find out why HTML entities are escaped and not interpreted when the data arrives via ajax


pglewis:
This one is set for review and testing, all the basics seem to be working now.

rob:
I can confirm that this is working perfectly! Both adding new posts and showing entities. Thanks for the quick fix @pglewis!

sc0ttkclark:
👍

classes/fields/pick.php
```diff
@@ -2079,7 +2079,7 @@ public function get_object_data( $object_params = null ) {
 
 				if ( $autocomplete ) {
 					if ( 0 === $limit ) {
-						$limit = 30;
+						$limit = 3;
```
sc0ttkclark: 3?

sc0ttkclark: Nevermind, this ultimately was not changed in the final diff

Issue #4403, opened 2017-08-08
IF Template Tag on Single-Select Relationships (and Images) strips Related_Field


jimtrue:
I ran into this today on a Single-Select Relationship field:

```
[if related_field]
   <a href="{@related_field.permalink}">{@related_field.post_title}</a>
[/if]
```
This returns the permalink and post_title of the CURRENT post, completely ignoring the related_field traversal. If I remove the `if` wrapper, it works properly.

I went back on several versions to 2.6.7, 2.6.8, 2.6.9, 2.6.10, 2.x and the new release/2.7 and it's happening in all of them.  I can compare this to #3481 and #4354

jimtrue:
This has been corrected as of #4408

PR #4408, opened 2017-08-08
Don't strip field traversal if the specified field does not contain a period

jamesgol:
Needs further tests built to make sure this doesn't break anything else.

Related to #4403

Currently the code appears to expect:
```
[if rel_field.permalink]
{@permalink}
[/if]
```
When it appears that the desired behavior is:
```
[if rel_field]
{@rel_field.permalink}
[/if]
```
This should allow both behaviors to work, in the first case the context is "switched" inside the if to the object related via rel_field.  In the second case the context remains with the original object but allows it to be referenced via dot traversal.

Need to build tests for the other frontier related shortcodes to ensure more isn't broken by this


jimtrue:
I tested this against single-select and it worked and didn't break any of my multi-select. I'll dig into the other 4 cases to make sure this also hits the image issues.

PR #4415, opened 2017-08-11
Fixes for showing/hiding some of the List View specific options

### LINKED ISSUE #4038: 2.7/feature Editing List View Field doesn't show container for 'Show Icons, Show Edit, Show View' until you change the field type.
2.7/feature Editing List View Field doesn't show container for 'Show Icons, Show Edit, Show View' until you change the field type. submitted via Slack by jimtrue

PR: #4415

pglewis:
Fixes #4038 

This is a two-parter: 

- [x] Don't check multi-format dependencies if single select is the type and vice versa
- [x] Come up with a way to run dependency checks again when the selection type changes

### DIFF
--- a/classes/PodsForm.php
+++ b/classes/PodsForm.php
@@ -825,65 +825,65 @@ class PodsForm {
      * @param array $options array( 'depends-on' => ..., 'excludes-on' => ...)
      * @param string $prefix
      *
-     * @return string
+     * @return array
      * @static
      * @since 2.0
      */
     public static function dependencies( $options, $prefix = '' ) {
-
         $options = (array) $options;
         $classes = $data = array();
-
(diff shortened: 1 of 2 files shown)


pglewis:
Note that this workaround will be a stop-gap for 2.7.  Dynamic dependency checking will be reworked for 2.8.

classes/PodsForm.php
```diff
@@ -825,68 +825,68 @@ public static function field_setup( $field = null, $core_defaults = null, $type
      * @param array $options array( 'depends-on' => ..., 'excludes-on' => ...)
      * @param string $prefix
      *
-     * @return string
+     * @return array
```
pglewis: The changes in this commit were in preparation for some more changes to be made but that turned out to be unnecessary.

Issue #4097, opened 2017-03-31
display() shows value|Label results inconsistently for custom defined list


rpall:
This is reproducible on a fresh Wordpress install. 

WP version: 4.7.3
Pods version: 2.6.8

No other plugins installed.
Tested on PHP 7.0.15 and 5.6.30

Post type: Advanced Content Type

Create a new ACT type pod. Add one field with type Relationship: Simple custom defined list with multiple select enabled with checkboxes. Add these value|Label options: 
yellow|Kollane
green|Roheline
brown|Pruun
red|Punane

Add a new item to the pods and mark green, brown, red as selected. 

Using this code in a custom wordpress page template when displaying the pods on the frontend: 

```
$params = array(
   'limit' => -1
);

$listings = pods( 'listing', $params );

while ( $listings->fetch() ) {
   echo $listings->display( 'color' );
}
```

The result displayed on the page is this: "green, brown and Punane".

When you also mark yellow|Kollane as selected then the result displayed on the page is correct: Kollane, Roheline, Pruun, and Punane

jamesgol:
Could not reproduce this issue on Pods 2.6.8, 2.x, or 2.7b.  Tried with CPT and ACT

jamesgol:
Weird, was just able to reproduce it but not sure what I changed.  Will dig in some more

jamesgol:
Issue fixed in release/2.7.  Unit test added to prevent future regression.  Please test and re-open if you are still experiencing issues.

PR #4420, opened 2017-08-15
Allow table based storage to properly work with custom simple relationships

jamesgol:
If the first element of multiple custom simple element isn't checked then the display values are only used for the last element.

Related to #4097


sc0ttkclark:
You are an angel!

PR #4423, opened 2017-08-16
Simplify logic for frontier [if] shortcodes (again)

jamesgol:
Complicated issue with nested shortcodes that fall under non-pods shortcodes (ie: inside a builder shortcode).  The frontier output isn't available to the typical WordPress do_shortcodes since
it is obfuscated in base64.

Also eliminates some code duplication, seeing that same array of tags all those times was driving me crazy.

Related to #4303

Reimplementation of #4373 based of release/2.7

### DIFF
--- a/components/Templates/includes/functions-view_template.php
+++ b/components/Templates/includes/functions-view_template.php
@@ -16,6 +16,19 @@ add_shortcode( "pod_before_template", "frontier_template_blocks" );
 add_shortcode( "pod_if_field", "frontier_if_block" );
 
 /**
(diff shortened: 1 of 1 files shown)


jamesgol:
Passes all tests in #4402 even the ones I marked incomplete

pglewis:
100% Rock Star.  

I've set Jim as a reviewer because he has a personal pile of magic tag/template tag tests. We should probably arrange a pow-wow when you both have time so we can get formal tests for any other cases he has beyond what you've added.

Issue #4437, opened 2017-08-30
Relationship field with custom post statuses doesn't save


rosalynpoort:
I have an advanced content type pod with a relationship field that relates to a custom post type. i have custom post statuses for the custom post type. on the relationship field in the advanced content type pod i can choose to display the posts with my custom statuses but the field won't save. it only saves when i choose a published post

jimtrue:
@pglewis Need to verify this is an issue on 2.7 pick fields. I'm testing that today. If it's corrected in 2.7 Yay! But if we give the option to allow relationships on specific post statuses, we should be saving the connection.

jimtrue:
@rosalynpoort What are the custom post statuses you have that you're trying to assign relationships to? I tested with 'draft' and while the Draft items showed up in the pick list, they wouldn't save to the relationship field.

rosalynpoort:
@jimtrue yes i tested with 'draft' as well as 'pending-review' and 'maybe'. same results, they all show up in the pick list but won't save to the field

jimtrue:
I just tested in the beta 2.7 as well and while it allows me to add a new draft relationship post and it adds it to the list. When I save, it's no longer in the related record post. This is most definitely a bug.

rosalynpoort:
@jimtrue thanks for looking into it. I've worked around it for now by just making it a text field for the ID as I'm adding using the pods api anyway and not the WP dashboard.

jimtrue:
For clarification: this is the 'Post Status' under Additional Options. This provides what posts show up in the Relationship pick list. Changing the status allows them to show, but they aren't saving to the field on save.

<img width="706" alt="screenshot 2017-08-30 12 50 46" src="https://user-images.githubusercontent.com/3052875/29884648-18a0bdd6-8d82-11e7-86d6-0d329dafeede.png">

jamesgol:
Looking at the database it looks like the data is actually being saved, but just never loaded to be displayed.  I'm closing in on the problem.

th-santos:
I am facing the same issue.

I have an Extended Post Type ("post") with a "Relationship" Field Type (non-bidirectional) related to a Custom Post Type ("company"). 

I am using `pods_api_post_save_pod_item_post` to get field values from my "company" post type. If "company" status is "publish", I can get those values, but if it is "private", I cannot get any field value.

Issue #4409, opened 2017-08-09
Unable to save a relationship field if relating to draft post


lougreenwood:
I have a relationship field which I have enabled 'draft' posts for:

<img width="564" alt="screen shot 2017-08-09 at 18 51 08" src="https://user-images.githubusercontent.com/12051736/29135978-c070de88-7d33-11e7-94be-d6e536b3200b.png">

I have a post of this Pod type, when I edit the post, I can use the relationship field to choose a draft post, but when I save the post, the relationship isn't saved and the relationship field is empty.

sc0ttkclark:
Can you clarify -- what post type is it related to?

lougreenwood:
Standard Pods table-based CPT

lougreenwood:
Update: I just went through and updated many of the draft posts to 'publish' and they appeared in the dropdown selector - no need to re-link after just changing the post's status.

sc0ttkclark:
Oh that helps a ton, thanks!

sc0ttkclark:
Fixed

Issue #3556, opened 2016-05-31
Problem with related field with posts in draft status


MMiX2K:
Hello! We have a problem with pods 2.x. I'm showing publish, draft and pending posts in a relationship field (select and checkboxes) adding the post_status to the where condition (a custom post type in admin). The value is saved in database but the checkbox is only showing marked / select only showing selected, when the related post is published.

I saw that the problem is "where_default", always filtering published posts. I tried "pods_api_get_table_info_default_post_status" filter without success.

If I comment the line "$this->where_default = $this->pod_data[ 'where_default' ];" in __construct of "PodsData.php" the related fields show selected values flawlessly (draft and pending post, of course, published post always have worked).
I never have modified a plugin code, but at least works in this way. Is there another way (any filter for example or modify "where_default" value) to make this? It's a bug? I would like to know.

Thanks!

jamesgol:
Just for a note here because I vaguely remember working on this way back when.  The function that calls the 'pods_api_get_table_info_default_post_status' filter caches the results so the filter is only called the first time and then cached.  You could try clearing the cache and see how it works, but the code definitely needs to get refactored so everything is handled properly.

MMiX2K:
I deleted the cache and now it's working perfect :)
Thanks @jamesgol!

jamesgol:
I'm not sure I would close this, I'd say it's a legit issue that needs to be fixed when the code gets refactored.  This is something that should be able to be changed by the filter every time based on the variables passed in.

JoryHogeveen:
Related: https://github.com/pods-framework/pods/pull/3626

quasel:
no idea - is this still an issue? @jamesgol

Issue #3451, opened 2016-03-17
Private pods not displaying front end to author


enquirer32:
I’ve created two pods. The entries associated with these are saved as ‘private’. One cant view the content on the front end of the website.

As admin on this WP installation. 
I created new a CPT. 
Created one field. 
I added new 'testpod' from admin - 'Add New'. 
Filled in my one field. 
I created a new post and inserted the following shortcode: `[pods name="testpod"]<br />{@test_field}<br />[/pods]`
I made sure the pod was 'private'
If one views the post on front end one can see the title but no output (it is 'Private')
Reset the pod item to 'public' - the content shows up on the front end.

For me this is a serious issue and it isn't how core WP works. When something is 'Private' the author can always see their own posts. 

My set-up: My intention was to restrict the content others could see but always intending that the creator of the content (adding content through a Pod form) could see their own. For example, I have a logged in user who creates content in both pods. He has all the permissions required to edit ‘private’ content and it his content. So long as the data in the pods he created is private he can't access this data.

In my real install I also have a couple of single relationship dropdowns. However, he can’t access these unless the pods are ‘published’ – they don’t work as private either. This is also an unforseen problem which means I can't use Pods the way I had intended.

Is there a temporary fix for this - some hack I can add to the files? Was this always your intention?

You don't need files or images to replicate this yourself - it is very straight forward.

enquirer32:
OK, so I've revisisted this. After a follow-up from Jim I installed a clean WP and installed only your plugin and worked up my example above creating public and private posts with public and private pods.... I'm attaching the json file for this. When the pod is marked as private it does not show in the front end as it should - following WP rules. I set up two pods one public and one private. The private pod did not show (I verified this by making it public when it did).

[...]

PR #4442, opened 2017-09-02
Allow where_default to pass through SQL queries

jamesgol:
What's happening is that the 'where_default' is always getting used because the query is a post_type but does not include the post_status in the where, so PodsData->query_fields() was using the
'where_default' param which wasn't getting passed through properly.

Related to #4437


jamesgol:
Related to #4409 #3556 #3451 #1654 #1577 and possibly others

PR #4449, opened 2017-09-06
Fix php notice from missing array key

sc0ttkclark:


jamesgol:
D'oh, sorry was my fault

PR #4452, opened 2017-09-06
Allow magic tags engine to be replace via WordPress filter

### LINKED ISSUE #4447: Allow magic tags engine to be replaced via WordPress filter
PR incoming

jamesgol:
Creates new WordPress filter 'pods_pre_do_magic_tags'.  By default returns false which continues on with normal execution, if something other than false is returned that value is returned.

Resolves #4447 

Thought I opened a PR on this but I guess I just created the branch


sc0ttkclark:
Updated to default to null instead of false, updated phpdoc info

sc0ttkclark:
Pending Travis confirmation tests pass, then merge

PR #4487, opened 2017-09-29
Change `word-break` to `break-all` to deal with long lines in codemirror boxes

### LINKED ISSUE #4379: Field boxes extending to obscure side meta-boxes
When editing a post and pasting a Google embed code into an html field, the box extends out to obscure all meta boxes on the right side of the page. If there are other fields on the page, all field boxes will extend out in the same way. It seems to be where a line exceeds a certain length it fails to wrap around to the next line.

https://screencast-o-matic.com/watch/cbivYHlaPt

I get the same issue with both the latest stable version and 2.7 beta.

pglewis:
Fixes #4379


JoryHogeveen:
Checked and looks good to me!

pglewis:
Yeah, mostly had you and Nic in there for a sanity check.  I kinda want to check the Git history since that's in an over-ride file... slight nagging that was set that way for a reason in the past.

pglewis:
> I kinda want to check the Git history since that's in an over-ride file... slight nagging that was set that way for a reason in the past.

Turns out those aren't technically "overrides" as we're not loading any default styles for codemirror; all the codemirror styling is in that partial and doesn't override anything else.  That `word-break` setting was just what we started with and has been there since the dawn of time.  Or 5 years ago, which is nearly the same thing.

PR #4492, opened 2017-10-05
Roles AJAX data

### LINKED ISSUE #4477: PODS 2.7RC1 Roles component goes in loop when save
schermata 2017-09-22 alle 09 46 33

pglewis:
Fixes #4477

### DIFF
--- a/ui/js/jquery.pods.js
+++ b/ui/js/jquery.pods.js
@@ -265,7 +265,7 @@
                             if ( -1 === d.indexOf( '<e>' ) && -1 === d.indexOf( '</e>' ) && -1 !== d ) {
 
                                 // Added for modal add/edit support.  If we get a valid JSON object, we assume we're modal
-                                if ( 'object' === typeof data ) {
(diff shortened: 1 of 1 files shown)


pglewis:
> Hmm how is data strictly an object and strictly null?

It's if `data` is an object but NOT null: `if ( 'object' === typeof data && null !== data )` 

`data` is the result from parseJSON, `d` is the raw data.  In the case of a roles save the AJAX call returns either `true` or an empty string in `d` (can't recall which)... but in any case the result of running that through `parseJSON` will return `null`.  

That new section of code assumes a valid object from parseJSON signals DFV support for modal add/edit but due to legacy Javascript stuff `typeof null` is 'object' as well.

sc0ttkclark:
Oh FFS JavaScript, leave it to them to make null an object

pglewis:
It was a glaring mistake in early implementations... but the real prob is the usual suspect: back-compat.  Once it's in the wild...

pglewis:
Also, still not as glaring as `typeof NaN` = number :)

PR #4501, opened 2017-10-07
Add test for saving template

### LINKED ISSUE #4500: Saving Pods templates is broken
Appears to be caused by https://github.com/pods-framework/pods/commit/9698a88364073e4f9e43348f0d839c2afc0aad4a

( ! ) Notice: Trying to get property of non-object in /var/www/html/wp-content/plugins/pods/classes/PodsMeta.php on line 1209
--

1 | 0.0033 | 436200 | {main}( ) | .../post.php:0
2 | 0.4352 | 28830320 | edit_post( ??? ) | .../post.php:193
3 | 0.4481 | 28834392 | wp_update_post( ???, ??? ) | .../post.php:378
4 | 0.4488 | 28844880 | wp_insert_post( ???, ??? ) | .../post.php:3583
5 | 0.4976 | 29928560 | do_action( ???, ???, ???, ??? ) | .../post.php:3510
6 | 0.4976 | 29928936 | WP_Hook->do_action( ??? ) | .../plugin.php:453
7 | 0.4976 | 29928936 | WP_Hook->apply_filters( ???, ??? ) | .../class-wp-hook.php:323
8 | 0.5033 | 29908624 | PodsMeta->save_post( ???, ???, ??? ) | .../class-wp-hook.php:298

Simple fix is to just return if (!$pod) but need to dig a bit more.

jamesgol:
Creating test under directory, presumably more template tests will go there

This should fail initial tests, after tests I'll add a commit to fix the issue.

Fixes #4500

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -1210,6 +1210,7 @@ class PodsMeta {
 				return;
 			}
 		}
+		// The following code will run for all post_types (not just Pods)
 
 		if ( false !== $nonced && ! empty( $groups ) ) {
 			foreach ( $groups as $group ) {
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -1202,16 +1202,13 @@ class PodsMeta {
 		$pod  = self::$current_pod;
 		$data = array();
 
-		if ( !$pod ) {
-			// Do not handle if there is no Pod associated with this post_type
-			return;
-		}
+		if ( $pod ) {
+			$rest_enable = (boolean) pods_v( 'rest_enable', $pod->pod_data['options'], false );
 
-		$rest_enable = (boolean) pods_v( 'rest_enable', $pod->pod_data['options'], false );
-
-		// Block REST API saves, we handle those separately in PodsRESTHandlers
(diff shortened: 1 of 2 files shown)


classes/PodsMeta.php
```diff
@@ -1202,6 +1202,11 @@ public function save_post( $post_id, $post, $update = null ) {
 		$pod  = self::$current_pod;
 		$data = array();
 
+		if ( !$pod ) {
```
sc0ttkclark: I appreciate this condition @jamesgol but we wanted to run some of the code below for non-Pods too.

What we want is to probably put the `$rest_enable` code and the subsequent return into a `if ( $pod ) {` block. I know this whole class is getting messy, we're going to have to clean things up at some point :/

jamesgol: That is how the other functions added in the same commit do it (wrapping the rest code)

pglewis:
If The Scott approves, I'll just blindly approve.

jamesgol:
I'm still not sure why you (or anyone else) didn't see the error

pglewis:
Yeah, that was still nagging me as well, I don't like solutions with unanswered questions.  At the same time this was an unscheduled side-trip and your time could have been spent looking into what you started on, so I'm good with "fixed is fixed" for once.

Issue #4484, opened 2017-09-29
Relationship to Users is not Traversing inside an [each related_user][/each] block


PodsBot:
Relationship to Users is not Traversing inside an [each related_user][/each] block submitted via Slack by jimtrue

jimtrue:
Believe is similar to or related to the same issue with #3376 

If you have a relationship field to Users (ie `rel_users`) and inside a Pods Template you do:

`{@rel_users}`
You'll get a list display of the Users related

If however, you do something proper that will force traversal like the following:

`
[if rel_users]
    <ul>
    [each rel_users]
    <li>{@display_name}, {@user_email}</li>
    [/each]
    </ul>
[else]
     No Users
[/if]
`
You will get a list item for each user in the relationship, but if you look at the {@ID} it's also returning the 'parent' ID. That means there's something wrong with the code when there's a relationship to a User Object, in much the same way as in #3376 when it returns the values from the parent object, not the items _in_ the relationship unless you use _src or _img. Something in our handling of these objects is providing issues in Pods Template traversals.

JoryHogeveen:
2.7.1? :(

pglewis:
> 2.7.1? :(

Keep in mind we're nearly at 2.7.1 now.

JoryHogeveen:
Are we? 2.7 isn't out yet haha

pglewis:
We may un-punt it anyway, if James's work in #4503 addresses this.

PR #4503, opened 2017-10-09
Allow handling of arbitrary data that aren't Pods

jamesgol:
Relationships to things like Pods don't really have data to traverse into with each.  So here it creates a pseudo magic tag handler that uses data from the array instead of from a Pod.

This is really closer to how all tags should be handled anyways and lines up closer to how Twig handles things.

Related to #4484

### DIFF
--- a/components/Templates/includes/functions-view_template.php
+++ b/components/Templates/includes/functions-view_template.php
@@ -339,6 +339,7 @@ function frontier_do_subtemplate( $atts, $content ) {
  * @param array $data
  *
  * @return string
+ * @since 2.7
  */
 function frontier_pseudo_magic_tags( $template, $data, $pod = null ) {
 
--- a/components/Templates/includes/functions-view_template.php
+++ b/components/Templates/includes/functions-view_template.php
@@ -256,13 +256,14 @@ function frontier_do_subtemplate( $atts, $content ) {
 
 	$out = null;
 	$pod = pods( $atts[ 'pod' ], $atts[ 'id' ] );
+	$field_name = $atts[ 'field' ];
 
-	$params = array(
-		'name' => $atts[ 'field' ],
-	);
-
-	$entries = $pod->field( $atts[ 'field' ] );
+	$entries = $pod->field( $field_name );
 	if ( ! empty( $entries ) ) {
+
+		$field = $pod->fields[ $field_name ];
+		// Object types that could be Pods
+		$object_types = array( 'post_type', 'pod' );
 
 		/**
 		 * Note on the change below for issue #3018:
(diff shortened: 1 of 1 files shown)


components/Templates/includes/functions-view_template.php
```diff
@@ -297,30 +299,106 @@ function frontier_do_subtemplate( $atts, $content ) {
...
 /**
+ *
+ * Search and replace like Pods magic tags but with an array of data instead of a Pod
+ *
+ * @param Pod $pod
+ * @param string $template
+ * @param array $data
+ *
+ * @return string
+ */
```
quasel: add @since 2.7. ?

jamesgol: Good catch.  If this is going to get added we can add whatever version it is beforehand.

jamesgol:
I don't really like adding more hacks on top of hacks, but we're stuck with this setup for now.  There's another way to go at this but it would be far more hacky.  Converting anything within an each to stick with the same pod but add dot notation for traversals.

jamesgol:
I don't know or have an opinion on the best way to handle elements that will be single elements in array.  For example if you end up with a relationship field to the predefined list Countries.  How should it be referenced in the template?
```
[each rel_to_countries]
{@???}
[/each]
```
To the code (frontier_do_subtemplate() ) it will get passed an array with a list of the keys  for example:
```
Array
(
    [0] => DZ
    [1] => AS
    [2] => AD
)
```

components/Templates/includes/functions-view_template.php
```diff
@@ -256,14 +256,15 @@ function frontier_do_subtemplate( $atts, $content ) {
...
-
-	$entries = $pod->field( $atts[ 'field' ] );
+	$entries = $pod->field( $field_name );
 	if ( ! empty( $entries ) ) {
 
+		$field = $pod->fields[ $field_name ];
+		// Object types that could be Pods
+		$object_types = array( 'post_type', 'pod' );
+
 		/**
```
pglewis: I have no memory of this place...

components/Templates/includes/functions-view_template.php
```diff
@@ -297,30 +299,106 @@ function frontier_do_subtemplate( $atts, $content ) {
...
+ * Search and replace like Pods magic tags but with an array of data instead of a Pod
+ *
+ * @param Pod $pod
+ * @param string $template
+ * @param array $data
+ *
+ * @return string
+ */
+function frontier_pseudo_magic_tags( $template, $data, $pod = null ) {
+	return preg_replace_callback( '/({@(.*?)})/m',
```
pglewis: Maybe `/({@(.*?)}.+)/m` so that "{@}" isn't matched?

jamesgol: I took the pods magic tag code and dropped it in, trying to keep it as identical as possible.

pglewis: Fair enough; jostle things as little as possible until we rework it.

components/Templates/includes/functions-view_template.php
```diff
@@ -297,30 +297,117 @@ function frontier_do_subtemplate( $atts, $content ) {
...
+				$out .= pods_do_shortcode( frontier_pseudo_magic_tags( $template, $entry, $pod ), frontier_get_shortcodes() );
+			}
 		}
 	}
 
 	return pods_do_shortcode( $out, frontier_get_shortcodes() );
 }
 
 /**
+ *
```
sc0ttkclark: Why don't we add something like this into a more reusable function outside of the frontier namespace. We might even find a way to abstract some of this logic so both normal magic tags processing and this one aren't too different?

pglewis: If we can do it reasonably safely in time for 2.7.  If not, I'd vote for saving that for a big rework of templates.

jamesgol: I thought about this but then decided against it because I didn't want to encourage anyone to go down the dead end of the existing magic tag system.

sc0ttkclark: Fair enough, magic tags aren't super amazing anyway

components/Templates/includes/functions-view_template.php
```diff
@@ -258,7 +258,7 @@ function frontier_do_subtemplate( $atts, $content ) {
 	$pod = pods( $atts[ 'pod' ], $atts[ 'id' ] );
 	$field_name = $atts[ 'field' ];
 
-	$entries = $pod->field( $field_name );
+	$entries = (array) $pod->field( $field_name );
```
sc0ttkclark: One of my fav things about this PR is  this commit

sc0ttkclark: But we ought to cast $entries as an array after we check if it's empty, right?

pglewis:
Any other issues besides #4484 that this may fix, or could fix with a little more work?

components/Templates/includes/functions-view_template.php
pglewis: I'm not sure if it matters but it doesn't hurt to change it.

pglewis:
> I'm not sure of an easy way to reference the display name associated with the list and I'm certain more field specific code should not be added to various places like this.

The serious rub is the primary use case I see is getting to the display name... we're in a display context and that's its job after all.

jamesgol:
> The serious rub is the primary use case I see is getting to the display name... we're in a display context and that's its job after all.

That's true and when everything is refactored to be more generic (PodsObject?) it will be trivial to do, but I just can't see adding more of these edge case handlers all over the place like we have now.

AFAIK so far nobody has asked for that specific functionality, only explicitly the user info.  I just did further testing on other objects for completeness.  Honestly I wouldn't even document or encourage people to access fields like that.

PR #4506, opened 2017-10-11
Use the correct field index, `post_title` is only gonna work for posts

### LINKED ISSUE #4505: 2.7 RC1: Search (in a Settings pod) doesn't work with a list greater than 30
Possibly connected to #1259 and #4504.

I'm testing a few custom settings, both multiple-select. When using the *Autocomplete* or *List view* formats, the autocomplete doesn't show any possibilities when the total number of possibilities is greater than 30.

## Example
I have a setting to select multiple users from the user base on the site. Here is what it looks like when I have 30 or less registered users:

30-users

When I add the 31st user, I get this:

more-than-30

... And no matter what I type in there, I get no results.

If I happen to know the exact result I'm looking for, type it in, and hit Enter, I'll get the addition, but with a blank entry:

If I refresh the page, the name will be there:

## Other Notes
- I've tested this both with Users and ~~Posts~~. Same result for both. **(UPDATE: Posts are fine. Users seem to be the only problem.)**
- 30 seems to be the magic number. Results are ok below thirty, and I've gone as high as almost 9,000 users

pglewis:
Fixes #4505

### DIFF
--- a/classes/fields/pick.php
+++ b/classes/fields/pick.php
@@ -2297,7 +2297,7 @@ class PodsField_Pick extends PodsField {
 						}
 
 						if ( 'admin_ajax_relationship' === $context ) {
-							$items[] = $this->build_dfv_field_item_data_recurse_item( $result['ID'], $result[ $search_data->field_index ], (object) $object_params );
+							$items[] = $this->build_dfv_field_item_data_recurse_item( $result[ $search_data->field_id ], $result[ $search_data->field_index ], (object) $object_params );
 						} else {
 							$data[ $result[ $search_data->field_id ] ] = $result[ $search_data->field_index ];
 						}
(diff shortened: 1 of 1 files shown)


classes/fields/pick.php
```diff
@@ -2297,7 +2297,7 @@ public function get_object_data( $object_params = null ) {
 						}
 
 						if ( 'admin_ajax_relationship' === $context ) {
-							$items[] = $this->build_dfv_field_item_data_recurse_item( $result['ID'], $result['post_title'], (object) $object_params );
+							$items[] = $this->build_dfv_field_item_data_recurse_item( $result['ID'], $result[ $search_data->field_index ], (object) $object_params );
```
sc0ttkclark: `$result['ID']` should be changed to `$result[ $search_data->field_id ]` too

pglewis: Thank you for catching this

sc0ttkclark:
👍

PR #4512, opened 2017-10-12
Fix i18n component metabox error on latest 2.7 branch version

JoryHogeveen:
See title.
I just re-used the same branch.
The parameter that was passed isn't the Pod but the WP_Post object. No idea why this was changed but no need to use it after all since I already store the current Pod in a parameter.


pglewis:
This is good to go @JoryHogeveen ?

pglewis:
Oh, and is there a related issue for it?

JoryHogeveen:
Nah, I just used the same branch again.

And yes, it's good to go!

Issue #4518, opened 2017-10-16
DFV CSS settings pages


JoryHogeveen:
The UI for setting pages is not 100% for DFV fields due to styles added by `_admin.scss`.
Some of these styles are needed though. Any ideas on CSS structure are welcome.

![image](https://user-images.githubusercontent.com/826148/31624332-a7c8f1aa-b2a2-11e7-8705-f9055d67c8cc.png)

pglewis:
Working backwards from what I'd like to ultimately see, these are the built style files I think we need: 

- "Main" pods.css file, limited to admin pages, contains most of what we need
- wizard styles, extracted for some size savings in the main file because they're only needed in specific places
- styles needed for the front end only (eventually probably named 'pods-front.css').  We want to keep the stylesheet overhead on front-end pages to a minimum.  

The intention thus far has been for 'pods-form.css' to transition into 'pods-front.css' but I don't think we're quite there yet.  As the name suggests, it was originally styles for where PodsForm is used and those are more scattered than the breakdown we're after.  

I'm open to discussion on that breakdown, that's just what I've had in mind thus far.

nicdford:
I think @pglewis is on the right track. I think the biggest issue we had before was just some overlapping of styles and it could probably get into ship-shape with a good review/rewrite. I'm happy to take on _admin.scss since that's the main culprit at the moment.

pglewis:
Proposed patch in #4526

PR #4526, opened 2017-10-20
Fix DFV field view on settings pages

nicdford:
Fix for #4518


nicdford:
Hold off on merging this, i didn't fully test how this would impact other fields, so i'm going through that now.

ui/styles/src/base/_admin.scss
```diff
@@ -486,9 +486,9 @@ div.pods-manage-field .pods-field-option-group .pods-field-option-group-label {
 /**
  * Pods DFV
  */
-.pods-manage-field .pods-dfv-field .pods-pick-values > ul {
-    border: 0;
-}
+// .pods-manage-field .pods-dfv-field .pods-pick-values > ul {
```
pglewis: We have a history in git, best to just remove lines unless there's another compelling reason here

nicdford: Affirmative. Shaking off the rust :)

nicdford:
OK, this should be a much better fix IMO. Let me know if something doesn't look right and I can fix it.

JoryHogeveen:
@nicdford 
Can you post a screenshot of what it should look like with this fix? I think I'm missing something.

ui/styles/src/base/_admin.scss
```diff
@@ -486,8 +479,9 @@ div.pods-manage-field .pods-field-option-group .pods-field-option-group-label {
 /**
  * Pods DFV
  */
-.pods-manage-field .pods-dfv-field .pods-pick-values > ul {
-    border: 0;
+.pods-manage-field .pods-dfv-field .pods-ui-list-autocomplete {
```
JoryHogeveen: The images view should be similar right?
`.pods-manage-field .pods-dfv-field .pods-ui-file-list.pods-field-template-rows`.

pglewis:
Is this one ready to go now in your opinion, @JoryHogeveen ?

JoryHogeveen:
Yep, works for me. Might have some improvements but that will require a larger refactor so let's punt that to a later version.

PR #4535, opened 2017-10-31
Inline `wp_print_scripts` causes load order problems with script dependencies

### LINKED ISSUE #4514: JS errors when using pods form with 2.7
Hi Pods team
I've encountered an error when using pods form with 2.7. It causes problems to wysiwyg fields. Could you have a look please?

https://pods.io/docs/code/pods/form/ 
wysiwyg

Thanks

pglewis:
Fixes #4514

The dependency tree for the `pods` script has grown via its dependency on `pods-dfv` resulting in a script load-order bug.

### DIFF
--- a/ui/front/form.php
+++ b/ui/front/form.php
@@ -1,5 +1,6 @@
 <?php
 wp_enqueue_style( 'pods-form' );
+wp_enqueue_script( 'pods' );
 
 // unset fields
 foreach ( $fields as $k => $field ) {
--- a/ui/front/form.php
+++ b/ui/front/form.php
@@ -1,10 +1,6 @@
 <?php
 wp_enqueue_style( 'pods-form' );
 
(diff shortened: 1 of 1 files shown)


jamesgol:
For future reference (in case someone has to dig back in another 5 years to see this (Hi Mom!)). This all came from https://github.com/pods-framework/pods/commit/d3d9bb2a582062a2e6464e885252c5f8720dcda3
which appears to have added a force script printing due to reordering in PodsUI which appears to be a long deprecated method.

With the 2.7 changes, 'pods-dfv' is a dependency on 'pods'.  If the pods scripts are force loaded early (like in this specific case) before the WordPress media scripts the wp.media.view.settings does not get populated properly.

pglewis:
Big props to @jamesgol for sharing the pain and ultimately pinpointing the culprit.

PR #4549, opened 2017-11-03
Downgrade nodejs to 8.9.0

jamesgol:
This PR is for the 2.x branch


jamesgol:
Apparently this is unneeded because 2.x doesn't use node-sass

PR #4550, opened 2017-11-03
Force node v7 for precise/PHP 5.3 tests

pglewis:
Fixes #4547

PR #4551, opened 2017-11-03
See if we can drop the forced nvm command

jamesgol:
I believe travis will properly setup the environment


jamesgol:
Closed in favor of #4550

Issue #4445, opened 2017-09-04
select2  set delay option (Performance Improvement)


jneto81:
I need a way to set the delay option, because now its 100 ms by default  and my wordpress almost die using the select2 search because it send several ajax request to get the response. ( i have a large posts table )

pglewis:
This may be a good candidate for an advanced setting to override the default; whatever default we choose is likely to be sub-optimal for someone.

jamesgol:
I think a filter to allow the select2 options to be adjusted is the right way to go.  That will allow advanced users to adjust without cluttering up the UI with too many options that most people don't use

pglewis:
PR: #4555

pglewis:
Done, and will be improved in #4561

PR #4555, opened 2017-11-06
Feature/select2 ajax option filters

### LINKED ISSUE #4444: select2 minimumInputLength (Performance Improvement)
I need a way to set the minimumInputLength per field ( can be a hook )

`                minimumInputLength : 1
`
Default value is 1, and with a large number of posts, it takes a lot of time.

pglewis:
Fixes #4444 and #4445


pglewis:
This is just a minimal patch to support this for 2.7 release.  @jamesgol has a more flexible approach that will support arbitrary select2 options to be customized via filter, I'm just too paranoid to jostle how we work with the JS options this close to release.  

For now, this will do the trick: 

```add_filter( 'pods_field_dfv_data', 'set_select2_options', 10, 2 );
function set_ajax_delay( $data, $args ) {
	$data['fieldConfig']['ajax_data']['delay'] = 1000;
	$data['fieldConfig']['ajax_data']['minimum_input_length'] = 3;

	return $data;
}```

You can just opt-in when we re-organize how we do this and the above should continue to work either way.

File: classes/fields/pick.php [hunk 1]
```diff
@@ -872,7 +872,7 @@ public function build_dfv_field_options( $options, $args ) {
 
 		$options[ $args->type . '_limit' ] = $limit;
 
-		$options[ 'ajax_data' ] = $this->build_dfv_autocomplate_ajax_data( $options, $args, $ajax );
+		$options[ 'ajax_data' ] = $this->build_dfv_autocomplete_ajax_data( $options, $args, $ajax );
```
sc0ttkclark: That's a weird diff

classes/fields/pick.php
```diff
@@ -895,7 +895,7 @@ public function build_dfv_field_options( $options, $args ) {
 	 *
 	 * @return array
 	 */
-	public function build_dfv_autocomplate_ajax_data( $options, $args, $ajax ) {
+	public function build_dfv_autocomplete_ajax_data( $options, $args, $ajax ) {
```
sc0ttkclark: That's a weird diff

classes/fields/pick.php
pglewis: I may have made some mention about reading glasses....

PR #4558, opened 2017-11-08
Placeholder for select2/single-select requires an empty selected option

### LINKED ISSUE #4457: 2.7 RC-1: Single select autocomplete fields have a default value
Using 2.7 RC-1, when creating a new record in WP-Admin for a Pod which has a single select autocomplete relationship field, the field will have a default value pre-set.

It seems that the default value set for the field is the first related entity by alphabetical order.

screen shot 2017-09-08 at 12 24 23

pglewis:
Fixes #4457

https://select2.org/placeholders#single-select-placeholders
https://github.com/select2/select2/issues/3553


lougreenwood:
@pglewis Not sure if I've found a bug:

* In WP Admin
* using an existing post which is linked to some other post via bi-directional relationship
    * single select, autocomplete Select2 field
    * not required
* click 'x' icon to clear selection
* click 'update'
* see that after save the record is still linked to the related post
    * expected to be un-linked because select2 input was cleared

pglewis:
@lougreenwood: Are you testing on this branch?  (it's not merged into 2.7 yet) 

Make sure to cache-bust any old script copies in the browser?  

I tested this with single-select autoselects on both ends of a pair of bi-di posts and everything seems to be working correctly in a quick run-through.

lougreenwood:
@pglewis my bad, the Pod I was testing with had some code which ensure it's always linked to a record (so creates a new item to relate to if empty) - that was triggering this 'bug' :)

Approved :) Thanks!

PR #4559, opened 2017-11-08
Add new always active

### LINKED ISSUE #4469: Add New button is disabled for some relationships
We should keep "Add New" always active, if they click it and add a new item it should just replace the selection. So for a dropdown with "Something" already selected, currently the "Add New" button is disabled. It should always be enabled so if they add a new item then the "Something" selection will be changed to the new item they just added.

Just a usability concern here, not a huge problem.

The bigger problem is that when you currently go back to "-- Select --" the "Add New" button does not become available. That should be solved by always having "Add New" enabled.

pglewis:
Fixes #4469


pglewis:
Not ready to merge just yet, single-select List View does not replace the selection.

sc0ttkclark:
Just confirming, this still means that Add New won't show up for simple relationships right? Or if they don't have permission to Add New?

pglewis:
> Just confirming, this still means that Add New won't show up for simple relationships right? Or if they don't have permission to Add New?

Simple relationships _should_ always disable the add new option.  

I believe this was just getting rid of cases where we show an add new button but disable it.

pglewis:
> Not ready to merge just yet, single-select List View does not replace the selection.

This one is not quite complete just yet

JoryHogeveen:
@pglewis Any idea when you can finish this PR so we can merge/close it?

pglewis:
I nominate this as a candidate to just close and deal with it in 2.8.  It's not a bug, per se, as no one is waiting on it to fix anything and 2.7 is reaching end of life tbh.

PR #4561, opened 2017-11-09
Add filter for select2 overrides

sc0ttkclark:
Addresses what #4444 and #4445 is asking for long-term for flexibility


sc0ttkclark:
I can't approve a PR I created but this one is ✅

bkonia:
This filter works well for overriding simple options. However, it doesn't work for the callback options, because it passes the function name as a string, but Select2 is expecting it to be passed as a function (without quotes).

pglewis:
@bkonia: this is a close pull request, you should open a new issue for this.

Issue #4571, opened 2017-11-15
Add .editorconfig file


JoryHogeveen:
Simple and convenient to have better code consistency between users and their tools.

http://editorconfig.org/

PR #4569, opened 2017-11-15
PHP Codesniffer fixes

sc0ttkclark:

Issue #4216, opened 2017-05-18
Update CONTRIBUTING.md with coding standards


PodsBot:
Update CONTRIBUTING.md with coding standards submitted via Slack by keraweb

JoryHogeveen:
@sc0ttkclark Pods should be using 100% WP coding standards or are there modifications?

jimtrue:
@JoryHogeveen Until you get the list of items you want in here, don't assign it to me yet ;) I tossed you guys on here, because you know which of the tabs/spaces stuff you'd want included in here as well.

JoryHogeveen:
Oh sorry, I didn't meant to assign it to you. This is more are general issue so we can determine the actual code standards for Pods. Also to be used for code checkers (CodeClimate for example).

jimtrue:
Makes sense, that's why I assigned all of us ;)

quasel:
thats solved I guess ;)

Issue #226, opened 2012-08-12
Can't edit advanced content type after creation


pglewis:
JS alert: "Error: created is reserved for internal Pods usage, please try a different name."  

Error is triggered by Pods' internally created field.

PR #4572, opened 2017-11-15
Add .editorconfig

JoryHogeveen:
Related: #4569 #4216 
Fixes: #4571 

`spaces_around_operators` and `spaces_around_brackets` not available yet.

Comment from @pglewis 
Maybe remove the tab size options. No problem if the user has their own preferences. **Only issue it _could_ cause is when dev's are using precision alignment with spaces.**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,10 +9,7 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
 indent_style = tab
-indent_size = 4
-tab_width = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
 # spaces_around_operators = true
@@ -21,6 +18,6 @@ trim_trailing_whitespace = true
 [*.{txt,md}]
 trim_trailing_whitespace = false
 
-[*.yml]
+[*.{yml,json}]
 indent_style = space
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+# https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties
+
+# WordPress Coding Standards
+# http://make.wordpress.org/core/handbook/coding-standards/
+
+root = true
+
+[*]
+charset = utf-8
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+# https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties
+
+# WordPress Coding Standards
+# http://make.wordpress.org/core/handbook/coding-standards/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
```
lougreenwood: From the editorconfig docs:

>end_of_line
> Line ending file format (Unix, DOS, Mac)
>The values are case insensitive. They will be lowercased by the core library.

>NOTE: if you want to use native line endings between different operating systems it is better not to set this option and leave that task to the VCS! In the future we might add a value like native for this scenario (cf #226).

So maybe should be omitted?

JoryHogeveen: For PHP it shouldn't matter but since v2.7 is will be required for JS because of the DFV structure templates templates. (@pglewis)

pglewis:
> Comment from @pglewis
> Maybe remove the tab size options. No problem if the user has their own preferences. Only issue it could cause is when dev's are using precision alignment with spaces.

Alignment with spaces should not be affected by the tab size setting, thus why the WordPress whitespace rules use tabs _only_ for indentation at the start of a line.

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+indent_size = 4
+tab_width = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+# spaces_around_operators = true
+# spaces_around_brackets = both
+
+[*.{txt,md}]
+trim_trailing_whitespace = false
+
```
pglewis: Maybe we should consider a section for JSON as well (JSON specs are 2 space indent like YAML)

pglewis:
Also, technically YAML will detect whatever indentation is present and 2 space indent is only a suggestion.  I think I'd rather go ahead and normalize to 2 spaces anyway but both the `.scrutinizer.yml` and `.travis.yml` would need normalized (looks like we're using 4 spaces in each)

PR #4585, opened 2017-11-22
Fixed: Relationships to Advanced Content Types in [each] template tags.

0xLBF:
ACTs have a lowercase 'id' but while parsing the [each] template tags only the uppercase 'ID' was checked.


sc0ttkclark:
Good fix, thanks!

PR #4608, opened 2017-12-05
Update term field tooltips and labels

### LINKED ISSUE #4543: Custom Taxonomy Creation in v2.7
screen shot 2017-11-01 at 9 03 19 pm

I expect these options to be more yes/no based. Both seem to be "yes". 

The tooltip content says that WP taxes don't support this natively, then the first option appears to be a native solution which is contradictory to what the tooltip just said. It's confusing from a UX standpoint. 

The solution here is to prefix each option with "Yes, meta-based..." and "Yes, table-based" and "No, no extra fields".

JoryHogeveen:
Fixes #4543


sc0ttkclark:
Closing in favor of #4609

PR #4609, opened 2017-12-05
Update Pods Add New screen for extended taxonomies, the storage type is weird

### LINKED ISSUE #4543: Custom Taxonomy Creation in v2.7

sc0ttkclark:
Fixes #4543


pglewis:
We'll get the tooltip text straightened out via #4835

PR #4629, opened 2017-12-07
Add join table logic for taxonomy

### LINKED ISSUE #4613: Save Taxonomy Pod After Renaming Field Label/Name Produces SQL Error
I have two Pod-created taxonomies: Difficulty and Property.  In the Property taxonomy, I have defined 5 fields.  I tried to rename both the label and name of one of those fields from "thickness" to "width" and the following error shown in the screenshot was produce upon attempting to save the pod.  

screenshot 2017-12-06 04 42 27

sc0ttkclark:
Fixes #4613


pglewis:
Works for me, also did a quick tests of meta field renames for post, media, and user to double-check those.

Issue #4624, opened 2017-12-07
WP Core Style Conflict in _advanced.scss


napa-web-designer:
It looks like you are using a generic class (.left which is styled to float left) in your css that is conflicting with the WP Admin area styling and causing a formatting issue. See attached screenshots looking at WP admin area Custom Fields which is affected along with Chrome Dev tools showing the .left rule toggled on and then off, exemplifying the affect, and then the file pathway leading to the file _advanced.scss.

Steps to duplicate, I created a new Upload file custom field for the post type Posts. Then when I went to edit a post and was showing custom fields, I got what is in the screenshots.

<img width="1280" alt="screen shot 2017-12-07 at 10 00 08 am" src="https://user-images.githubusercontent.com/1827904/33732813-7b08b2fa-db3c-11e7-8e08-17aee6b02865.png">
<img width="1280" alt="screen shot 2017-12-07 at 10 00 23 am" src="https://user-images.githubusercontent.com/1827904/33732814-7b255356-db3c-11e7-83ea-2501eec771eb.png">
<img width="563" alt="screen shot 2017-12-07 at 10 01 23 am" src="https://user-images.githubusercontent.com/1827904/33732815-7b44c72c-db3c-11e7-8478-62a051bab198.png">

pglewis:
Related #4612 

Should probably be added to #4615

nicdford:
Fixed with #4632

nicdford:
Verified this was fixed with #4632

Issue #4612, opened 2017-12-06
Pods jquery ui css messing with gravity forms


saltnpixels:
Eric (shamai) [9:03 PM] 
https://www.dropbox.com/s/vpk354ghlpmkev3/Screenshot%202017-12-05%2021.02.54.png?dl=0
jquery ui of pods messing with gravity forms edit screen
![screenshot 2017-12-05 21 02 54_preview](https://user-images.githubusercontent.com/3052875/33640821-c9fd8256-da00-11e7-841d-7f9dc09fad94.png)

how it should look:
https://www.dropbox.com/s/ajvci2p9ljsnnhf/Screenshot%202017-12-05%2021.02.20.png?dl=0
![screenshot 2017-12-05 21 02 20_preview](https://user-images.githubusercontent.com/3052875/33640841-ea7675ce-da00-11e7-81d9-ad5687d109e5.png)

how it does look:
https://www.dropbox.com/s/6gu5jdxo0f6bffa/Screenshot%202017-12-05%2021.05.42.png?dl=0
![screenshot 2017-12-05 21 05 42_preview](https://user-images.githubusercontent.com/3052875/33640857-fb864182-da00-11e7-8a96-90f3f9a04756.png)

sc0ttkclark:
@nicdford @pglewis

Looks like during the SCSS refactor we started pulling ALL of our CSS library code any time we need any of our CSS -- it's now one big file /ui/styles/src/pods.scss that gets built from.

Here's the exact diff line where this is made clear:

https://github.com/pods-framework/pods/pull/4455/files#diff-e34156692b93ef7d263924ac927338f2L238

That's PodsInit.php at line 238 in register_assets where lines like these are removed where we were registering stylesheets only if they weren't registered before (in case WP adds it's own jquery-ui -- a good check)

```php
if ( ! wp_style_is( 'jquery-ui', 'registered' ) ) {
    wp_register_style( 'jquery-ui', PODS_URL . 'ui/css/smoothness/jquery-ui.custom.css', array(), '1.8.16' );
}
```

That's the case for this and other libraries now. We no longer check if they exist already, we just load our CSS up and it starts to work.

One fix should be that we wrap our `@import` for the libs scss with a class like:

```css
form.pods-form, form.pods-admin {
	@import "libs/jquery-ui-custom";
	@import "libs/jquery-qtip";
}
```

Coincidentally, we need to do this same thing for our overrides too:

```css
form.pods-form, form.pods-admin {
[...]

PR #4632, opened 2017-12-08
Feature/4612 prefix css various

nicdford:
Fixes specificity issues revealed with relation to #4612


nicdford:
The only set of styles I wasn't able to test was 4bb2877 because I can't trackdown the timepicker html this is being applied to, but I wrapped it in the same `form.pods-form`, so hopefully that does the trick.

nicdford:
26c9c19 fixes #4624 

Similar to the timepicker, I couldn't find where these advanced.scss styles are getting applied? I tried to track it down, but without luck. I'd rather remove them to fix this and have another issue popup and fix it properly with specificity, but since I don't know where these's helpers are getting used, it's hard to come up with a perfect fix.

sc0ttkclark:
@nicdford please remove the wrap within the lib files, let's wrap them around the `@import` instead.

nicdford:
@sc0ttkclark I really don't think we should use @import to wrap the tags, it's better if we can actually prefix each css line ie `.pods-codemirror`. This fix doesn't address that goal specifically, but I see it as a short-term fix as we start rebuilding more of the CSS to follow a better/more consistent prefixing standard. The goal of the pods.scss file is to keep it as an index of all the imports and not as a place for any actual styles so we can easily see what is getting imported and what isn't.

pglewis:
Immediate use for the new 'Team Discussion' label :)

We should work out a solid plan moving forward after 2.7.1 goes out.  Some points to consider in both directions.

JoryHogeveen:
I'm also not for wrapping around `@import` since it's not very dev friendly imo.

pglewis:
> The goal of the pods.scss file is to keep it as an index of all the imports and not as a place for any actual styles so we can easily see what is getting imported and what isn't.

This definitely makes a lot of sense from a high level, organizational perspective.  

The opposite side is that wrapping the imports with our prefixes is a convenient DRY way to have a single point to change if those prefixes change, vs. search replace as well as a quick way to wrap generic styles sections like the jquery-ui stuff without having to go hacking up the whole stylesheet.

nicdford:
Would it make sense to abstract third party code bases to make them more modular? Something like creating a new partial in libs that we have the @import wraps there? so like:
```
_jquery-qtip.scss
_jquery-ui-custom.scss
_libraries.scss
```
Then in `pods.scss` we'd just import `libs/libraries`? This would maintain the structure of `pods.scss`, allow us to then do the @import for the external libraries, and also theoretically would enable us to update libraries without having to re-prefix everything?

I think this makes sense for external libraries/overrides. I think for base styles it would make sense to base them on the html structure created for the module, like with dfv everything is prefixed with `pods-dfv-list`.

sc0ttkclark:
theoretically, we'd want to wrap all libraries and qtip/jquery-ui stuff. So _libraries.scss would contain that too.

nicdford:
> single point to change if those prefixes change, vs. search replace
Also, @pglewis, for base styles that pods is creating (like dfv) we can do some cool stuff with [storing prefixes in variables](http://www.scottohara.me/article/sass-selector-variables.html) with sass which might also be useful.

nicdford:
> theoretically, we'd want to wrap all libraries and qtip/jquery-ui stuff. So _libraries.scss would contain that too.

Yeah, totally!

pglewis:
> I'm also not for wrapping around @import since it's not very dev friendly imo.

I'd like to understand the downside(s) to this approach to see if we can come up with the "best of all worlds".  

The downsides of wrapping the entire file contents for me, like we're currently doing:

*  Interrupts the git blame history, as every line in the file is seen to be changed by Nic in this commit: https://github.com/pods-framework/pods/blame/26c9c19/ui/styles/src/libs/_jquery-qtip.scss
* Creates an extra nesting level around all 300 lines, when possible I really really prefer to see the full nesting structure for something in one screenful
* Introduces changes to a lot more files than wrapping the imports that need it
* Distributes the placement of prefixes among more files which can lead to more complicated search/replace in the future

The way I've been viewing our 3 main style files:
* pods: all the admin css
* pods-form: smaller subset for use with Pods Form on the front-end
* pods-wizard: made into its own bundle since we only need it in a few places and cuts the main file size

With that in mind, the three "main" .scss files basically represent the "build instructions" for their compiled counter-parts and it kinda feels to me like the right, DIY place to be wrapping our generic includes with name-spacing specifiers.  

I'm not sure another layer of partials with multiple libraries in it is a better choice, that makes the import of /it/ all or nothing.  Seems a lot of hoops to jump through just to avoid wrapping anything in the main file, but there may be good arguments for doing that, that I just don't see yet.

pglewis:
Changelog: add appropriate namespace prefixes to avoid style conflicts (#4612)

PR #4742, opened 2018-02-03
Fixed printing numbers having more than 6 decimals with sprintf. #4741

### LINKED ISSUE #4741: Numeric field in Advanced Content Type rounded to 6 decimals
I have a "Plain number" field with the following settings
Format: 1,234.00
Decimals: 8
Maximum Length: -1

Input value: 0.12345678
The number is rounded to 0.12345700

The field looks correctly setup in database: DECIMAL(64,8)

This only happens with the Advaned Content Type fields. With other types (custom post types/taxonomies) the number is not rounded to 6 decimals

0xLBF:
Fixes #4741 , not sure if there are unintended consequences though it seems fairly harmless.

### DIFF
--- a/classes/fields/number.php
+++ b/classes/fields/number.php
@@ -184,7 +184,10 @@ class PodsField_Number extends PodsField {
 
 		$decimals = $this->get_max_decimals( $options );
 
-		if ( 0 < $decimals ) {
+		if( 6 < $decimals ) {
+			// %F only allows 6 decimals by default
+			$format = '%.' . $decimals . 'F';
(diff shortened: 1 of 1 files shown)


pglewis:
I'm going to do a quick search to see if there are any other spots in the code that could use the same fix, but good find!

pglewis:
Changelog: `%F` with PHP's `sprintf` function defaults to a precision of 6 decimal places, we now include the precision specifier whenever more than 6 decimal places of precision is needed.

Issue #4770, opened 2018-02-25
Make sure 'new' terms don't exist before `wp_insert_term`


mgratch:
Before `wp_insert_term` is fired in [pods/classes/PodsAPI.php#L649](https://github.com/pods-framework/pods/blob/2.x/classes/PodsAPI.php#L649) it should check to make sure the term infact does not exist.

Backstory: I was using the GF + Pods extension to create a custom post type "pod item".  I needed the ability to add tags. I used another plugin to create a "taggable" field, essentially replacing a text field with a dynamic multi-select field that allowed adding custom values.  This seemed to handle the behavior I was looking for until I submitted my first test entry.  Where I was presented with a white screen stating: `A term with the name provided already exists in this taxonomy.`

So I followed the logic of the Pods + GF extension -> Pods -> WP Core where I found the above error comes from.  The Pods + GF extension logically takes all the data from the submitted form and passes it to `Pods->add`, eventually it gets to the term field and because a `term_id` is not passed it assumes the term doesn't exist. This assumption is what brings me here.

mgratch:
See #4771

PR #4771, opened 2018-02-25
Make sure 'new' terms don't exist before `wp_insert_term`

mgratch:
This might not be the most performant but it solves the problem. Honestly I don't know if there is a more efficient answer.


classes/PodsAPI.php
```diff
@@ -701,7 +701,23 @@ public function save_term( $term_data, $term_meta, $strict = false, $sanitized =
 
 			unset( $term_data['name'] );
 
-			$term_data['term_id'] = wp_insert_term( $term_name, $taxonomy, $term_data );
+			// Check to be certain the 'new' term doesn't exist.
+			if ( is_int( $term_name ) ) {
```
sc0ttkclark: Is this necessary for your use-case?

classes/PodsAPI.php
```diff
@@ -701,7 +701,23 @@ public function save_term( $term_data, $term_meta, $strict = false, $sanitized =
...
+				$term = get_term_by( 'term_id', $term_name, $taxonomy );
+			} else {
+				$term = get_term_by( 'name', $term_name, $taxonomy );
+
+				if ( ! $term ) {
+					$term = get_term_by( 'slug', $term_name, $taxonomy );
```
sc0ttkclark: Is this necessary for your use-case?

classes/PodsAPI.php
sc0ttkclark: @mgratch

classes/PodsAPI.php
sc0ttkclark: @mgratch

mgratch: So after reviewing this further, this PR seems unnecessary. Core adds terms by the same process and doesn't do any last minute checks.  Validation should happen before this, in my case in another plugin.

pglewis:
What's the best label here, "out of scope" @mgratch?  I'll label it and close it out

pglewis:
Also, I assume the related issue #4770 can be closed as well?

mgratch:
Yes, the issue can be closed as well :)

sc0ttkclark:
Thanks!

PR #4777, opened 2018-03-01
Expose `IframeFrame`

### LINKED ISSUE #4776: DFV: Expose the IframeFrame object so it can be reused externally

pglewis:
Fixes #4776


pglewis:
The IFrame isn't really a field so `PodsDFV.fields` in the first commit isn't the correct place place to put it.

pglewis:
I need to modify where we expose the iFrame and we've also merged formatting changes into DFV since this PR was made.  This branch will be deleted and we'll do a fresh one.

pglewis:
This is on hold, "fields" is not the proper place to expose it.  No rush though, I don't believe @JoryHogeveen has tried to use this yet.

pglewis:
This hasn't had a high priority to test so I'm going to close this out and do another PR after 2.7.2 goes out.

Issue #4298, opened 2017-06-11
Feature: New Field Type Dashicon/FontAwesome/Custom Icon Pack Support


PodsBot:
Feature: New Field Type Dashicon/FontAwesome/Custom Icon Pack Support submitted via Slack by jimtrue

jimtrue:
https://github.com/bradvin/dashicons-picker jquery Dashicons Picker
https://github.com/micc83/fontIconPicker jQuery Font Icon Picker

JoryHogeveen:
One "problem" with this is that these library change constantly (especially Font Awesome) so we'll have to maintain this very actively.

> Allow use of Custom Icon Packs loaded through CSS or other methods (Filter?)
A filter is required in that case since the PHP side won't have any knowledge of any CSS classes that are available.

SVG support would be a simple (image) file uploader.

jimtrue:
We could use the CDN, that's the typical way with FontAwesome (and some others). And yep, I'd also be pointing towards an icon picker that supported multiple libraries of font icons.

JoryHogeveen:
https://github.com/ineagu/wp-icon-picker

JoryHogeveen:
@sc0ttkclark Where would you add external github modules in the folder structure?
We currently do not setup any .gitmodules links anywhere as far as I can see.

jimtrue:
Aren't these particular fonts typically just loaded through CDN stylesheets?

JoryHogeveen:
Yes, that is unrelated to the above module :) That module can handle all the Icon/SVG selection stuff and through filters you can choose to load locally or from CDN.
Locally is default since it's better for SEO and performance.

PR #4782, opened 2018-03-05
Feature: New icon picker field

JoryHogeveen:
Issue: #4298 

It will add a composer dependency for WP Icon Picker library:
https://github.com/ineagu/wp-icon-picker

### DIFF
--- a/classes/PodsForm.php
+++ b/classes/PodsForm.php
@@ -1419,6 +1419,7 @@ class PodsForm {
             'pick',
             'boolean',
             'color',
+            'icon',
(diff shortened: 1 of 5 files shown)


sc0ttkclark:
Can we store formatted separated by pipe "|"?

Like "dashicons|the-icon"

JoryHogeveen:
I don't see why not. A space would work even better I guess.

vgstef:
Is there any developement or a roadmap to add this icon picker in Pods?

PR #4796, opened 2018-03-11
Avoid printing null message in PodsUI

davidatwhiletrue:
The `pods_ui_message` filter hook in `pods_ui` allows to change the message being printed in the admin notice box. In case the returned message is `null` the box shouldn't be printed at all.


jimtrue:
Changelog: Avoid printing null message in PodsUI #4796 (@davidatwhiletrue)

Issue #4802, opened 2018-03-14
WPML: cannot save translations with the translation editor if there are required fields


dgwatkins:
## Issue Overview
When you enable the Translation Editor in WPML you are not able to save a translation that contains required custom fields.

## Expected Behavior
Saving a translation with the translation editor fires an AJAX request and when completed we should be redirected to the previous page.

## Current Behavior
Right now, Pods is replacing the AJAX reply with a validation error of its own (required field) that WPML doesn't understand. Because of this saving the translation doesn't complete.

## Debug
The problem is in the way WPML saves updates to the translated post custom fields.
When we save a translation, WPML is deleting all previous post meta and adding the updated post meta. But Pods doesn't like that we remove a required post meta and shows a validation error.

## Possible Solution
You are already making an exception in validation for WPML ajax calls, it is in https://github.com/pods-framework/pods/blob/2.x/classes/PodsAPI.php#L3628. But it doesn't cover saving with the Translation Editor. The solution would be to extend this condition to cover the Translation Editor too.

## Steps to Reproduce (for bugs)
* Install and configure WPML and Pods
* Add a required custom field in Pods
* Enable the Translation Editor (WPML -> Translation Management -> Multilingual Content Setup)
* Create a post containing the required field
* Try to translate it, you will see it stalls when saving the translation

## Related Issues and/or PRs
I will submit a PR straight after creating this issue.
This issue was originally reported in https://wpml.org/forums/topic/translation-wont-save-with-custom-fields-in-it-pods/

## Todos
- [ ] Tests
- [ ] Documentation

PR #4803, opened 2018-03-14
Extend the condition for WPML validation exception to cover the TE

dgwatkins:
Corresponds to issue #4802


jimtrue:
ChangeLog Blurb:
* Fixed: Extend the condition for WPML validation exception to cover the Translation Editor so required fields don't break the Translation Editor in WPML. (#4802) (@dgwatkins)

PR #4811, opened 2018-03-27
travis-ci: test with php 7.3

Ramoonus:


pglewis:
7.3 should probably be added to the matrix. (see also: #5376)

Ramoonus:
sure

JoryHogeveen:
What is the status for this?

This PR also drops PHP 5.3 but doesn't says it does in the title or description.
For backlog I'd say either change the title or only add PHP 7.3 without dropping PHP 5.3

PR #4819, opened 2018-03-31
Update phpcs.xml.dist for WP 4.5 and PHP 5.3

GaryJones:
Pods 2.7 requires WP 4.5 and PHP 5.3.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -9,17 +9,13 @@
 	<exclude-pattern>vendor/*</exclude-pattern>
 	<exclude-pattern>components/Markdown.php</exclude-pattern>
 
+	<config name="minimum_supported_wp_version" value="4.5"/>
 	<rule ref="WordPress-Core"/>
 	<rule ref="WordPress-Docs"/>
 	<rule ref="WordPress-Extra"/>
-	<rule ref="PHPCompatibility"/>
 
-	<rule ref="WordPress.PHP.DiscouragedPHPFunctions">
-		<!-- From "Extra": The create_function group is excluded as WP core still supports PHP 5.2 and 5.2 does not support anonymous functions. -->
-		<properties>
-			<property name="exclude" value="create_function"/>
-		</properties>
-	</rule>
+	<config name="testVersion" value="5.3-"/>
+	<rule ref="PHPCompatibility"/>
 
 	<rule ref="Squiz.Commenting">
 		<!-- Excluding the need for a full stop at the end of a comment (common, harmless error) -->


pglewis:
PR either needs to be done again after #4818 is merged or merge back up from 2.x after it has.

GaryJones:
I can rebase as needed.

pglewis:
The Travis test hiccups should be ironed out in 2.x so I think this one is good to go once rebased with that change.

phpcs.xml.dist
```diff
@@ -9,17 +9,13 @@
...
 	<rule ref="WordPress-Extra"/>
-	<rule ref="PHPCompatibility"/>
 
-	<rule ref="WordPress.PHP.DiscouragedPHPFunctions">
-		<!-- From "Extra": The create_function group is excluded as WP core still supports PHP 5.2 and 5.2 does not support anonymous functions. -->
-		<properties>
-			<property name="exclude" value="create_function"/>
-		</properties>
-	</rule>
+	<config name="testVersion" value="5.3-"/>
```
sc0ttkclark: Why remove exclusion for DiscouragedPHPFunctions rule?

GaryJones: With the project as PHP 5.3+, removing this exclusion will flag current and future instances of `create_function` that should be replaced with anonymous functions.

sc0ttkclark: We don't use it directly, but the Markdown PHP parser we use does. Will need to look at using a different markdown parser in the future.

GaryJones: Since it only seems to be a [single instance](https://github.com/pods-framework/pods/blob/a404a389c192849ded643c3a0925de213c5e1fdd/components/Markdown.php#L1996), and [edits](https://github.com/pods-framework/pods/commits/a404a389c192849ded643c3a0925de213c5e1fdd/components/Markdown.php) have been made to that embedded dependency, I'd suggest adding a `// phpcs.ignore WordPress.PHP.DiscouragedPHPFunctions` to the end of the line (or an empty line before it), so that you're explicitly dealing with _just that instance_ of `create_function()`, and not other instances that might be introduced elsewhere.

pglewis:
@sc0ttkclark, @GaryJones: Are we all good to go here?  

I'm fine merging this in as is.  There's nothing that affects executable code, it won't even be distributed in the plugin download, and the test breakage is known to be elsewhere.

GaryJones:
@pglewis I'm fine with this still. As noted in my last [comment](https://github.com/pods-framework/pods/pull/4819#discussion_r183228418), there may be instances of things to fix, but this particular PR is about updating the config.

PR #4831, opened 2018-04-10
Fixes overlapping fields issue when content length is longer than normal

### LINKED ISSUE #4700: Backend - overlapping fields from middle (contents) columns disable right column actions (e.g. Update)
When you are editing a post, if you use a narrow Windows pane (such as the auto-half screen), the WordPress "middle column" edit boxes may "overlap" the right hand column and suppress the operation of contents therein. (Like Update, Move to Bin, or Category seleciton)

These images from TwentySeventeen Theme (responsive front and back end)

image

image

nicdford:
Fixes #4700


nicdford:
Changelog: Fixed issue where fields with particularly long post titles would extend past the width of the container

Issue #4832, opened 2018-04-10
Tooltips not displaying in 2.x


pglewis:
It appears to be introduced in 7d7cb0b.  The other PR was reverted by 9705add but it appears tooltips did not start working again.

pglewis:
It's the helper classes being removed: 26c9c19

pglewis:
Possibly related #4781

pglewis:
PR: #4834

PR #4834, opened 2018-04-10
Create container for  qTip tooltips

nicdford:
Isolates qtip styles and functionality to pods only which should hopefully resolve any issues with conflicting with other plugins/themes that may use the tooltip functionality.


pglewis:
Fixes #4832

PR #4846, opened 2018-04-18
codemirror fixes for Pods Templates

### LINKED ISSUE #4845: Codemirror CSS display issues
Likely introduced in #4612 and or #4615

pglewis:
Fixes #4845


jamesgol:
Can confirm this does fix the CodeMirror issue

pglewis:
The only namespaced wrappers we have around the CodeMirror box on a Pods Template page are `.pods-submittable .pods-form`, so the `div.pods-form-ui-field` does not do the trick.  

It's possible we just need to swap out `div.pods-form-ui-field` for `.pods-form` instead of my quickie fix.  

@JoryHogeveen, @nicdford: Let me know what you think the best course of action is.

pglewis:
Also, even `.pods-form` isn't enough... the form containing the field autotemplate options doesn't have the .pods-form class on it, only `.pods-submittable`

pglewis:
I've widened the scope on this a little to address some more qtip issues I discovered as well.  Closing this in favor of a new PR

Issue #4752, opened 2018-02-09
User Meta in Multiple-Select Traversed User Relationships is Not Displaying In Pods Templates


jimtrue:
Related to #4484 I believe and #4723 term meta not displaying.

jimtrue:
I'll be adding a configuration and template test to work with this one.

jimtrue:
Additional Tests and Detail:

1.  Single Traversal displays fine, ie {@related_user.meta_field} works.
2.  Direct Traversal from the User Pod (ie in a shortcode directly against User) also displays fine, {@meta_field} in a `[pods name="user"]{@meta_field}[/pods]`
3. Problem seems to be specific to Related User in an each block (multiple select), ie
`[each related_user]
<li>{@display_name} - {@meta_field}</li>
[/each]
`
The direct user meta will display, but not the traversed user meta. Migrate Packages Import attached

[user_meta_each_4752_json.txt](https://github.com/pods-framework/pods/files/1807780/user_meta_each_4752_json.txt)

jamesgol:
After spending some time looking into this I thought I would add my notes since I'm out of time for now.  This does not appear to be exactly related to the template tags but to a limitation in Pods->field() when querying for an array.  It only returns the WordPress user data and not all data (including extended data like I would expect).

This is likely a shortcoming either in PodsData->build() or Pods->field()

A possible workaround would be to edit components/Templates/includes/functions-view_template.php:frontier_do_subtemplate() so the call is $pod->field( array( 'name' => $field_name, 'output' => 'pod' ) );

The frontier_do_subtemplate() function would have to be modified to deal with the data in that fashion and I didn't have time to finish that up.

jimtrue:
Can confirm #4847 fixes this issue

PR #4847, opened 2018-04-18
Allow user-meta (and other extended Pods) to work in templates

jamesgol:
In the else case which was handling users before the Pod extended data was not available because calling Pods->field() on a relationship to a user field only returns the WordPress core fields

Related to #4752


jamesgol:
Would really like to add some tests related to this but not sure I have time.  This should not break anything that wasn't already broken

pglewis:
Just holding on me to look into adding a test or three now.

pglewis:
Test can wait for another time.

PR #4851, opened 2018-04-19
Qtip init was registered but not enqueued in certain cases

### LINKED ISSUE #4850: Tooltips not working in Pods Template or Roles and Caps Components
I've spot-checked back as far as 2.6.6 and have not found it working at any point.  

The `pods-qtip` class is there, it's inside a `pods-submittable` container, there is associated help text and it's getting put into the image alt attribute correctly.  

Chances are we're just not enqueueing and/or initializing.

pglewis:
Fixes #4850

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -489,6 +489,7 @@ function pods_help( $text, $url = null ) {
 				'jquery-qtip2',
 			), PODS_VERSION
 		);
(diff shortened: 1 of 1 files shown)


pglewis:
Changelog: Tooltips were not working in the Pods Templates or Roles and Capabilities components (#4850)

Issue #4853, opened 2018-04-21
Improve Relationship CRUD and Meta


pcfreak30:

quasel:
Could you bee a bit explain that a bit more to make it an actionable item? 

Sorry for the late reply, feel free to comment if you still have questions!

pcfreak30:
@quasel I generally make blank issues  when i intend to do a PR to use it in the feature branch name.

quasel:
oh missed the PR sorry issue scrubbing all day :(

PR #4854, opened 2018-04-21
Improve Relationship CRUD and Meta

pcfreak30:
So there appears to be a scaling problem with relationships due to the saving of meta and the saving of weights.

* In both `delete_relationships` and `save_relationships` you are delete all meta records and re-add. This can not scale with links that are in the thousands or more. It should only need to delete the specific item or add whats new.
* In  `save_relationships` you are running a php loop to save the ordering in a relationship. This means thousands of queries potentially and can kill performance. Based on research of mysqli extension, you can use a counter variable in mysql and have a single query for mysql to handle the re-ordering in one go.

The results here cause a system that can scale indefinitely ideally. The issue came to my attention due to a client site that is a internal task management system. They have thousands of records in complex relationships. The adding of new items was causing a 30 second load every single time. This refactor dropped it to no more than 3 seconds.

Thanks.

#4853


sc0ttkclark:
Does this PR cover deleting related IDs that were removed?

pcfreak30:
Yes? delete_relationships is modified and simplified. i dont see any reason you would delete relationships inside save_relationships. As is you do not want to mass delete and rebuild meta on every method call, its insane INHO.

Am i missing something?

pcfreak30:
Yes so as you can see overall, if you keep their responsibilities distinct you can remove database overhead.

pcfreak30:
@sc0ttkclark what needs to be done on this PR?

sc0ttkclark:
delete/save relationships needs more refactoring, it's got too much duplicated code and not sharing responsibilities properly yet. I plan on working on this, it's just in my work backlog which I may have time for this weekend.

quasel:
i see :D seems more complicated or it fell through the cracks?

pcfreak30:
@sc0ttkclark would like to see this get in sometime soon :P. Would help us all.

PR #4856, opened 2018-04-22
Support for JSON/YML configuration files in theme and other custom locations

sc0ttkclark:
Like Pods Packages, but automatic and more versatile.

### Example files

https://gist.github.com/sc0ttkclark/95d9b5b5c497ffc6e0fdd7b01238153f

### Features

* Supports theme configs (works for both json or yml files)
  * theme/pods.json
  * theme/pods/pods.json
  * theme/pods/fields.json
  * theme/pods/templates.json
  * theme/pods/pages.json
  * theme/pods/helpers.json
* Supports child / parent theme locations
* Supports additional customizable types of configs (my-whatever.json) with filters/actions to handle processing of them as needed
* Supports registration of custom paths for looking for the configs (register a path in a custom plugin folder, for a plugin that has it's own configs)


dedanirungu:
I can see the feature has been in active development for 4 years. I am not sure when this feature will be completed.
My question is;

Why not update the pods by checking changes in version number in the same way the plugins work. Auto-discovery might be complex and end up delaying the project.

dedanirungu:
I have worked on a plugin to perform pods migration that is saved as pods.json on themes or plugins.
https://github.com/dedanirungu/mybizna-pods-migration
I have added new functions to Config.php that has been in development and new functionality work as follows;

- Autodiscover pods.json in themes and those add via pods_config_pre_load_configs.
- prepare an array by converting JSON to an array.
- MD5 the array and compare previously saved MD5 string for previous migrations so as to determine if migration is required.
- Call add_pod at PodsAPI to save the pod
- Save new MD5 for the array.

I have only tested with pods with fields and groups only and it is working both for adding, and updating.
Note that, if MD5 is the same the pod is ignored.

sc0ttkclark:
## Here's where we're at

* Can register Pods, Fields (for one-off cases), Templates, Pages, and Helpers via JSON/YML configs
* Pods are loaded side-by-side with other Pods shown on Pods Admin > Edit Pods

## Things for the future, but that we can do after merging

* DB sync tool
	* Pull the file-based configs into the DB and allow for editing from there
	* Allow pushing to file when done
* Show other non-DB Templates/Pages/Helpers in custom PodsUI tables
	* Build custom add/edit forms to replace current post type screens used
* Determine file-based configs that do not seem to match DB configs (with link to DB sync tool to overwrite DB with file or update file to match DB)
* Update PodsUI to allow for name-based handling instead of ID-based handling, needed for additional action support for Pods Admin > Edit Pods and other areas

## What the Pods Admin > Edit Pods screen looks

Shown below are how Pods will look when there are multiple sources present. This includes:

* New source type filtering that lets you filter and see the counts of each source type (only shows if multiple sources are found)
* New source column only shows when there are non-DB sources (this includes some of the initial code implemented in `feature/update-how-code-registered-pod-group-fields-compatibility` which will go out in a maintenance release)
	* **DB** is for anything registered normally
	* **Code** is for anything registered manually via `pods_register_*()` functions
	* **File** is for anything registered through the new JSON/YML configuration handling in this PR
		* File sources also show the path from which the config was loaded from for easy reference
	* Any non-DB source has edit/delete actions disabled since they do not have IDs and cannot work with PodsUI yet

Screen Shot 2022-07-24 at 11 54 24 AM

Screen Shot 2022-07-24 at 11 45 18 AM

sc0ttkclark:
@dedanirungu Thanks for some extra eyes on this, your config files helped give me ideas on how to test this.

Ultimately, I have kept all DB syncing outside of scope for now. This will allow us to implement fully file-based configs as phase 1 and return to scope and build the DB sync side of things in a future release.

PR #4871, opened 2018-05-01
Disable the "Add New" button for relationships when we're inside a media modal window

### LINKED ISSUE #4864: Disable "Add New" button for media items from within grid mode edit
Now that #4795 is merged we get an Add New button in this context for relationships, however the CSS has not been tweaked to display well for a modal from a modal.  

This is a completely valid use case for the Add New functionality as it is an item edit, so In the long run we should iron out the CSS wrinkles so it displays well.  

Since we're this close to 2.7.2 the quick fix is to hide the button.

pglewis:
Fixes #4864


pglewis:
Changelog: Disable the "Add New" button for relationships when we're inside a media modal window (#4864).  This is just a workaround until we can iron out the CSS and display for a modal from a modal.

PR #4888, opened 2018-05-04
Allow all tags to use function callbacks just like within templates

### LINKED ISSUE #4887: Allow function calls in all tags, not just template tags.
## Issue Overview
Within templates it's possible to add a function call to a tag. `{@field,function}`
It would be neat to allow this with all pods tags.

## Possible Solution
Add callback handling within `pods_evaluate_tag`.

## Related Issues and/or PRs
PR: #4888 

## Todos
- [ ] Tests
- [ ] Documentation

JoryHogeveen:
Fixes #4887

### DIFF
--- a/includes/data.php
+++ b/includes/data.php
@@ -1539,6 +1539,13 @@ function pods_evaluate_tag( $tag, $sanitize = false ) {
 		$tag = $tag[2];
 	}
 
+	$callable = explode( ',', $tag );
+	if ( 1 < count( $callable ) ) {
+		$tag = array_shift( $callable );
+	} else {
+		$callable = false;
+	}
+
 	$tag = trim( $tag, ' {@}' );
 	$tag = explode( '.', $tag );
 
@@ -1599,6 +1606,12 @@ function pods_evaluate_tag( $tag, $sanitize = false ) {
 
 	if ( $sanitize ) {
 		$value = pods_sanitize( $value );
+	}
+
+	if ( $callable ) {
+		foreach ( $callable as $callback ) {
+			$value = call_user_func( $callback, $value );
+		}
 	}
 
 	return $value;


sc0ttkclark:
Is there a reason we don't recommend they use their own filter on `pods_evaluate_tag` here? Should we have a whitelist of allowed callbacks? Should callbacks run before sanitize?

JoryHogeveen:
I think it's way more convenient and easier to edit/change if this is possible without the `pods_evaluate_tag` filter. Besides, I don't think the code will allow passing anything through that filter as to what functions a user would like to call.
As for restricting functions, why would you want to do that? Yes, with wrong usage it could generate errors, but that is the responsibility of the user wanting to use a more advanced feature for the shortcodes (i.m.o.)

sc0ttkclark:
Not restricting callbacks would be a security problem for anyone to execute arbitrary commands by passing values into those functions for instance.

jimtrue:
Where else are they able to use Magic Tags outside of Custom Templates (the spaces between `[pods][/pods]` and Pods Templates, anyway?

sc0ttkclark:
Thats mainly it, though we use special magic tags in a number of other areas too. This code cannot happen without tightening up that aspect.

quasel:
@sc0ttkclark @JoryHogeveen any settlement on this?

sc0ttkclark:
There's a "disallowed" list of callbacks for magic tags, we should at minimal disallow them here too.

See: https://github.com/pods-framework/pods/blob/main/classes/Pods.php#L3693

JoryHogeveen:
@sc0ttkclark Good point. Maybe even better to create a general function for this instead of relying on a Pods instance?

JoryHogeveen:
This PR has become redundant since #5735 
Closing here.

@sc0ttkclark The mentioned PR uses `Pods::helper()` so that fixes the issue with disallowed helpers.

PR #4917, opened 2018-05-07
Handle saving related values if passed as an array of information

sc0ttkclark:
Examples below use a "book" pick field, multiple select, related to a "book" custom post type created or extended by Pods).

---

## Example 1

```
$pod = pods( 'author', 1234 );

$data = array(
    'books' => array(
        array(
            'post_title' => 'Book 1',
            'post_content' => 'Some content',
            'some_custom_field' => 'Value here',
        ),
    ),
);

$pod->save( $data );
```

Expectation would be that the above code would save a new value for the related book field.

---

## Example 2

```
$pod = pods( 'author', 1234 );

$data = array(
    'books' => array(
        array(
            'ID' => 1234,
            'post_title' => 'Book 1',
            'post_content' => 'Some new content',
            'some_custom_field' => 'New value here',
        ),
    ),
);

$pod->save( $data );
```

Expectation would be that the above code would update the data for the the related book field.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -4005,6 +4005,33 @@ class PodsAPI {
 								}
 							} elseif ( $is_file_field && isset( $v['id'] ) ) {
 								$v = (int) $v['id'];
+							} elseif ( 'pods' === $data_mode && ! isset( $v[0] ) ) {
+								// Save related field values.
+								$v_id = false;
+
+								if ( isset( $v[ $search_data->field_id ] ) ) {
+									$v_id = $v[ $search_data->field_id ];
+
+									unset( $v[ $search_data->field_id ] );
+								}
+
+								if ( ! empty( $v_id ) ) {
+									$search_data->save( $v, null, $v_id );
+								} else {
+									$row_has_values = array_filter( $v );
+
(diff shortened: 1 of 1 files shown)


JoryHogeveen:
@sc0ttkclark Could you give me a bit more info about the nature of this PR?
What was the bug and how can I replicate that to make sure this PR fixes that?

sc0ttkclark:
@JoryHogeveen To test this, use the example code. Perhaps we can write tests based on those two?

pdclark:
Tried example 1 without success.
Relationship field is relating "posts" to "posts", since Pods gives an error for "author" that it's a reserved name.
e.g., 
```php
$pod = pods( 'post', 1 ); // The first post; Hello World

$pod->save(
	[
		// Multi-select Relationship field titled "other_posts".
		'other_posts' => [
			[
				'post_title'     => 'Another Post',
				'post_content'   => '',
				'favorite_color' => '#33ee33', // Color Picker field
			],
		],
	]
);
```
Expectation was that it would create another post and assign it to the relationship field on post ID `1`.
Didn't work for me. Also didn't report any error and `->save()` returned `(integer) 1`.
@sc0ttkclark

Issue #4919, opened 2018-05-07
Non-required color field keeps returning validation error when empty


JoryHogeveen:
## Issue Overview
Since the PHPCS fixes the field validation is incorrect:
https://github.com/pods-framework/pods/blame/2.x/classes/fields/color.php#L114

It should check if the `$color` var is not empty.

## Steps to Reproduce (for bugs)
1. Create color field and make sure `required` is not checked.
2. Go to that CPT / Tax or setting page
3. Make sure the color field is set to empty/nothing
4. You'll see the error

## Related Issues and/or PRs
PR: #4920

PR #4920, opened 2018-05-07
Check if $color has a value before checking this value.

JoryHogeveen:
Fixes: #4919

### DIFF
--- a/classes/fields/color.php
+++ b/classes/fields/color.php
@@ -111,7 +111,7 @@ class PodsField_Color extends PodsField {
 					// @todo Ask for a specific format in error message
 					$errors[] = __( 'Invalid value provided for this field.', 'pods' );
 				}
-			} elseif ( ! in_array( strlen( $color ), array( 3, 6 ), true ) ) {
+			} elseif ( $color && ! in_array( strlen( $color ), array( 3, 6 ), true ) ) {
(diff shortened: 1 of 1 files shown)


classes/fields/color.php
```diff
@@ -111,7 +111,7 @@ public function validate( $value, $name = null, $options = null, $fields = null,
 					// @todo Ask for a specific format in error message
 					$errors[] = __( 'Invalid value provided for this field.', 'pods' );
 				}
-			} elseif ( ! in_array( strlen( $color ), array( 3, 6 ), true ) ) {
+			} elseif ( $color && ! in_array( strlen( $color ), array( 3, 6 ), true ) ) {
```
pglewis: Not sure it'll make any difference in this context, but lets change it to `! empty( $color )` to match the logic before the breakage

JoryHogeveen: Won't matter since it's already declared but fine with me. Also enhances readability so that's a plus :)

JoryHogeveen:
I think it would be good to add some unit tests to this PR. I'll get to that a.s.a.p. tomorrow unless anyone else wants to jump in.

pglewis:
I'm all for adding tests but I'd rather create a separate issue and PR for them.

sc0ttkclark:
@JoryHogeveen :taco: just ran into this in my own testing (without realizing i was running 2.7.3), you are a saint!

PR #4923, opened 2018-05-10
Fix CodeMirror fields in Taxonomy and User edit forms

### LINKED ISSUE #4913: Code Syntax Fields Overflowing primary container on Taxonomy/Category post 2.7.2
After update Pods to version 2.7.2 and the last 2.7.2.1, there are problems in admin of category pages.

All fields are stretched at a width larger than the width of the page. The problem persists even if I switch on one of the default WordPress theme.

No other plugin has been updated in recent days. Just Pods.

When I deactivate Pods plugin, everything is fine.

Can you fix it, please?

Thank you, David

pglewis:
Fixes #4913


pglewis:
The sole purpose of `pods-compat-container` is to wrap any widgets we may need to style but don't want to cause conflicts with other plugins that my use them as well.  If we don't like the name, speak now before it proliferates more.

pglewis:
Branch name typo, #4927 is the correct PR

PR #4959, opened 2018-05-25
'00:00:00' is midnight and a perfectly valid time value

### LINKED ISSUE #4937: Hour field blank if set to 12.00am
The hour field remains blank on save when I select 12.00am as a time.

This is a re-up of an earlier report of mine here - [Issue 3499](https://github.com/pods-framework/pods/issues/3488)

Also - and perhaps this should be a separate issue - it isn't possible to copy and paste into the hour fields which would be handy for multiple duplicate fields. 

blank-time-field

pglewis:
Change `empty_value` to the empty string and override `format_value_display` in the time field to allow midnight as a valid time and allow a proper empty value to be set (where allowed).

Fixes #4937, #3488

This is hopefully just a final patch to fix a couple issues with the time field before converting date/time/datetime to DFV.

### DIFF
--- a/classes/fields/time.php
+++ b/classes/fields/time.php
@@ -40,7 +40,7 @@ class PodsField_Time extends PodsField_DateTime {
 	 * @var string
 	 * @since 2.7
 	 */
-	public static $empty_value = '00:00:00';
+	public static $empty_value = '';
 
 	/**
 	 * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


pglewis:
I have manually tested this some: 

- Accepts midnight as a valid time for time field
- Accepts an empty value or 0 to leave it blank, when allowed
- Sets the field to the current time if allow blank is unchecked
- Various other randomly tested values appear to be working

jimtrue:
This might be a separate issue entirely (but based on this), but midnight is showing as 'empty' in the `[if timefield][else][/if]` template tag tests. @pglewis I can generate this as a separate issue 'cause I'm not sure how easy this would be to fix considering the is_empty tests were added in 2.7 to finally fix that issue in the first place.

PR #4971, opened 2018-06-01
Fix false check for file uploads directory check when exporting

### LINKED ISSUE #4970: Export process failed
https://github.com/pods-framework/pods/blob/ae9f431aec0419a940c7a8701efb7dc26d198c6d/classes/PodsMigrate.php#L798

The export process fails cause wrong check of $uploads['error'] that in normal way is false.

The line fix could be: if ( ! $uploads || false !== $uploads['error'] )

sc0ttkclark:
Fixes #4970


pglewis:
Changelog: Fixed: Resolve file uploads directory check logic for file exports #4970 (@sc0ttkclark)

PR #4975, opened 2018-06-02
PHP Codesniffer audit script

### LINKED ISSUE #4974: PHP Codesniffer audit script improvements
Currently no matter what you do it won't queue any files.  The problem seems to be twofold as both these seem to break any processing:

1. The ignore list given to the command line in `phpcs-audit.sh`
2. The `exclude-pattern` section of `phpcs.xml.dist`

I would guess the first one could just be yanked out since I assume the exclude-pattern in the xml file handles that. 

Removing those two things allows me to run the script against a single file again.

pglewis:
Fixes #4974

### DIFF
--- a/bin/phpcs-audit.sh
+++ b/bin/phpcs-audit.sh
@@ -24,5 +24,5 @@ echo Running PHPCS reports
 for type in ${types[@]}
 do
   echo Running PHPCS report: ${type}
-  ./vendor/bin/phpcs -v --report="${type}" --report-file="report-${type}${file_affix}.txt" ${file_path}
+  ./vendor/bin/phpcs -v --report="${type}" --report-file="phpcs-report-${type}${file_affix}.txt" ${file_path}
 done;
--- a/bin/phpcs-audit.sh
+++ b/bin/phpcs-audit.sh
@@ -17,8 +17,6 @@ file_path=${1:-$default_path}
 default_file_affix=""
 file_affix=${2:-$default_file_affix}
 
-ignore="\.idea,\.saas-cache,node_modules,vendor,bin,tests"
-
 declare -a types=("full" "source")
 
 echo Running PHPCS reports
@@ -26,5 +24,5 @@ echo Running PHPCS reports
 for type in ${types[@]}
 do
   echo Running PHPCS report: ${type}
-  ./vendor/bin/phpcs -v -s -p -d memory_limit="256M" --extensions="php" --report="${type}" --ignore="${ignore}" --report-file="report-${type}${file_affix}.txt" ${file_path}
+  ./vendor/bin/phpcs -v -s -p -d memory_limit="256M" --extensions="php" --report="${type}" --report-file="report-${type}${file_affix}.txt" ${file_path}
 done;
--- a/bin/phpcs-audit.sh
+++ b/bin/phpcs-audit.sh
@@ -24,5 +24,5 @@ echo Running PHPCS reports
 for type in ${types[@]}
 do
   echo Running PHPCS report: ${type}
-  ./vendor/bin/phpcs -v --report="${type}" --report-file="phpcs-report-${type}${file_affix}.txt" ${file_path}
+  ./vendor/bin/phpcs --report="${type}" --report-file="phpcs-report-${type}${file_affix}.txt" ${file_path}
 done;
--- a/bin/phpcs-autofix.sh
+++ b/bin/phpcs-autofix.sh
@@ -12,4 +12,4 @@ default_path="."
 file_path=${1:-$default_path}
 
 echo Running PHPCS autofix
-./vendor/bin/phpcbf -v ${file_path}
+./vendor/bin/phpcbf ${file_path}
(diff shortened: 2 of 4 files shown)


bin/phpcs-audit.sh
```diff
@@ -17,14 +17,12 @@ file_path=${1:-$default_path}
...
-
 declare -a types=("full" "source")
 
 echo Running PHPCS reports
 
 for type in ${types[@]}
 do
   echo Running PHPCS report: ${type}
-  ./vendor/bin/phpcs -v -s -p -d memory_limit="256M" --extensions="php" --report="${type}" --ignore="${ignore}" --report-file="report-${type}${file_affix}.txt" ${file_path}
+  ./vendor/bin/phpcs -v -s -p -d memory_limit="256M" --extensions="php" --report="${type}" --report-file="report-${type}${file_affix}.txt" ${file_path}
```
GaryJones: I'd advise that as many of these args as possible, get moved into the `phpcs.xml.dist`. See https://github.com/GaryJones/plugin-boilerplate/blob/bfe875179f641797576f6cbcd6b45a7e3931c186/phpcs.xml.dist#L11-L17 as an example. It would usually just be possible to call `./vendor/bin/phpcs`, whether manually, from Composer script or a bash script, and have it run consistently.

GaryJones:
The other benefit of having these args in `phpcs.xml.dist`, is that one won't have to remember to update `phpcs-autofix.sh` when the args change in `phpcs-audit.sh` ;-)

pglewis:
Hat tip, @GaryJones.

pglewis:
Fairly happy with where it stands now, the only thing left to decide for me is whether to use `-v` or `-p` for the terminal output as those args appear to be mutually exclusive.  My vote is for the summary of `-p`.

GaryJones:
> My vote is for the summary of -p

I would agree, but wasn't sure if this decision had previously been made, so didn't change it in my patch to this PR.

Background info the uninitiated reader:

Typically `-p` is used to show progress - it gives a series of dots with an indication of how many files are done out of a total. It removes the developer thinking "Is this working?", though the large file sizes mean it can appear to pause.

`-v` is a form of debugging - while it shows the number of tokens in a file, and the file names, for a typical run, it becomes a distraction, since there's too much redundant info to be useful.

pglewis:
Agreed and done.  

In cases where `-v` is desired for debugging it's probably better to just run phpcs directly.

pglewis:
I guess we might want to take `-v` out of the autofix file as well.

bin/phpcs-autofix.sh
```diff
@@ -11,7 +11,5 @@
 default_path="."
 file_path=${1:-$default_path}
 
-ignore="\.idea,\.saas-cache,node_modules,vendor,bin,tests"
-
 echo Running PHPCS autofix
-./vendor/bin/phpcbf -v -d memory_limit="256M" --extensions="php" --ignore="${ignore}" ${file_path}
+./vendor/bin/phpcbf -v ${file_path}
```
GaryJones: With the `-v` gone, this bash file becomes redundant, in so much as it provides no additional args to `phpcs`, than just calling `phpcs` directly. The only extra thing would be that `echo`. If that's not deemed critical, could the whole be removed (I'm not sure where it is used)?

phpcs.xml.dist
```diff
@@ -2,12 +2,19 @@
...
-	<exclude-pattern>.saas-cache/*</exclude-pattern>
-	<exclude-pattern>node_modules/*</exclude-pattern>
-	<exclude-pattern>tests/*</exclude-pattern>
-	<exclude-pattern>vendor/*</exclude-pattern>
-	<exclude-pattern>components/Markdown.php</exclude-pattern>
+	<file>.</file>
+	<exclude-pattern>*/.idea/*</exclude-pattern>
+	<exclude-pattern>*/.sass-cache/*</exclude-pattern>
+	<exclude-pattern>*/node_modules/*</exclude-pattern>
+	<exclude-pattern>*/tests/*</exclude-pattern>
```
GaryJones: Shouldn't the files in the `/tests/` directory follow the same code standards as well?

Yes, these aren't provided to the plugin users, but they are (or should be) as much a part of the project as the source files.

pglewis:
@GaryJones 

Autofix: I would agree but I've never used the autofix script and probably won't too soon.  That's one to check with @sc0ttkclark in case he still has a need for it.  

Tests: In the longer term I agree but with considering all the things we could do to improve Pods without a dev _team_, PHPCS scans on the PHP tests are very very low on my list.

bin/phpcs-autofix.sh
sc0ttkclark: I like having the script so I (and others) don't have to always type the path to phpcbf etc

phpcs.xml.dist
sc0ttkclark: Yeah we can probably remove the tests exclusion

pglewis:
Alright, this one is merging as soon as tests finish.

bin/phpcs-autofix.sh
GaryJones: Fair enough, but adding `./vendor/bin` to your `$PATH` (before ~/.composer/vendor/bin)` and this becomes un-necessary.

PR #4978, opened 2018-06-03
Refactor logic for determining whether or not to activate AJAX mode for autocomplete based field views

### LINKED ISSUE #4504: Autocomplete/List view broken with "Other WP Objects"
Testing stuff on #4484 I discovered that Autocomplete and List view did not pull up anything for most (all?) of the "Other WP Objects".

One of the fields was initially a multi select and once the setting was switched to List view it would not actually remove the entry when removing it and saving.

pglewis:
Fixes #4504

### DIFF
--- a/classes/fields/pick.php
+++ b/classes/fields/pick.php
@@ -780,12 +780,12 @@ class PodsField_Pick extends PodsField {
 
 		$ajax = false;
 
-		if ( ( 'custom-simple' !== pods_v( $args->type . '_object', $options ) || empty( $custom ) ) && '' !== pods_v( $args->type . '_object', $options, '', true ) ) {
+		if ( $this->can_ajax( $options ) ) {
 			$ajax = true;
-		}
 
-		if ( ! empty( self::$field_data ) && self::$field_data['id'] === $options['id'] ) {
-			$ajax = (boolean) self::$field_data['autocomplete'];
+			if ( ! empty( self::$field_data ) && self::$field_data['id'] === $options['id'] ) {
+				$ajax = (boolean) self::$field_data['autocomplete'];
+			}
 		}
 
 		$ajax = apply_filters( 'pods_form_ui_field_pick_ajax', $ajax, $args->name, $args->value, $options, $args->pod, $args->id );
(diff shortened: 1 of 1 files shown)


classes/fields/pick.php
```diff
@@ -2389,6 +2391,27 @@ public function is_autocomplete( $options ) {
...
+	 *
+	 * @param array $options Field options.
+	 *
+	 * @return bool
+	 *
+	 * @since 2.7.4
+	 */
+	public function can_ajax( $options ) {
+
+		$is_simple_tableless = in_array( $options['pick_object'], PodsForm::simple_tableless_objects(), true );
```
sc0ttkclark: Other tableless fields might extend the Pick field type class, we will want to use `pods_v( $options['type'] . '_object', $options )` instead of `$options['pick_object']`

pglewis:
Holding for AJAX mode testing, but looking good so far

classes/fields/pick.php
```diff
@@ -2389,6 +2391,28 @@ public function is_autocomplete( $options ) {
...
+	/**
+	 * Check if a field supports AJAX mode
+	 *
+	 * @param string $type
+	 * @param array  $options Field options.
+	 *
+	 * @return bool
+	 * @since 2.7.4
+	 */
+	public function can_ajax( $type, $options ) {
```
GaryJones: This method can be simplified further down to:

GaryJones: But that still leaves half the method being about determining if this is simple tableless (whatever that might mean), so let's extract that logic into its own method as well:

```php
public function is_simple_tableless( $type, $options ) {
	$field_object = pods_v( $type . '_object', $options );
	
	return in_array( $field_object, PodsForm::simple_tableless_objects(), true );	
}

public function can_ajax( $type, $options ) {
	return $this->is_autocomplete( $options ) && ! $this->is_simple_tableless( $type, $options );
}
```

The first method could be made `protected` if it's not needed to be public.

pglewis: This will need a fresh issue since I already merged this PR.  

I actually want to set both those methods as well `is_autocomplete()` (only added in 2.7) to private, crossed my mind but I forgot about it.  I'm not sure those methods survive a proper full refactor and anything public becomes a back-compat liability until the end of time.

GaryJones: > This will need a fresh issue since I already merged this PR.

Done in #4993 + #4994.

PR #4996, opened 2018-06-05
PHPCS: Contents of `@since` and `@deprecated` tags

### LINKED ISSUE #4995: PHPCS: Ensure since tags use three digits
## Issue Overview
Some `@since` tags are only set to `x.y`, but the [WPCS says](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#since-section-changelogs) it should be `x.y.z`, even when `z` or `y` are `0`.

## Expected Behavior
All `@since` and `@deprecated` tags should use three digits. Apart from creating consistency, it also means that it's easier to find instance of functions added / deprecated since exactly `2.7.0`, rather than just `2.7.*`.

## Current Behavior
A regex search of `@since\W+\d\..$` shows over 100 instances that would need fixing.

GaryJones:
Documentation code standards for the _contents_ (not the position in DocBlock) of `@since` and `@deprecated` tags.

Fixes #4995.

### DIFF
--- a/deprecated/classes/Pods.php
+++ b/deprecated/classes/Pods.php
@@ -22,7 +22,7 @@ class Pods_Deprecated {
 	 * @param object $obj The Pods object
 	 *
 	 * @license http://www.gnu.org/licenses/gpl-2.0.html
-	 * @since   2.0
+	 * @since 2.0.0
 	 */
 	public function __construct( $obj ) {
 
@@ -500,7 +500,7 @@ class Pods_Deprecated {
 	/**
 	 * Search and filter records
 	 *
-	 * @since      1.x
+	 * @since 1.x.x
 	 * @deprecated deprecated since version 2.0
 	 *
 	 * @param null $orderby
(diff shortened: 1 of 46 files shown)


deprecated/classes/Pods.php
```diff
@@ -500,8 +500,8 @@ public function get_pod_id() {
 	/**
 	 * Search and filter records
 	 *
-	 * @since      1.x
-	 * @deprecated deprecated since version 2.0
+	 * @since 1.x.x
```
pglewis: I assume the '1.x.x' entries are details lost in the depths of time.  Are these Kosher to leave like this?

GaryJones: I found and added in a couple of missing values for `@deprecated`, but wanted to get this PR in for 2.7.4.

GaryJones:
One other thing I noticed - there are some `@since` tags with `0.1.0` or `0.2.0`. Does that level of pre-1.0 release need to be kept? Or could they be bumped to `1.0.0`?

Issue #4229, opened 2017-05-18
Remove dead-code php field files that have been replaced by DFV


pglewis:
I think at least select2, plupload, media... probably others.  

farbtastic is yanked in #4208 not here

Issue #4067, opened 2017-03-18
Convert Avatar field to DFV


pglewis:
Main 2.7 issue: #3364

It's basically a file upload field, which has been converted.  This will remove all dependencies on the old `attachment.php`, `media.php`, and `plupload.php` files under `ui/fields`.

PR #4997, opened 2018-06-05
JS: jQuery.fn.size() is deprecated

### LINKED ISSUE #3898: jQuery.fn.size() is deprecated
https://github.com/pods-framework/pods/blob/30dff/ui/fields/media.php#L301
https://github.com/pods-framework/pods/blob/30dff/ui/js/jquery.pods.attach.js#L84

GaryJones:
Use `.length` property instead.

Fixes #3898

Only one instance of `size()` existed, compared to the two mentioned in #3898.


pglewis:
> Only one instance of size() existed, compared to the two mentioned in #3898.

I actually had to dig up why, I completely forgot: 
#4067
#4229

PR #4998, opened 2018-06-05
Update issue templates

jimtrue:
Updated to the new formats provided by GitHub


jimtrue:
@pglewis and @sc0ttkclark I pulled in details from our existing ones. These are specifically to differentiate Bug Reports from Feature requests.

GaryJones:
@jimtrue Was there an announcement from GH about a change in template format?

Issue #4992, opened 2018-06-05
PHPCS: Use interpolation for dynamic hook names


GaryJones:
## Issue Overview
The [coding standards says](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#interpolation-for-naming-dynamic-hooks):

> Dynamic hooks should be named using interpolation rather than concatenation for readability and discoverability purposes.

Among other reasons, it allows WP-Parser and other documentation extraction tools to more easily read the whole double-quoted first arg string, instead of the tokens in the first arg up to the first comma.

## Expected Behavior
A dynamic hook name like:

```php
apply_filters( 'pods_pods_field_' . $this->fields[ $params->name ]['type'], ... );
```

_could_ be updated to be:

```php
apply_filters( "pods_pods_field_{$this->fields[ $params->name ]['type']}", ... );
```

and _should_ be updated to be:

```php
$field_type = $this->fields[ $params->name ]['type'];
apply_filters( "pods_pods_field_{$field_type}", ... );
```

## Current Behavior
A quick regex search with `~apply_filters\([^,]*\..*,~` and `~do_action\([^,]*\..*,~` suggests there are at least 98 instances of concatenation (plus any `*_ref_array()` and `*_deprecated()` variations).

There are some filter and action hooks which are already named with interpolation, so this isn't a new concept the Pods codebase.

## Possible Solution
Update filter and action hook names where necessary. This doesn't affect the PHP runtime behaviour.

Note that it's [not yet possible](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/751) to use WPCS to detect the instances that would need changing.

pglewis:
PR: #5011

PR #5011, opened 2018-06-08
Feature/4992 interpolated hook names

GaryJones:
## Description
PHPCS: Convert filter and action hook names to be interpolated

Where the value was stored in something other than a scalar variable, a scalar variable has been added nearby with the value assigned. This keeps the name of the hooks cleaner.

See #4992.

I've not added commits to adjust existing interpolated hook names to scalar variables. These can be added if desired.

## Types of changes
Coding standards / improved generated documentation.

## ChangeLog
Improvement: Use interpolated hook names, per WPCS, #4992 (@GaryJones)

## Checklist:
- [ ] My code is tested.
- [X] My code follows the WordPress code style.

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -214,7 +214,9 @@ class PodsAdmin {
 						continue;
 					}
 
-					$pod = apply_filters( 'pods_advanced_content_type_pod_data_' . $pod['name'], $pod, $pod['name'] );
+					$pod_name = $pod['name'];
+
+					$pod = apply_filters( "pods_advanced_content_type_pod_data_{$pod_name}", $pod, $pod['name'] );
 					$pod = apply_filters( 'pods_advanced_content_type_pod_data', $pod, $pod['name'] );
 
 					if ( 1 === (int) pods_v( 'show_in_menu', $pod['options'], 0 ) ) {
@@ -789,7 +791,9 @@ class PodsAdmin {
 				'actions_disabled' => $actions_disabled,
 			);
 
-			$ui = apply_filters( 'pods_admin_ui_' . $pod->pod, apply_filters( 'pods_admin_ui', $ui, $pod->pod, $pod ), $pod->pod, $pod );
+			$pod_pod_name = $pod->pod;
+
+			$ui = apply_filters( "pods_admin_ui_{$pod_pod_name}", apply_filters( 'pods_admin_ui', $ui, $pod->pod, $pod ), $pod->pod, $pod );
 
 			// Force disabled actions, do not pass go, do not collect $two_hundred
 			$ui['actions_disabled'] = $actions_disabled;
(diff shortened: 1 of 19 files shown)


pglewis:
How on Earth could that have passed tests considering 9c7606e?

classes/PodsAdmin.php
```diff
@@ -802,7 +802,7 @@ public function admin_content_settings() {
 		} else {
 			$pod_pod_name = $pod->pod;
 			do_action( 'pods_admin_ui_custom', $pod );
-			do_action( "pods_admin_ui_custom_{$pod_pod_name}, $pod );
+			do_action( "pods_admin_ui_custom_{$pod_pod_name}", $pod );
```
GaryJones: Dammit. Thanks.

GaryJones:
Maybe because the tests don't cover checking that that particular filter is called?

pglewis:
I got a fatal run-time just attempting to go to the dashboard in the browser though.

pglewis:
I'm good here but will hold on @sc0ttkclark in case he wants to add one more set of eyes.

ui/admin/form.php
```diff
@@ -15,7 +15,7 @@
 $groups = PodsInit::$meta->groups_get( $pod->pod_data['type'], $pod->pod_data['name'], $fields );
 
 $pod_options = $pod->pod_data['options'];
-$pod_options = apply_filters( 'pods_advanced_content_type_pod_data_' . $pod->pod_data['name'], $pod_options, $pod->pod_data['name'] );
+$pod_options = apply_filters( "pods_advanced_content_type_pod_data_{$pod->pod_data['name']}", $pod_options, $pod->pod_data['name'] );
```
sc0ttkclark: Doesn't fit the same way the others were changed, can you set as var and then put in filter name?

components/Helpers.php
```diff
@@ -513,7 +513,7 @@ public static function helper( $params, $obj = null ) {
 		$out = ob_get_clean();
 
 		$out = apply_filters( 'pods_helpers_post_helper', $out, $params, $helper );
-		$out = apply_filters( 'pods_helpers_post_helper_' . $helper['slug'], $out, $params, $helper );
+		$out = apply_filters( "pods_helpers_post_helper_{$helper['slug']}", $out, $params, $helper );
```
sc0ttkclark: Doesn't fit the same way the others were changed, can you set as var and then put in filter name?

ui/admin/form.php
GaryJones: Addressed in a8f4ee864afea764b8ba67c836dc1c675c1f5582.

components/Helpers.php
GaryJones: Addressed in a8f4ee864afea764b8ba67c836dc1c675c1f5582.

GaryJones:
Fixes for the two small issues have now been added.

Issue #5014, opened 2018-06-08
Broken: Add New doesn't 'SELECT' record or refresh list view in 2.7.5


jimtrue:
**Describe the bug**
Reported from: https://pods.io/forums/topic/2-7-5-not-linking-relationship-field/#post-198847

List View, Add New creates supplementary record, but the record created isn't selected and doesn't refresh List View so record isn't 'added' to Relationship. Refreshing the post will allow selection, but obviously this isn't ideal.

**To Reproduce**
1. Go to any existing Relationship List View in 2.7.5
2. Add New and Save Child Record
3. Modal Closes, but record is not selected and is not listed in List View
4. Refresh screen, record can be selected.

**Expected behavior**
This should work the way it used to ;)

**Screenshots**

pglewis:
Appears to have been introduced in 2.7.4, off to find out exactly where.

pglewis:
Preliminary, but I believe the winner is: 6f15f04

pglewis:
PR: #5017

PR #5020, opened 2018-06-08
Update inline docs for pick field selected logic

sc0ttkclark:
Docs for logic fixes in #5014 and #5017

Changelog: Update inline docs for pick field selected logic


classes/fields/pick.php
```diff
@@ -1298,25 +1298,37 @@ public function build_dfv_field_item_data_recurse_item( $item_id, $item_title, $
...
 			}
+		}
 
-			if ( ! $selected ) {
-				// Cast values in array as string.
-				$args->value = array_map( 'strval', $args->value );
+		// If we do not have a key match, the normal values may still match.
+		if ( ! $selected ) {
+			// Cast values in array as string.
+			$values = array_map( 'strval', $values );
```
pglewis: I'm considering this just another step as part of properly initializing $values.... any reason not to just do this unconditionally above, after initializing and casting $values?  ~L1308?

GaryJones:
I'd consider most of these comments to be redundant.

The additional comments say _what_ the code is doing, but I can see that from the code.

The comments don't really add anything specific about _why_ the code is doing what it is doing. That where the value of comments are.

If anything could be extracted into named private/protected methods, then that would self-document the _intent_ of the code, and comments would still not be generally needed.

References on comments being a _code smell_:

- https://refactoring.guru/smells/comments
- https://medium.com/@fagnerbrack/code-comment-is-a-smell-4e8d78b0415b
- https://blogs.agilefaqs.com/2013/03/07/inline-comments-in-the-code-is-a-smell-but-document-the-why/
- https://simpleprogrammer.com/why-comments-are-stupid-a-real-example/
- and many more.

pglewis:
A lot of good points, but this is something I asked Scott to do for my own edification.  I plan to boil that down somewhat.  

> I'd consider most of these comments to be redundant.

There are areas where I personally disagree with some of the "comments as code smell" opinions.  A good portion of one's time spent in a legacy codebase is finding _where_ as opposed to _what_.  I think obvious comments at regular intervals can be a very useful thing for how we work because you can scan those obvious comments without digesting the code logic to locate the _where_ more quickly.

classes/fields/pick.php
sc0ttkclark: Yeah this sounds good above.

classes/fields/pick.php
```diff
@@ -1305,6 +1305,9 @@ public function build_dfv_field_item_data_recurse_item( $item_id, $item_title, $
 		if ( ! empty( $args->value ) ) {
 			// The value may be a single non-array value.
 			$values = (array) $args->value;
+
+			// Cast values in array as string.
+			$values = array_map( 'strval', $values );
```
pglewis: This is still casting conditionally for no good reason that I can see.  It shouldn't functionally matter right now as the only other case is an empty array as the code stands, but code changes.  We insist on a strict string comparison on keys later in the code so we should always, unconditionally cast the keys IMO.

pglewis: The danger IME is someone comes along later and modifies the enclosing `if` so it's now `if (! empty( $args->value )  && now_only_executes_some_of_the_time )` and we start failing to cast again.

sc0ttkclark: Fixed

JoryHogeveen:
@pglewis @sc0ttkclark What is the status for this one?

classes/fields/pick.php
```diff
@@ -1304,28 +1304,41 @@ public function build_dfv_field_item_data_recurse_item( $item_id, $item_title, $
 			$edit_link = add_query_arg( array( 'pods_modal' => '1' ), $edit_link );
 		}
 
-		// Determine if this is a selected item
+		// Determine if this is a selected item.
+		// Issue history for setting selected: #4753, #4892, #5014
```
tr1b0t: :no_entry_sign: Inline comments must end in full-stops, exclamation marks, or question marks

<sub>hash: a47f55de534fcce5fe0d9ad2ce17205b</sub>

sc0ttkclark:
```suggestion
		// Issue history for setting selected: #4753, #4892, #5014.
```

PR #5027, opened 2018-06-10
PHPCS: Fix WordPress.WhiteSpace.PrecisionAlignment

### LINKED ISSUE #5026: PHPCS: WordPress.WhiteSpace.PrecisionAlignment
`WordPress.WhiteSpace.PrecisionAlignment` is a non-autofixable code standards check that looks for rogue spaces before non-tab characters.

Addressing these means that other more-important issues don't get lost in the relatively non-controversially-fixable noise.

GaryJones:
## Description
Fix violations from the `WordPress.WhiteSpace.PrecisionAlignment` sniff.

Fixes #5026

## Screenshots (jpeg or gifs if applicable):

Before: 

screenshot 2018-06-10 19 36 35

After:

screenshot 2018-06-10 19 35 41

## Types of changes
Code standards.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -3638,11 +3638,11 @@ class PodsAPI {
 			$options = pods_var( 'options', $field_data, array() );
 
 			// WPML AJAX compatibility
-			if ( is_admin() &&
-				( isset( $_POST['action'] ) && 'wpml_save_job_ajax' === $_POST['action'] ) ||
-				( isset( $_GET['page'] ) && false !== strpos( $_GET['page'], '/menu/languages.php' ) &&
-				isset( $_POST['icl_ajx_action'] ) && isset( $_POST['_icl_nonce'] ) &&
-				wp_verify_nonce( $_POST['_icl_nonce'], $_POST['icl_ajx_action'] . '_nonce' ) )
+			if ( is_admin()
+				&& ( isset( $_POST['action'] ) && 'wpml_save_job_ajax' === $_POST['action'] )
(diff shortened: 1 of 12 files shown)


classes/PodsAPI.php
```diff
@@ -3640,9 +3640,9 @@ public function save_pod_item( $params ) {
 			// WPML AJAX compatibility
 			if ( is_admin() &&
 				( isset( $_POST['action'] ) && 'wpml_save_job_ajax' === $_POST['action'] ) ||
-				( isset( $_GET['page'] ) && false !== strpos( $_GET['page'], '/menu/languages.php' )
-				 && isset( $_POST['icl_ajx_action'] ) && isset( $_POST['_icl_nonce'] )
-				 && wp_verify_nonce( $_POST['_icl_nonce'], $_POST['icl_ajx_action'] . '_nonce' ) )
```
sc0ttkclark: I liked it this way, is the coding standards requiring conditional to be on the left of new line or right side on previous line?

pglewis: Ditto.  The right edge is ragged and may require horizontal scrolling just to see the logic,  To quote the PEAR manual `Keeping the operators at the beginning of the line has two advantages: It is trivial to comment out a particular line during development while keeping syntactically correct code (except of course the first line). Further is the logic kept at the front where it's not forgotten. Scanning such conditions is very easy since they are aligned below each other.`.

GaryJones: My apologies - the CS issue here was the single space on front of `&&` for both lines.

quasel: if you keep the operator at the start which feels a bit more readable the && from line 3641 should move to 42 consistency wise ^^

PR #5041, opened 2018-06-15
Fixed the CLI export/export-item commands.

0xLBF:
When the export_data method of a pod without find parameters is called, and no 'params' are passed explicitly to it, an empty array is returned. Assuming this is the intended behavior, I added find parameters to the pod created by the export CLI command.

The export_data_to_file method did not set the 'columns' field of the data, which is required for any of the exports to work, so I added code to derive the columns from the supplied data.

The export_data_to_file method always forwarded the data enclosed in an array. This caused the JSON export-item to enclose the item in an array, the XML export to report only 1 item, and the CSV export not to work at all. Forwarding the data leads to a more expected result for the JSON export-item, and required handling the 'single' parameter for the CSV/XML export-item to continue to work.

Tested the export/export-item CLI commands for CSV/XML/JSON exports and they all seemed to work fine now.


classes/PodsMigrate.php
```diff
@@ -570,7 +570,12 @@ public function build_sv( $data = null, $delimiter = null ) {
 
 		$head = substr( $head, 0, - 1 );
 
-		foreach ( $this->data['items'] as $item ) {
+		$items = $this->data['items'];
+		if ( ! empty( $this->data['single'] ) ) {
```
GaryJones: There should be a check that the array has the `single` key.

classes/PodsMigrate.php
```diff
@@ -636,10 +641,15 @@ public function build_xml( $data = null ) {
 			return false;
 		}
 
-		$head  = '<' . '?' . 'xml version="1.0" encoding="utf-8" ' . '?' . '>' . "\r\n<items count=\"" . count( $this->data['items'] ) . "\">\r\n";
+		$items = $this->data['items'];
```
GaryJones: This is the second time this logic appears - it should be in its own private method:

classes/PodsMigrate.php
```diff
@@ -1258,10 +1268,22 @@ public static function export_data_to_file( $file, $data, $single = false ) {
...
-			'items'  => array( $data ),
+			'items'  => $data,
 			'single' => $single,
 		);
 
+		// try to guess the column labels based on the supplied data
+		$first_item = null;
+		if ( $single ) {
+			$first_item = $data;
+		} else if ( is_array( $data ) ) {
```
GaryJones: Should be `elseif`.

0xLBF:
@GaryJones Thanks, corrected the elseif and refactored the logic into a method. A more general method can also be added later if needed but for now I just added one that checks this->data directly. this->data is always set with set_data which should ensure that the 'items' field is set, so if something has gone so horribly wrong that 'items' is still missing, it's ok to just let PHP throw an error ? My understanding is the empty function should still just return false even if 'single' is not set, so it should be fine ?

classes/PodsMigrate.php
```diff
@@ -111,6 +111,17 @@ public function set_data( $data ) {
 
 		$this->data = array_merge( $defaults, (array) $data );
 	}
+	
+	/**
+	 * @return array	Array of data items
+	 */
+	private function get_items ( ) {
```
sc0ttkclark:
```suggestion
	private function get_items() {
```

classes/PodsMigrate.php
```diff
@@ -111,6 +111,17 @@ public function set_data( $data ) {
 
 		$this->data = array_merge( $defaults, (array) $data );
 	}
+	
+	/**
+	 * @return array	Array of data items
```
sc0ttkclark:
```suggestion
	 * Get items.
	 *
	 * @return array List of data items.
	 *
	 * @since 2.7.13
```

classes/PodsMigrate.php
```diff
@@ -1258,10 +1271,22 @@ public static function export_data_to_file( $file, $data, $single = false ) {
 		}
 
 		$migrate_data = array(
-			'items'  => array( $data ),
+			'items'  => $data,
 			'single' => $single,
 		);
 
+		// try to guess the column labels based on the supplied data
```
sc0ttkclark:
```suggestion
		// Try to guess the column labels based on the supplied data.
```

classes/PodsMigrate.php
```diff
@@ -1258,10 +1271,22 @@ public static function export_data_to_file( $file, $data, $single = false ) {
...
+		$first_item = null;
+		if ( $single ) {
```
sc0ttkclark:
```suggestion

		if ( $single ) {
```

classes/PodsMigrate.php
```diff
@@ -1258,10 +1271,22 @@ public static function export_data_to_file( $file, $data, $single = false ) {
...
+			$first_item = $data;
+		} elseif ( is_array( $data ) ) {
+			$first_item = reset( $data );
+		}
+		if ( is_array( $first_item ) ) {
```
sc0ttkclark:
```suggestion

		if ( is_array( $first_item ) ) {
```

classes/PodsMigrate.php
```diff
@@ -1258,10 +1271,22 @@ public static function export_data_to_file( $file, $data, $single = false ) {
...
+			$fields = array_keys( $first_item );
```
sc0ttkclark:
```suggestion
			$fields                  = array_keys( $first_item );
```

sc0ttkclark:
@0xLBF Can you give us access to your Pods repo fork or accept the changes we suggested in the PR?

classes/PodsMigrate.php
```diff
@@ -570,7 +585,7 @@ public function build_sv( $data = null, $delimiter = null ) {
 
 		$head = substr( $head, 0, - 1 );
 
-		foreach ( $this->data['items'] as $item ) {
+		foreach ( $this->get_items() as $item ) {
```
sc0ttkclark:
```suggestion
		$items = $this->get_items();

		foreach ( $items as $item ) {
```

PR #5042, opened 2018-06-19
Add _total and _total_found special fields

### LINKED ISSUE #5039: Enhancement: total and total_found magic tag
**Pod Count Tag**
It would be very useful to have a tag to show the total items on a given pod. For example, in a Pod named Book, there could be a {@books-total} tag that could be used to display the total amount of books.

## Tasklist

- [x] Define special magic tag name for total and total found
- [x] Add logic for `total` (or whatever we name it) that returns `$this->total()`
- [x] Add logic for `total_found` (or whatever we name it) that returns `$this->total_found()`

sc0ttkclark:
Fixes #5039

## Example

```
Currently showing {@_total} rows out of {@_total_found} rows found.
```

- [x] `_total` Pods::total()
- [x] `_total_found` Pods::total_found()
- [ ] Get `[once]` handling magic tags
- [ ] Get `[before]` handling magic tags
- [ ] Get `[after]` handling magic tags

### DIFF
--- a/tests/phpunit/includes/tests-traversal.php
+++ b/tests/phpunit/includes/tests-traversal.php
@@ -668,6 +668,11 @@ class Test_Traversal extends Pods_UnitTestCase {
 
 		$this->assertEquals( (string) $data['id'], (string) $p->id(), sprintf( 'Item ID not as expected (%s) [%s]', $data['field_id'], $variant_id ) );
 
+		$this->assertEquals( 1, (int) $p->field( '_total' ), sprintf( 'Total value not as expected (%s) [%s]', $p->total(), $variant_id ) );
+		$this->assertEquals( '', $p->field( '_total_s' ), sprintf( 'Total "s" value not as expected (%s) [%s]', $p->total(), $variant_id ) );
+		$this->assertEquals( 1, (int) $p->field( '_total_found' ), sprintf( 'Total found value not as expected (%s) [%s]', $p->total_found(), $variant_id ) );
+		$this->assertEquals( '', $p->field( '_total_found_s' ), sprintf( 'Total found "s" value not as expected (%s) [%s]', $p->total_found(), $variant_id ) );
+
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


sc0ttkclark:
@GaryJones The way WP uses `_n( string $single, string $plural, int $number, string $domain = 'default' )` does not account for that either. The idea was just to have a simple way to add the s if needed, it's not a complete solution until we refactor `Pods::field()` to abstract everything out to a point where the special magic tags can accept extra arguments for different use cases.

I think we could potentially do a `[n]` shortcode like we have for `[once]` and `[each]` in our template handling, but I didn't want to complicate this and get it too far down the rabbit hole. I can back out the `_s` variations as they aren't 100% important.

GaryJones:
> The way WP uses _n( string $single, string $plural, int $number, string $domain = 'default' ) does not account for that either.

But, it does do exactly that - that's basically why it exists.

Also, there seems to be an assumption that just adding `s` will make something plural. It doesn't consistently work even in en_* (think `box` => `boxes`), let alone other locales.

tests/phpunit/includes/tests-traversal.php
```diff
@@ -668,6 +668,9 @@ private function _test_find_base( $variant_id, $options, $query_fields = false )
 
 		$this->assertEquals( (string) $data['id'], (string) $p->id(), sprintf( 'Item ID not as expected (%s) [%s]', $data['field_id'], $variant_id ) );
 
+		$this->assertEquals( 1, (int) $p->field( '_total' ), sprintf( 'Total value not as expected (%s) %s [%s]', $p->total(), var_export( $p->field( '_total' ), true ), $variant_id ) );
```
pglewis: It's a really small thing and this is in the tests, but I think we've switched to consistently using numbered placeholders for `sprintf` in all cases?  @GaryJones ?

pglewis: Actually looking up where we did it, this might only be for translatable strings

GaryJones: The general rule is - if it's more than a single placeholder _in a translatable string_, then yes, it should be numbered, since translators for other locales may need to re-arrange them.

In this case, the string isn't translatable, and the placeholder fillers are not re-used, so there's no technical benefit to using numbered placeholders here.

As an aside, my _personal_ preference (which would seem to go against the current code style in this project), is to lay out sprintf args on new lines, since it makes it easier to read (and also easier to see if sufficient escaping has been added). So instead of:

```php
$this->assertEquals( 1, (int) $p->field( '_total' ), sprintf( 'Total value not as expected (%s) %s [%s]', $p->total(), var_export( $p->field( '_total' ), true ), $variant_id ) );
```

I'd say it was more readable to use:

```php
$message = sprintf(
	'Total value not as expected (%s) %s [%s]',
	$p->total(),
	var_export( $p->field( '_total' ), true ),
	$variant_id
);

$this->assertEquals( 1, (int) $p->field( '_total' ), $message );
```

which doesn't just have a long line of commas and parentheses.

pglewis:
Needs to work with in `[before]`, `[after]`, and `[once]` contexts.

quasel:
Comment from @codente (JME) 
Also thinking a total of related items would be VERY useful
´{@related_field._total_found } ´  ( _total won't really make sense would it? ) 

Agree on that ;)

sc0ttkclark:
After further review, this work will require refactoring of the special Pods shortcodes `[if] [else] [each] [once] [before] [after]` and how they're handled. Currently, they aren't optimal and don't pass around the Pods object state. They create new Pods() objects as they go and it's preventing interacting with the current loop data (`{@_total}`) totally undoable.

JoryHogeveen:
@sc0ttkclark Is this already done for review? I see the todo list isn't marked yet..

sc0ttkclark:
Closing in favor of the work in https://github.com/pods-framework/pods/commit/43a3b2d8fad22cd7543d24634b59ebdc173a9fe5

After abstracting the special magic tags this was much easier to work into place.

Still need to look at supporting magic tags in once/before/after shortcodes in the future.

Issue #4960, opened 2018-05-25
PodsUI/PodsData cache bug


pcfreak30:
See #3577

In `PodsUI::get_row` `$this->pod->fetch();` is called. However in `PodsData::fetch`, it checks if the ID is a number and falls back to a slug. But there seems to be a case where both `id` and `slug` are null.

With object cache, that causes the cache key to be only the pods version number? The result is a weird edge case situation which i have yet to debug the actual saving of the key, but the key never gets cleared and causes the UI to never update, and maybe other side effects.

quasel:
@pcfreak30 is this still an issue did you dig deeper?

pcfreak30:
Yes, I had to re-apply my patch to it my client recently on the latest 2.x branch.

JoryHogeveen:
Closed by #5062

PR #5062, opened 2018-07-12
Fix PodsData fetch for when using object cache and settings pages (possibly other types too)

pcfreak30:
## Description
Fixes issue #4960

## How Has This Been Tested?

Tested on client site to solve issues involving redis cache

## Types of changes
Bug fix

## ChangeLog
If row isnt a number, check if the id property is and use it before falling back to a slug on cases were an id/row is not passed at all. In our case its due to use of PodsData fetch() in PodsUI not giving an id for a settings page.

## Checklist:
- [ * ] My code is tested.
- [ * ] My code follows the WordPress code style.
- [ * ] My code follows has proper inline documentation.


classes/PodsData.php
```diff
@@ -1999,8 +1999,12 @@ public function fetch( $row = null, $explicit_set = true ) {
 			$id   = pods_absint( $row );
 
 			if ( ! is_numeric( $row ) || 0 === strpos( $row, '0' ) || $row != preg_replace( '/[^0-9]/', '', $row ) ) {
-				$mode = 'slug';
-				$id   = $row;
+				if ( ! is_numeric( $this->id ) ) {
```
sc0ttkclark:
```suggestion
				if ( $this->id && ! is_numeric( $this->id ) ) {
```

PR #5080, opened 2018-07-26
Prior to PHP 5.5, 'empty()` only supports variables

### LINKED ISSUE #5079: 2.7.7 + PHP 5.4 is_empty bug?
**Describe the bug**

We're seeing the exception `Can't use function return value in write context` on a number of sites from this line: https://github.com/pods-framework/pods/blob/2.x/2.7.7/classes/fields/time.php#L168

Seems to be an issue with PHP < 5.5. I think this is the problem: https://stackoverflow.com/a/2173318/1110798

edit: the note in the PHP docs about this is here - http://php.net/manual/en/function.empty.php

I see 5.3 in the tests so I assume you still plan to support < 5.5?

**To Reproduce**

Just open the site after updating the plugin. Sometimes it only shows up in WP admin.

## Pods Version
Please provide the version of Pods you are running: 

2.7.7

pglewis:
Fixes #5079

### DIFF
--- a/classes/fields/time.php
+++ b/classes/fields/time.php
@@ -165,7 +165,8 @@ class PodsField_Time extends PodsField_DateTime {
 	 */
 	public function is_empty( $value = null ) {
 
-		return empty( trim ( $value ) );
+		$value = trim ( $value );
+		return empty( $value );
 	}
 
 	/**


GaryJones:
@jrfnl - do you know why the [EmptyNonVariableSniff](https://github.com/PHPCompatibility/PHPCompatibility/blob/master/PHPCompatibility/Sniffs/PHP/EmptyNonVariableSniff.php) didn't pick this up?

GaryJones:
This would have been flagged with PHPCompatibility's [EmptyNonVariableSniff](https://github.com/PHPCompatibility/PHPCompatibility/blob/master/PHPCompatibility/Sniffs/PHP/EmptyNonVariableSniff.php): 

screenshot 2018-07-27 12 57 28

PR #5092, opened 2018-08-08
Add support for Pods::field() 'keyed' argument

sc0ttkclark:
Allows for keys to be maintained for relationships. The default functionality is to do an array_values() on the returned value.

Later: Add support for output type `ids=>names`

## ChangeLog
Added: Support for Pods::fields() argument `keyed` which when set to true will return the array for relationship fields with the IDs used as keys.

### DIFF
--- a/classes/Pods.php
+++ b/classes/Pods.php
@@ -788,6 +788,7 @@ class Pods implements Iterator {
 			'get_meta'    => false,
 			'output'      => null,
 			'deprecated'  => false,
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
This is needed for the FacetWP Pods integration to work properly, it needs access to IDs and Names when using `'output' => 'names'`. By default, it returns an `array_values()` version of the names, which loses the usefulness of the array returned since it no longer has the IDs readily available.

PR #5105, opened 2018-08-16
Feature/5104 update and prep tools for react

### LINKED ISSUE #5104: Update NPM packages and prepare tools for React
This is prep for 2.8 but the build should still work for 2.7.  I would prefer have these changes in place to cut down on future merge issues between 2.7 and 2.8.

pglewis:
Fixes #5104


pglewis:
Branched from 2.8 by mistake, this was supposed to be against 2.x

PR #5106, opened 2018-08-16
Update packages and install tools in prep for React

### LINKED ISSUE #5104: Update NPM packages and prepare tools for React

pglewis:
Fixes #5104


pglewis:
This is pretty much tool-chain only and doesn't need a changelog entry

PR #5113, opened 2018-08-17
PodsDFVText

### LINKED ISSUE #4662: DFV: Text field type
DFV: Text field type 

- [x] Placeholder
- [x] Default value 
- [x] Max length
- [x] Readonly
- [x] Required
- [x] Additional CSS Classes

pglewis:
Fixes #4662


pglewis:
Also @sc0ttkclark, I believe we can delete the old PHP version in `ui/fields`?

ui/js/pods-dfv/_src/text/text.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class PodsDFVText extends React.Component {
+	constructor ( props ) {
```
sc0ttkclark: Could this type of stuff be put in a new PodsDFVReactField type of class where it does the most common stuff and you just extend it to customize?

pglewis:
Definitely makes sense to at least start off by encapsulating the propTypes that will be used everywhere, so done.  Ready for review again.

pglewis:
With password: 
- [ ] Remove prototyping code in `password::input` put there to switch between DFV and the old field
- [ ] Remove old PHP password field
- [ ] Sticking `fieldType` directly on the input's type attrib isn't a great generic solution, create a mapping function

PR #5114, opened 2018-08-17
Always enqueue the DFV script by default if in the admin

### LINKED ISSUE #5111: Always enqueue PodsDFV in the admin by default
It is difficult to anticipate all the places we may need to enqueue the PodsDFV script.  If you extend WordPress media with a DFV field then DFV will need to be enqueued anywhere someone can open a media modal for edit.  That may include via other plugins in cases we simply can't know up front.  

Also provide a filter to bypass loading DFV if desired.

pglewis:
Fixes #5111

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -174,6 +174,18 @@ class PodsAdmin {
 			}//end if
 		}//end if
 
+		/**
+		 * Filter to disable default loading of the DFV script. By default, Pods
+		 * will always enqueue the DFV script if is_admin()
+		 *
+		 * @param bool Whether or not to enqueue by default
+		 *
+		 * @since 2.7.10
+		 */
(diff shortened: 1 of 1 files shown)


classes/PodsAdmin.php
```diff
@@ -174,6 +174,18 @@ public function admin_head() {
...
 
+		/**
+		 * Filter to disable default loading of the DFV script. By default, Pods
+		 * will always enqueue the DFV script if is_admin()
+		 *
+		 * @param bool Whether or not to enqueue by default
+		 *
+		 * @since 2.7.10
+		 */
+		if ( apply_filters( 'pods_default_enqueue_dfv', is_admin() ) ) {
```
sc0ttkclark: You can set this as `true` since the action only runs in the admin area already.

PR #5117, opened 2018-08-17
2.7.10 changelog

pglewis:
- [x] Set release date when known


pglewis:
Could have sworn I switch to 2.x first

PR #5129, opened 2018-08-22
Compatibility with Admin Columns 3.3+

JoryHogeveen:
## Description
Since `ac_is_version_gte()` was removed Pods returned to the old filters of AC that generated deprecated messages.
@DGStefan mentioned this issue and the function `AC()` was also added in 3.0+ so is better to use for this check.

## ChangeLog
Fix compatibility issue with Admin Columns 3.3+

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -451,8 +451,8 @@ class PodsMeta {
 	 */
 	public function integrations() {
 
-		// `ac_is_version_gte` is since AC 3.0+
-		if ( ! function_exists( 'ac_is_version_gte' ) ) {
+		// `AC` is since AC 3.0+
+		if ( ! function_exists( 'AC' ) ) {
 			// Codepress Admin Columns < 2.x
 			add_filter( 'cac/storage_model/meta_keys', array( $this, 'cpac_meta_keys' ), 10, 2 );
 			add_filter( 'cac/post_types', array( $this, 'cpac_post_types' ), 10, 1 );


pglewis:
I fully assume you tested and verified this locally?

JoryHogeveen:
I got notified by @DGStefan about this (didn't notice it myself) so this PR is on behalf of him.
I'll test this locally now.

pglewis:
And thanks @DGStefan for the heads up

JoryHogeveen:
Hold on merging.
In the newer version `AC()` exists only after the `after_setup_theme` hook.

Just discussed with @DGStefan again.
Why not just remove the old hooks, AC v3.0 is already out since may 2017 so we say it's better to just remove the old hooks.

JoryHogeveen:
@pglewis @sc0ttkclark 
Are you ok with removing the old compatibility code?

sc0ttkclark:
Let's only remove compat code on next major release, in a separate PR.

JoryHogeveen:
Ok, waiting for a reply from @DGStefan on how to this the best way.

DGStefan:
In that case, I think that we need an additional check for Admin Columns 3.3+
The two checks that are already there.
An extra check on the constant 'AC_FILE' could work, this should also load the compatibility code for AC3.0+

JoryHogeveen:
@DGStefan 
https://github.com/pods-framework/pods/pull/5129/commits/ba5695602eb12d60d66ee421491e5e1be3072244

Issue #5130, opened 2018-08-22
Extended media, grid mode edit: checkboxes aren't firing change events


pglewis:
The dynamic saving mechanism in the grid mode edit listens for change events that aren't sent.  Trigger the change event manually to ensure DFV checkboxes are saving properly there.

pglewis:
PR: #5131

pglewis:
Let me know if you're able to test this and verify the fix or not, @GitHubGreg

GitHubGreg:
Thanks and sorry for the delay (baby duties). I've tested it and it doesn't seem to change anything. Changing a checkbox in the media modal window doesn't save the change.

GitHubGreg:
I tried again without luck. I tried with disable caching in Dev Tools. I also tried in an incognito browser. And I tried with all functions disabled (I use some jQuery to control the display of the fields with conditional logic). In all cases the changes to checkboxes aren't saved. 

I am going to the main Media menu (/wp-admin/upload.php) in grid view, clicking a PDF item and making the change.

pglewis:
I failed to push the rebuilt bundle.  Sorry for the run-around, the revised PR should work.  

One oddity with saving per change event is each checkbox click does a round-trip save.  If the network is slow it may look odd when you're clicking checkboxes quickly.  You can see this in action in Dev Tools => Network => filter by XHR.

pglewis:
> One oddity with saving per change event is each checkbox click does a round-trip save. If the network is slow it may look odd when you're clicking checkboxes quickly.

I pushed a commit to debounce the change event so it'll only fire after a period with no further change events coming in, but there is still a glitch that needs to be worked out.  When the change event is fired and during the period while the ajax result is pending you're still able to make more changes but they'll be clobbered when the ajax result comes in.  Due to how WordPress does things here the most we may be able to do about this is disable the controls for changes while an ajax result is pending.  I'm still working on that detail.

GitHubGreg:
Thanks very much. I appreciate all the help. I tested what you have so far and it works, thank you.

pglewis:
Do beware of the caveat above, you can get unexpected results with checkboxes simply due to poor timing with the ajax request and no fault of your own.  I _think_ that's all we have left to address on it for the moment though, thankfully.

pglewis:
[...]

PR #5131, opened 2018-08-22
Manually trigger change events on pick/checkboxes for media library in grid mode

pglewis:
Fixes: #5130


JoryHogeveen:
@pglewis Whats the status for this PR? Changes are approved but you removed the QA pending label..

pglewis:
Changes were approved but the PR isn't good enough for release yet IMO.  

This one is really tricky, dealing with various event listeners, debouncing keystrokes to limit AJAX traffic, trying to ensure updates happen when they should without pounding the wire on every keystroke.  

When I left off I intended to rework the PR and attempt to only send JSON updates when the focus is lost.  With all the events going on it requires a lot of manual testing and it got pushed to the bottom of the stack at the time for the priority of Gutenberg compatibility.

pglewis:
> Changes are approved but you removed the QA pending label..

Also, I never set the PR as pending QA as it was never ready for review.

PR #5139, opened 2018-08-29
List View: Don't use an array for the HTML name for single select

### LINKED ISSUE #5138: "Single Select" Relationship Return differently depending on Format
## Issue Overview
"Single Select" Relationship Return differently depending on Format in the filter `pods_api_pre_save_pod_item`

## Expected Behavior
Consitent Behavior 

## Current Behavior
List View - array()
Autocomplete - single value
Radio Buttons  - single value
Drop Down - single value

## Possible Solution
no idea :/ - just tell users to check for an array ;)

pglewis:
Fixes #5138


quasel:
works as expected!

pglewis:
Ready for review @sc0ttkclark

Issue #5136, opened 2018-08-28
Bring form conditionals over to front-end


JoryHogeveen:
Directly related to: #609 
PR: #5137 

This is not the UI part in the field admin (working on that as well) but just bringing over the logic to frontend forms.

Pods already has the logic and handling for meta fields in place for metaboxes, but on frontend forms there is no code at all that checks these conditional options.

PR #5137, opened 2018-08-28
Frontend forms conditional logic

JoryHogeveen:
## Description
Fixes #5136 

## How Has This Been Tested?
I've tested this on a dev environment where I was using PodsForm on the front end. There was a need for conditionals (currently only manually configured).
I also verified this doesn't break the admin side of Pods. No issues found.

## Types of changes
The changes are only in code. No interface changes has been made, those would be related to #609 

## ChangeLog
**Enhancement:** Add pods conditional logic handling on frontend forms.

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style.
- [x] My code follows has proper inline documentation.

PR #5145, opened 2018-09-03
Frontend forms conditional logic v2

JoryHogeveen:
## Description
Fixes: #5136

## How Has This Been Tested?
Enhances dependencies on load + make simple select fields from DFV pick field compatible.
I also added an event listener for when jQuery is loaded in the footer to prevent jQuery errors.

## ChangeLog
See #5137 

## Checklist:
- [x] My code is tested.

### DIFF
--- a/ui/front/form.php
+++ b/ui/front/form.php
@@ -201,18 +201,16 @@ if ( !$fields_only ) {
 		var pods_form_init = true;
 
 		document.addEventListener( "DOMContentLoaded", function() {
-			jQuery(document).ready( function( $ ) {
-				if ( 'undefined' !== typeof jQuery( document ).Pods ) {
+			if ( 'undefined' !== typeof jQuery( document ).Pods ) {
 
-					if ( 'undefined' === typeof ajaxurl ) {
-						window.ajaxurl = '<?php echo pods_slash( admin_url( 'admin-ajax.php' ) ); ?>';
-					}
-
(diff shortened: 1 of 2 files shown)


JoryHogeveen:
I also added an event listener for when jQuery is loaded in the footer to prevent jQuery errors.

ui/front/form.php
```diff
@@ -200,18 +200,20 @@
 	if ( 'undefined' === typeof pods_form_init ) {
 		var pods_form_init = true;
 
-		jQuery(document).ready( function( $ ) {
-			if ( 'undefined' !== typeof jQuery( document ).Pods ) {
+		document.addEventListener( "DOMContentLoaded", function() {
+			jQuery(document).ready( function( $ ) {
```
pglewis: Listening for `DOMContentLoaded` should remove the need for the jQuery ready() block entirely.

JoryHogeveen: Sure? I thought jQuery.ready was an event that was triggered at a later stage.

pglewis: The only real difference between listening for that event and using jQuery's ready is that ready will still be called if the code is executed after `DOMContentLoaded` is fired whereas you'd obviously not catch an one-time event that already fired.  Since the code will be enqueued in the body of the document there should be zero risk of that happening and the ready call is just redundant.

PR #5164, opened 2018-09-28
#4961 - Fix read-only not having any effect on checkbox fields

### LINKED ISSUE #4961: It's impossible to make the yes/no fields read-only
## Issue Overview
It's impossible to make the yes/no fields read-only.

Restricting the field by selecting the `Make field "Read Only" in UI` has no effect.

## Expected Behavior
The yes/no field should become greyed out when `Make field "Read Only" in UI` is selected. See here: https://screenshots.firefox.com/ctgWdV8n5fi7Ygvh/www.datafreedoms.com

## Current Behavior
The field does not become "read-only" because it can still be changed. It should be greyed out so you can see it's current state (yes or no) but are unable to change it. See here: https://screenshots.firefox.com/fRd4CfbLfTP9Re8d/www.datafreedoms.com

## Steps to Reproduce (for bugs)

1. Add a yes/no custom field to a post
2. Under the advanced tab select `Make field "Read Only" in UI` 
3. Go to the post and you will see it is not read-only, because you can change it.

## WordPress Environment

<details>
```
Debug Information

WordPress Version: 4.9.6

PHP Version: 7.0.30-1~dotdeb+8.1

MySQL Version: 5.6.39

Server Software: Apache/2.4.10 (Debian)

Your User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0

Session Save Path: /var/lib/php/sessions

Session Save Path Exists: Yes

Session Save Path Writeable: Yes

Session Max Lifetime: 1440

Opcode Cache:

    Apc: Yes
    Memcached: No
    OPcache: Yes
    Redis: No

Object Cache:

    APC: Yes
    APCu: Yes
    Memcache: No
    Memcached: Yes
    Redis: No

WPDB Prefix: wp_

WP Multisite Mode: No

WP Memory Limit: 40M

Pods Network-Wide Activated: No

Pods Install Location: /home/43307-85855.cloudwaysapps.com/jwzqnqwkyc/public_html/wp-content/plugins/pods/

Pods Tableless Mode Activated: No

Pods Light Mode Activated: No

Currently Active Theme: Pro – Child Theme

Currently Active Plugins:

    Include Me: 1.1.5
    Pods - Custom Content Types and Fields: 2.7.3
    Pods - Custom Points Score:
    Search by Algolia – Instant & Relevant results: 2.11.1
```
</details>

atanas-angelov-dev:
## Description
Fixes #4961 

The "readonly" attribute only prevents the "value" attribute from changing which is not what we want in the case of checkboxes as interacting with a checkbox changes its `checked` attribute instead which is why we need to do this:

1. Add `disabled="DISABLED"` to the checkbox.
2. As a consequence of 1, even if the checkbox default value is set to `1`, the value stored will still be `0` since disabled inputs are not submitted.
3. To fix 2 we add a hidden input immediately after the checkbox only when the checkbox should be checked.
4. The hidden input will have the same name and value as the checkbox and will override it since it appears later in the html hierarchy. A bit hacky but avoids name clashes with other fields and the need for special-case back-end code.

| Default value | Result | Stored Value |
| --- | --- | --- |
| _NONE_ | Checkbox is unchecked and disabled; no hidden input | `0` |
| `0` | Checkbox is unchecked and disabled; no hidden input | `0` |
| `1` | Checkbox is checked and disabled; hidden input is present | `1` |

I decided not to include this lengthy explanation as an inline comment - let me know if I should try to condense it and add it.

Feedback is appreciated.

## How Has This Been Tested?
Manually tested a read-only checkbox field with no default value, 0 and 1 as the default value.

## Types of changes
Bug fix

## ChangeLog
Fix: Checkbox fields not respecting their `Make field "Read Only" in UI` setting.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/ui/fields/checkbox.php
+++ b/ui/fields/checkbox.php
@@ -62,6 +62,7 @@ if ( 0 < $data_count ) {
 
 		if ( pods_var( 'readonly', $options, false ) ) {
 			$attributes['readonly'] = 'READONLY';
+			$attributes['disabled'] = 'DISABLED';
 
 			$attributes['class'] .= ' pods-form-ui-read-only';
 		}
@@ -79,6 +80,12 @@ if ( 0 < $data_count ) {
 		<div class="pods-field pods-boolean"<?php echo $indent; ?>>
 			<input<?php PodsForm::attributes( $attributes, $name, $form_field_type, $options ); ?> />
 			<?php
+			if ( isset( $attributes['readonly'] ) && isset( $attributes['checked'] ) && 'CHECKED' === $attributes['checked'] ) {
+				?>
+				<input type="hidden" name="<?php echo esc_attr( $name ); ?>" value="<?php echo esc_attr( $attributes['value'] ); ?>" />
+				<?php
+			}
+
 			if ( 0 < strlen( $label ) ) {
 				$help = pods_var_raw( 'help', $options );
 


jimtrue:
QA Tests Passed:
- Yes/No Checkbox created with default of 0 and Read Only: Wrote 0
- Yes/No Checkbox created with default of 1 and Read Only: Wrote 1

Issue #4363, opened 2017-07-24
2.6.10 - Warning: A non-numeric value encountered


TerribleGit:
Hey there,

On the Pods help page in wp-admin I get the following error output:

Warning: A non-numeric value encountered in /app/public/wp-includes/SimplePie/Parse/Date.php on line 694

{main}( )	.../admin.php:0
do_action( )	.../admin.php:222
WP_Hook->do_action( )	.../plugin.php:453
WP_Hook->apply_filters( )	.../class-wp-hook.php:323
PodsAdmin->admin_help( )	.../class-wp-hook.php:298
pods_view( )	.../PodsAdmin.php:2312
PodsView::view( )	.../classes.php:206
PodsView::get_template_part( )	.../PodsView.php:118
require( '/app/public/wp-content/plugins/pods/ui/admin/help.php' )	.../PodsView.php:515
SimplePie->get_item_quantity( )	.../help.php:72
SimplePie->get_items( )	.../class-simplepie.php:2812
SimplePie_Item->get_date( )	.../class-simplepie.php:2925
SimplePie_Parse_Date->parse( )	.../Item.php:773
SimplePie_Parse_Date->date_w3cdtf( )	.../Date.php:603

pglewis:
This might be limited to PHP 7.1 or higher and almost certainly a WordPress bug.  We call through to SimplePie with `$feed->get_item_quantity( 10 )`... literally a hard-coded integer.  I'm not seeing any way we could be causing it nor fix it from the Pods side.  

https://github.com/pods-framework/pods/blob/ecbca6f7b78c77b988db75ab5d76fa9e143c823e/ui/admin/help.php#L27

pglewis:
Rather than close this out, I @sc0ttkclark may want to use his influence to get this into trac, if need be.  

In the call stack above we hand over control to SimplePie with a guaranteed numeric value and control never returns to Pods, so it seems clear it's a minor WordPress bug.  May only manifest on PHP 7.1 and above: http://php.net/manual/en/migration71.other-changes.php

pglewis:
All indications are this is coming from WordPress itself and there have been no other/further reports.

jimtrue:
Guessing there's no going to get any traction from the WP side. Can we bypass this library? I didn't check to see what it's doing for us but if it's part of the last forum posts and they're not the WP.org posts, we should be able to remove it completely. 

Which file drives the Help page? I can look at changing it all to our new resources.

pglewis:
[...]

PR #5180, opened 2018-10-26
Pods Help page: swap old forum feed with wordpress.org feed, fix warnings from SimplePie/PHP 7+

### LINKED ISSUE #5172: Remove the pods.io forum feed from the Pods help page and Update for More Support Items
If WordPress provides a feed for their built-in support forums I assume we want to insert that in its place.

pglewis:
Fixes #5172


pglewis:
There is the potential to "fix" #4363 with something like: 

```
// Temporarily turn off warnings and then restore, due to a bug in
// SimplePie's date parser with PHP 7.1+
// See: https://github.com/pods-framework/pods/issues/4363
$old_error_reporting = error_reporting( ~E_WARNING );
$max_items = $feed->get_item_quantity( 10 );
error_reporting( $old_error_reporting );
...
```

... which is effectively "fixing" the check engine light but putting a piece of tape over it so you can't see it.  I would normally be against this sort of thing, but it's a non-vital area and we can't fix SimplePie.

sc0ttkclark:
Doesn’t @$ also silence warnings?

pglewis:
The only other question is if we want to mark resolved issues in the feed via CSS (the title from the feed will be wrapped in a span with the "resolved" class).

sc0ttkclark:
That's a good idea, not a must-have but if you have time for it outside Gb stuff then that's fine.

jimtrue:
If we add the following CSS:
```
ul.ul-disc span.resolved:before {
    content: "Resolved: ";
}
```

This will put Resolved in front of the resolved tickets. @nicdford or @JoryHogeveen if you guys wanna pretty that up anymore, add it in :)

Issue #5192, opened 2018-11-07
Gutenberg Compatibility: Currency Field, Currency Symbol overlays input for numbers.


jimtrue:
**Describe the bug**
Currency Symbol overlays the currency input on Currency Fields in Gutenberg 'compatible' post-types.

**To Reproduce**
1. Create a Currency Field
2. Attempt to Input

**Expected behavior**
This should drop the currency symbol to the left without overlapping input field.

**Screenshots**
![screenshot 2018-11-07 13 49 22](https://user-images.githubusercontent.com/3052875/48153455-3fe7e780-e294-11e8-94bd-9aabda656f60.png)

## Pods Version
2.7.9

## WordPress Environment
WP 5.0 Beta Release

PR #5219, opened 2018-12-05
Fix incorrectly calculated margins for the currency symbol

pglewis:
Fixes: #5192


jimtrue:
QA Tests
- Currency Field on Refresh still has the issue: Fail
- Current Field on resize of window shows correctly.

sc0ttkclark:
Ready for retest

sc0ttkclark:
Just a note for the future, if you go into Gutenberg options and show/hide the meta box in the 'Advanced panels' section, there is no available event I could find that would hook into this type of thing, so if they load up a 'open' meta box in Gutenberg with the panel hidden, then click to show it -- the currency field may not be the correct JS resized width. The solution I have in place to mitigate this somewhat is on hover/focus of the field to run the resize too. It's not the best but it'll work until we have an action to hook into there.

PR #5222, opened 2018-12-05
Fix component header parsing for developer/tableless mode

### LINKED ISSUE #5221: PR #5207 Broke Components List in WP5.0
**Describe the bug**
Components List is empty in WP5.0 2.x branch

**To Reproduce**
1. Components List is not visible in WP5.0 in 2.x branch

sc0ttkclark:
Fixes #5221

## ChangeLog
N/A, introduced as regression only in this release

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/PodsComponents.php
+++ b/classes/PodsComponents.php
@@ -467,19 +467,19 @@ class PodsComponents {
 
 				$component_data['ID'] = sanitize_title( $component_data['ID'] );
 
-				if ( 'off' !== strtolower( $component_data['DeveloperMode'] ) && '0' !== $component_data['DeveloperMode'] ) {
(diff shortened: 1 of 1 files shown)


jimtrue:
QA Passed:
- Pods Admin, Clear Pods Cache after Fix, Components shown
- Fresh install or update, cache clear will not be required.

JoryHogeveen:
This commit negates https://github.com/pods-framework/pods/pull/5207 ?? If so, is that issue still considered fixed?

JoryHogeveen:
Another thing..

`'on' === strtolower( $component_data['TablelessMode'] ) && '1' === $component_data['TablelessMode']`

I guess this is impossible to ever return true since `$component_data['TablelessMode']` can never be both `1` and `on` right? Shouldn't this be `||` ??

PR #4768, opened 2018-02-25
Re Issue #3962: Make auto templates for a Taxonomy Pod behave sensibly

gwhitney:
I am submitting this pull request as suggested in the discussion to Issue #3962, where I requested that auto templates for taxonomy pods not be turned off. Let me start by saying that although I have considerable coding experience, I am relatively inexperienced with git workflows, so my apologies if I do not have the forking/branching structure correct, and if so, I am happy to close this pull request and create  a new better one, with some guidance as to what should be changed.

Next, I should say that this pull request is organized as a series of successive recommendations regarding auto template handling for taxonomy types; you may decide to take almost any initial sequence of these commits and leave the remainder for further deliberation or reject the remainder. 

Finally, let me explain the contents/motivation of each commit further, so you can better decide which to take, despite the risk of tl;dr. (Of course, I personally use and recommend all of these changes.) This list is in the order of the commits in the request:

59d2e7e -- this is just the suggested patch from the discussion concerning #3962. Just by itself it makes auto templates with taxonomies more usable and understandable, in that it changes the default type of the pod created to run the template against from the pod type of the _query_ that generated the post(s) to the post type of the _post_ that is actually currently being displayed.

f201889 -- Another simple change that extends the usefulness of auto templates with taxonomies to those taxonomies which span multiple post types.. By including a magic code like `{@post_type}` in the auto template name, one can use different templates for different post types that might show up in the taxonomy archive.
 
90a5f88 -- I noticed that 'the_content' was hard-coded into the temporary hook removal and re-installation in `load_template`, even though the actual filter used may have been set to something else, and that the priority of the hook was changing from 10.5 to 10. So instead, store the filters which have been hooked and their priorities in a member variable so that they can be maintained through `load_template`. Does not provide any new functionality, but is necessary for following commits, since they are just slightly more intensive in terms of setting/removing hooks.

4b78da6 -- auto templates were not firing in search results, leading to an inconsistent look for post summaries between taxonomy archives and search results. The issue was that no post_type can be inferred from the wp_query when it is a search query -- it is truly difficult to tell what types of posts might be turned up in the search. Therefore, this commit implements a mechanism by which if `$current_post_type` is undetermined by the wp_query outside the loop, it can be recomputed for each post inside the loop by making use of the post_type of the current post. Note that this change does _not_ alter the `$current_post_type` if it was determined by the wp_query itself, e.g. in case of a taxonomy archive; that behavior preserves the operation of auto templating for taxonomy archives.

[...]

PR #5289, opened 2019-02-06
Re Issue #3962: Make auto templates for a Taxonomy Pod behave sensibly

### LINKED ISSUE #3962: Remove option for Auto Template for Taxonomy until it can recognize Post Type Associated to it
Remove option for Auto Template for Taxonomy until it can recognize Post Type Associated to it submitted via Slack by jimtrue

Since Taxonomy Archives in WordPress can't automatically identify the post-type associated underneath them (since Taxonomy archives can cross post-type boundaries), typically the Taxonomy Archive in WordPress only shows the title and permalink. This means the Auto Template option for a Taxonomy won't work as people expect, because it can't show fields from the post-type underneath.

Until we can add some kind of logic/pre-loading to allow Taxonomy archives to display post-type fields, this option should be disabled (and should've never been _enabled_).

sc0ttkclark:
From #4768, fixes #3962 

Via @gwhitney

I am submitting this pull request as suggested in the discussion to Issue #3962, where I requested that auto templates for taxonomy pods not be turned off. Let me start by saying that although I have considerable coding experience, I am relatively inexperienced with git workflows, so my apologies if I do not have the forking/branching structure correct, and if so, I am happy to close this pull request and create  a new better one, with some guidance as to what should be changed.

Next, I should say that this pull request is organized as a series of successive recommendations regarding auto template handling for taxonomy types; you may decide to take almost any initial sequence of these commits and leave the remainder for further deliberation or reject the remainder. 

Finally, let me explain the contents/motivation of each commit further, so you can better decide which to take, despite the risk of tl;dr. (Of course, I personally use and recommend all of these changes.) This list is in the order of the commits in the request:

59d2e7e -- this is just the suggested patch from the discussion concerning #3962. Just by itself it makes auto templates with taxonomies more usable and understandable, in that it changes the default type of the pod created to run the template against from the pod type of the _query_ that generated the post(s) to the post type of the _post_ that is actually currently being displayed.

f201889 -- Another simple change that extends the usefulness of auto templates with taxonomies to those taxonomies which span multiple post types.. By including a magic code like `{@post_type}` in the auto template name, one can use different templates for different post types that might show up in the taxonomy archive.
 
90a5f88 -- I noticed that 'the_content' was hard-coded into the temporary hook removal and re-installation in `load_template`, even though the actual filter used may have been set to something else, and that the priority of the hook was changing from 10.5 to 10. So instead, store the filters which have been hooked and their priorities in a member variable so that they can be maintained through `load_template`. Does not provide any new functionality, but is necessary for following commits, since they are just slightly more intensive in terms of setting/removing hooks.

4b78da6 -- auto templates were not firing in search results, leading to an inconsistent look for post summaries between taxonomy archives and search results. The issue was that no post_type can be inferred from the wp_query when it is a search query -- it is truly difficult to tell what types of posts might be turned up in the search. Therefore, this commit implements a mechanism by which if `$current_post_type` is undetermined by the wp_query outside the loop, it can be recomputed for each post inside the loop by making use of the post_type of the current post. Note that this change does _not_ alter the `$current_post_type` if it was determined by the wp_query itself, e.g. in case of a taxonomy archive; that behavior preserves the operation of auto templating for taxonomy archives.

The remaining five commits really go together as more of a unit, even though they could be teased apart (and the commit messages do give some more detail about what's in each). The underlying idea of these five commits is to allow auto templates for taxonomies also to affect the taxonomy description that shows up at the top of the taxonomy archive (in most/many themes), since that's what I assumed it _would_ modify when I first started using Pods, and it is in fact handy to be able to incorporate custom fields into that description. The mechanism for doing so consists of: allowing the "single" templates for taxonomies as well; arranging that the "single" template for a taxonomy filters `get_the_archive_description` by default (note that filtering `get_the_archive_title` is another totally reasonable use which now can be specified in the auto template settings tab); allowing both the "single" and "archive" filters to be in place simultaneously so that the single (for the description) can apply outside the loop and the archive one can apply inside the loop; arranging that when a taxonomy "single" template runs, it runs against the item (term) in that Taxonomy Pod corresponding to the term of the query; and allowing there to be no "single" template, for the case in which the archive posts should be templated but the description doesn't need to be.

I understand that as a result, there is a lot going on in this pull request. I hope that doesn't reduce the chances of at least the first commit being adopted, which already has significant benefit. And I hope that you are able to consider the full package; all of the changes are in the spirit of that first one, and I believe overall they make the auto template feature significantly more powerful while also making it simpler and more understandable: single templates affect the presentation of the Pod items themselves, archive templates affect the post archives associated with the Pod; for Custom Post Type Pods these are the post archives, as well as search results, and for Taxonomy Pods, these are the Taxonomy archives. Thus, a given post of a Custom Post Type Pod can have appearance A in a single post page, appearance B in the post archive of that post type and in search results, and appearance C in the taxonomy archive of one associated Taxonomy Pod, appearance D in another Taxonomy, and so on. 

In case it's helpful for understanding the changes and their motivation, I use them for content of a publication. The custom post types are different sorts of content (article, letter, etc.). There are taxonomies for the contributor of the content, what issue it is in, what topics it relates to, etc. Articles are presented differently in the list of all articles than in the page for a given contributor (e.g., there's no need to mention the contributor in the latter case). And all of the content types look very different in the page for a given issue, which is (a mockup of) the physical layout of the paper version of the publication.

Hope this is of benefit and interest to you and the community of Pods users.


components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -297,29 +335,49 @@ public function front( $content ) {
...
+			$pod_name = $current_post_type;
+			$pod_item = $post->ID;
+			if ( in_the_loop() ) {
+				$pod_name = $post->post_type;
+			} else {
+				// Outside the loop in a taxonomy, we want the term
+				if ( is_tax() ) {
+					global $wp_query;
+					$obj      = $wp_query->get_queried_object();
+					$pod_item = $obj->slug;
```
JoryHogeveen: Shouldn't this be:
```php
$pod_item = $obj->term_id;
$pod_name = $obj->slug;
```

sc0ttkclark:
```suggestion
					$pod_name = $obj->slug;
					$pod_item = $obj->term_id;
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -297,29 +335,50 @@ public function front( $content ) {
...
-				$content = $this->load_template( $this_pod['archive'], $content, $pods, $this_pod['archive_append'] );
+			$pod_name_and_item = array( $pod_name, $pod_item );
+			/**
+			 * Change which pod and item to run the template against. The
+			 * default pod is the the post type of the post about to be
+			 * displayed, the default item is the post about to be displayed,
+			 * except outside the loop in a taxonomy archive, in which case it
+			 * is the term the archive is for.
+			 *
+			 * @since 2.7.2
```
JoryHogeveen: Update version number
```suggestion
			 * @since 2.7.16
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -78,12 +87,35 @@ public function hook_content(){
...
+
+		foreach ( $this->filtered_content as $filter => $priority ) {
+			add_filter( $filter, array( $this, 'front' ), $priority );
+		}
+	}
+
+	/**
+	 * Remove the hooks specified by the filtered_content member
+	 *
+	 * @since 2.7.2
```
JoryHogeveen: Update version
```suggestion
	 * @since 2.7.16
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -78,12 +87,35 @@ public function hook_content(){
...
+			$this->filtered_content['the_excerpt'] = 10;
 		}
 
+		$this->install_hooks();
+	}
+
+	/**
+	 * Install the hooks specified by the filtered_content member
+	 *
+	 * @since 2.7.2
```
JoryHogeveen: Update version
```suggestion
	 * @since 2.7.16
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -10,6 +10,15 @@
  */
 class Pods_Templates_Auto_Template_Front_End {
 
+	/**
+	 * Currently filtered content functions.
+	 *
+	 * @var array
+	 *
+	 * @since 2.7.2
```
JoryHogeveen: Update version
```suggestion
	 * @since 2.7.16
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Settings.php
```diff
@@ -220,7 +246,7 @@ public function options( $options, $pod ) {
 				}
 
 				$options['pods-pfat']['pfat_archive']['data'] = array( null => __( 'No Archive view template', 'pods' ) ) + ( array_combine( $this->get_template_titles(), $this->get_template_titles() ) );
```
JoryHogeveen: Duplicate code (first part, I can't get multiline suggestion to work)
```suggestion
				$template_titles = array_combine( $this->get_template_titles(), $this->get_template_titles() );
				$options['pods-pfat']['pfat_archive']['data'] = array( null => __( 'No Archive view template', 'pods' ) ) + $template_titles;
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Settings.php
```diff
@@ -220,7 +246,7 @@ public function options( $options, $pod ) {
...
-				$options['pods-pfat']['pfat_single']['data']  = array_combine( $this->get_template_titles(), $this->get_template_titles() );
+				$options['pods-pfat']['pfat_single']['data']  = array( null => __( 'No view template', 'pods' ) ) + array_combine( $this->get_template_titles(), $this->get_template_titles() );
```
JoryHogeveen: Duplicate code (second part)
```suggestion
				$options['pods-pfat']['pfat_single']['data']  = array( null => __( 'No view template', 'pods' ) ) + $template_titles;
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -297,29 +335,50 @@ public function front( $content ) {
...
+				$pod_name = $post->post_type;
+			} else {
+				// Outside the loop in a taxonomy, we want the term
+				if ( is_tax() ) {
+					global $wp_query;
+					$obj      = $wp_query->get_queried_object();
+					$pod_name = $obj->slug;
+					$pod_item = $obj->term_id;
+				}
+			}
```
JoryHogeveen: Just wondering whether this is always correct.
What about Users or Post Type Archives (non-term archives).
Is that something we should look into? Or isn't that supported at all?
Might be better to tackle this separately from this PR.

sc0ttkclark: This appears to be always correct for our supported use cases here.

Issue #4881, opened 2018-05-03
Website & Phone Custom Fields Value Masks are not working properly


TheDude12x:
## Issue Overview
When I use the Phone custom field and select “US (999) 999-9999 x123” as the format, when the field is left blank, it saves “() -” as the value. And when using the Website custom field and chose one of the “force www” as the format, when I enter domain.com, the saved value does not have the “www” added.

## Expected Behavior
When the Phone field is left blank, there should be no value saved. When a value is entered for the Website field without the "www", the "www" should be added to the value. 

## Field Formats to fix

- [ ] Phone should not put anything in the field input on 'blank' values
- [ ] Force WWW on Website/URL fields should add the `www.` to the front of field.

jimtrue:
@pglewis Let me know if these should be two separate bug reports.

TheDude12x:
Will this be addressed soon? I know it's not high priority but I use it quite a bit. Thanks.

jimtrue:
This should be addressed if not before 2.8 at least as part of 2.8.

pglewis:
PRs: #5290, #5291

PR #5290, opened 2019-02-13
Force WWW options have never worked

pglewis:
Part of the fix for #4881


jimtrue:
@pglewis Did this only impact the settings for these two input formats?
- (force www if no sub-domain provided) (screenshot attached)
screenshot 2019-02-18 09 54 46

I did not test the other masks, but let me know if I should.

pglewis:
> Did this only impact the settings for these two input formats?

Yeah, that should be the only corners touched.

jimtrue:
Awesome! Then yes, consider my QA Test Passed

Issue #5278, opened 2019-01-22
Versioned tag names are not compatible with Composer


simensen:
**Describe the bug**

**To Reproduce**
1. Go to https://github.com/pods-framework/pods
2. Click on `Branch: 2.x`
3. Click on `tags`
4. See valid Composer tags: `2.7.9` and `2.7.10`
5. See invalid Composer tags: `2.x/2.7.11` and `2.x/2.7.12`

**Expected behavior**

I would expect (and so would Composer) the tag for version 2.7.12 to be `2.7.12` and not `2.x/2.7.12`.

**Screenshots**

You can see that 2.7.9 and 2.7.10 are the only versions listed as installable [on Packagist](https://packagist.org/packages/pods-framework/pods):

![image](https://user-images.githubusercontent.com/191200/51554309-9cda4f00-1e3a-11e9-8c84-e6beba660ed6.png)

You can see from the tag list [on GitHub](https://github.com/pods-framework/pods) that this looks to be an intentional naming scheme and that `2.7.9` and `2.7.10` were merely... accidents?

![image](https://user-images.githubusercontent.com/191200/51554375-c4c9b280-1e3a-11e9-9a15-1e261737543e.png)

## Pods Version

I'm using `"pods-framework/pods": "^2.7"` but it won't install or upgraded past 2.7.10 due to the fact that Composer does not think `2.x/2.7.11` and `2.x/2.7.11` are versions.

## WordPress Environment

<details>

```
Debug Information
WordPress Version: 4.9.9

PHP Version: 7.1.23

MySQL Version: 5.5.5

Server Software: nginx

Your User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36

Session Save Path: /tmp/sessions

Session Save Path Exists: Yes

Session Save Path Writeable: Yes

Session Max Lifetime: 200000

Opcode Cache:

Apc: No
Memcached: No
OPcache: Yes
Redis: No
Object Cache:

APC: No
APCu: No
Memcache: No
Memcached: Yes
Redis: No
WPDB Prefix: wp_

WP Multisite Mode: No

WP Memory Limit: 40M

Current Memory Usage: 10.914M

Current Memory Usage (real): 2.000M

Pods Network-Wide Activated: No

Pods Install Location: /app/web/wp-content/plugins/pods/

Pods Tableless Mode Activated: No

[...]

PR #5292, opened 2019-02-13
Remove "2.x" prefix from the release tag

pglewis:
Fix for #5278


sc0ttkclark:
QA pass: This change does indeed tag the release in a way compatible with Composer.

PR #5297, opened 2019-02-18
Always use `date_i18n` to ensure proper localization.

### LINKED ISSUE #5296: Date output in magic tags for date fields uses DateTime class by default instead of date_il8n
**Describe the bug**

When outputting a magic tag in a Pods Template/field, etc. the Date Field is not translated to the local language (it does recognize the field display PHP format).

**To Reproduce**
1. Create a Date Field
2. Switch to another language (ie French/fr-BE)
3. Output that field with either `display` or in a magic tag
4. See that the date output doesn't translate to the local language

**Expected behavior**
The date output should be translated locally.

**Possible Workaround**
From notes from Scott, change the route of the field from DateTime >> Unix timestamp >> `date_il8n()`
https://codex.wordpress.org/Function_Reference/date_i18n

JoryHogeveen:
Fixes #5296

### DIFF
--- a/classes/fields/datetime.php
+++ b/classes/fields/datetime.php
@@ -404,13 +404,16 @@ class PodsField_DateTime extends PodsField {
 			// Try field format.
 			$date_local = $this->createFromFormat( $format, (string) $value );
 
+			// Convert to timestamp.
 			if ( $date instanceof DateTime ) {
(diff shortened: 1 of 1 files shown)


brigbe:
I tested on the website where the problem appears and it seems perfect now! Thanks!

PR #5298, opened 2019-02-18
Avoid redundant transient set

### LINKED ISSUE #4690: 2.7.1: using `get_post_meta()` re-triggers `pods_transient_set()` w/ 51kB storage payload on every page load
(EDIT 2018-07-05 still the same with 2.7.6, `PODS_API_CACHE false` gets rid of them, but at what cost?)

I have an archive page where every entry accesses 3-4 pods fields via `get_post_meta()`. Every time I load this page, Query Monitor reports 56 transients set. It's like Pods keeps setting the transients for caching, but never manages to load the data from cache.

Here's an example Query Monitor traceback (all 50+ transient sets are exactly the same, other than the different transient name hash):

Transient name: `pods_33bf92342012bb8c15c50bc755ee0c30`

```
1. hybrid_get_content_template()
wp-content/themes/hybrid-core-4.0/inc/template.php:188
2. the_content()
wp-includes/post-template.php:240
3. apply_filters('the_content')
wp-includes/plugin.php:203
4. do_shortcode()
wp-includes/shortcodes.php:197
5. preg_replace_callback()
wp-includes/shortcodes.php:319
6. do_shortcode_tag()
wp-includes/shortcodes.php:319
7. ConversionXL\Institute\Plugin\Features\MyDashboard->ConversionXL\Institute\Plugin\Features\{closure}()
wp-content/plugins/conversionxl-institute/src/Plugin/Features/MyDashboard.php:67
8. ConversionXL\Institute\Plugin\Features\MyDashboard->render_dashboard_section_certificates()
wp-content/plugins/conversionxl-institute/src/Plugin/Features/MyDashboard.php:110
9. hybrid_get_content_template()
wp-content/themes/hybrid-core-4.0/inc/template.php:188
10. get_post_meta()
wp-includes/post.php:1778
11. get_metadata()
wp-includes/meta.php:485
12. apply_filters('get_post_metadata')
wp-includes/plugin.php:203
13. PodsMeta->get_post_meta()
wp-content/plugins/pods/classes/PodsMeta.php:2392
14. PodsMeta->get_meta()
wp-content/plugins/pods/classes/PodsMeta.php:2883
15. Pods->field()
wp-content/plugins/pods/classes/Pods.php:1217
16. PodsAPI->get_table_info()
wp-content/plugins/pods/classes/PodsAPI.php:8095
17. pods_transient_set()
wp-content/plugins/pods/includes/general.php:1548
18. pods_view_set()
wp-content/plugins/pods/includes/general.php:1441
19. PodsView::set()
wp-content/plugins/pods/classes/PodsView.php:333
```

What's going on here? Is the cache loading mechanism maybe building the transient name differently and missing existing entries + forcing re-caching on every page load?

pglewis:
Fixes #4690

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -8684,7 +8684,7 @@ class PodsAPI {
 				$transient .= '_pre_init';
 			}
 
-			if ( $transient_cached ) {
+			if ( !$transient_cached ) {
 				pods_transient_set( $transient, $info );
 			}
 		}
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -8681,7 +8681,10 @@ class PodsAPI {
 			if ( ! did_action( 'init' ) ) {
 				$transient .= '_pre_init';
 			}
-			pods_transient_set( $transient, $info );
+
+			if ( false !== $_info && is_array( $_info ) ) {
+				pods_transient_set( $transient, $info );
+			}
 		}
 
 		self::$table_info_cache[ $transient ] = apply_filters( 'pods_api_get_table_info', $info, $object_type, $object, $name, $pod, $field, $this );
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -8293,6 +8293,7 @@ class PodsAPI {
 		}
 
 		$_info = false;
+		$transient_cached = false;
 
 		if ( isset( self::$table_info_cache[ $transient ] ) ) {
 			// Prefer info from the object internal cache
@@ -8302,6 +8303,7 @@ class PodsAPI {
 			if ( false === $_info && ! did_action( 'init' ) ) {
 				$_info = pods_transient_get( $transient . '_pre_init' );
 			}
+			$transient_cached = true;
 		}
 
 		if ( false !== $_info && is_array( $_info ) ) {
@@ -8682,7 +8684,7 @@ class PodsAPI {
 				$transient .= '_pre_init';
 			}
 
-			if ( false === $_info || !is_array( $_info ) ) {
+			if ( $transient_cached ) {
 				pods_transient_set( $transient, $info );
 			}
 		}
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -8682,7 +8682,7 @@ class PodsAPI {
 				$transient .= '_pre_init';
 			}
 
-			if ( false !== $_info && is_array( $_info ) ) {
+			if ( false === $_info || !is_array( $_info ) ) {
 				pods_transient_set( $transient, $info );
 			}
 		}


classes/PodsAPI.php
```diff
@@ -8681,7 +8681,10 @@ public function get_table_info( $object_type, $object, $name = null, $pod = null
 			if ( ! did_action( 'init' ) ) {
 				$transient .= '_pre_init';
 			}
-			pods_transient_set( $transient, $info );
+
+			if ( false !== $_info && is_array( $_info ) ) {
```
lkraav: I believe this condition has to be reversed.

```php
if ( false === $_info )
```

Because https://github.com/pods-framework/pods/blob/30dfc0a0822d816b1cd7b729e35028af7cc9e9e7/classes/PodsAPI.php#L8295

We need to avoid re-saving the exact same information that was just grabbed from the transient.

We only save, when transient didn't have anything.

Do I understand this correctly?

pglewis: Yeah, I reversed the entire logic in my head, I think the whole thing needs be negated

classes/PodsAPI.php
```diff
@@ -8682,7 +8682,7 @@ public function get_table_info( $object_type, $object, $name = null, $pod = null
 				$transient .= '_pre_init';
 			}
 
-			if ( false !== $_info && is_array( $_info ) ) {
+			if ( false === $_info || !is_array( $_info ) ) {
```
lkraav: Lgtm, but curious, which line did you identify the path where `$_info` becomes truthy, but a non-array?

pglewis: I simply copied logic from L8307, which has the comment "Data was cached, use that"

classes/PodsAPI.php
pglewis: It was introduced here, so I'll assume there is a reason: https://github.com/pods-framework/pods/commit/e9ad5fc7cdcbd63774c7081e38570946a6f74db5

classes/PodsAPI.php
pglewis: It was introduced here, so I'll assume there is a reason: [e9ad5fc](https://github.com/pods-framework/pods/commit/e9ad5fc7cdcbd63774c7081e38570946a6f74db5)

lkraav: While transient hammering seems to have stopped, on this screen, we still have ~100 of 1-row SQL queries, each with this traceback:

AFTER

```
SELECT item_id, related_item_id, related_field_id     FROM `wp_podsrel`     WHERE           `field_id` = 427525     AND `item_id` IN ( 560046 )         ORDER BY `weight`
```

```
get_post_meta()
wp-includes/post.php:2032
get_metadata()
wp-includes/meta.php:511
apply_filters('get_post_metadata')
wp-includes/plugin.php:208
PodsMeta->get_post_meta()
wp-content/plugins/pods/classes/PodsMeta.php:2556
PodsMeta->get_meta()
wp-content/plugins/pods/classes/PodsMeta.php:3096
Pods->field()
wp-content/plugins/pods/classes/Pods.php:1392
PodsAPI->lookup_related_items()
wp-content/plugins/pods/classes/PodsAPI.php:7769
pods_query()
wp-content/plugins/pods/includes/general.php:47
PodsData::query()
```

But this looks stay exactly the same on both systems and is probably something for another day.

I'd ship this PR in its latest date, tyvm. :taco:

pglewis:
Actually, I think the most recent commit may be a safer and  slightly less fragile way to do this if you can retest one more time.

classes/PodsAPI.php
pglewis: > While transient hammering seems to have stopped, on this screen, we still have ~100 of 1-row SQL queries, each with this traceback:

Yeah, this sounds like a job for a separate issue.  It would help if that info includes basic reproduction instructions, including any Pods exports and/or code that can be dropped into a vanilla WordPress/Pods install for local reproduction.

lkraav:
Lol, yeah latest is most clear.

PR #5310, opened 2019-03-04
Avoid SQL errors when using special magic tags

sc0ttkclark:
`ID = {@get.ID}` should avoid generating SQL errors if ID isn't set.

## QA task list for shortcodes

Test these on a post type created by or extended by Pods. Fill in a few posts and save them. Then try to put one of these shortcodes into the content of the post.

### Basic checks

For these, don't set `?ID=1234` or `?slug=the-slug` in the URL parameter.

- [ ] `[pods id="{@get.ID}"]{@ID}[/pods]` should output no SQL error and should not output anything (no item matched, no ID to show)
- [ ] `[pods slug="{@get.slug}"]{@ID}[/pods]` should output no SQL error and should not output anything (no item matched, no ID to show)
- [ ] `[pods where="t.ID = {@get.ID}"]{@ID}[/pods]` should output no SQL error and should not output anything (no item matched, no ID to show)
- [ ] `[pods where="t.post_name = '{@get.slug}'"]{@ID}[/pods]` should output no SQL error and should not output anything (no item matched, no ID to show)

### Sanitization checks

For these, set `?ID=1234` or `?slug=the-slug` in the URL parameter according to each associated instruction.

- [ ] `[pods where="t.ID = {@get.ID}"]{@ID}[/pods]` -- try passing in icky values and non-numbers into GET to see if it sanitizes the queries properly, set `?pods_debug_sql=1` in the URL (as admin) to see the SQL statement produced (no SQL error should be shown)
- [ ] `[pods where="t.post_name = '{@get.slug}'"]{@ID}[/pods]` -- try passing in icky values into GET to see if it sanitizes the queries properly, set `?pods_debug_sql=1` in the URL (as admin) to see the SQL statement produced (no SQL error should be shown)


JoryHogeveen:
Sidenote:
I'd add `PODS_SHORTCODE_ALLOW_EVALUATE_TAGS` to the list of functions in general.php and include it in the health check and query monitor reports.

sc0ttkclark:
@JoryHogeveen Updated the description with some better QA steps

JoryHogeveen:
@sc0ttkclark 
All test output the current post ID on screen (using shortcode in post content).
Getting no SQL errors though.

sc0ttkclark:
OK sounds like there's more work to do here. We'll want to ensure that if they had provided a value, it shouldn't try to default to the current post type / post ID, it should always assume it was not a match and fail.

JoryHogeveen:
@sc0ttkclark 
Can I assume that all filter/query type parameters should prevent reverting to post default?
And is so, is the list below accurate according to you?
```php
	$query_tags = array(
		'name',
		'id',
		'slug',
		'select',
		'join',
		'order',
		'orderby',
		'limit',
		'where',
		'having',
		'groupby',
		'search',
		'pagination',
		'page',
		'offset',
		'filters',
		'filters_label',
		'filters_location',
		'pagination_label',
		'pagination_location',
	);
```

See: https://github.com/pods-framework/pods/blob/2.x/includes/general.php#L787-L822

JoryHogeveen:
@sc0ttkclark 
I have a test commit ready that prevents reverting to the current post but that will return the error `Please provide a Pods name`. Is that correct? Or in other words, is a Pods name **always** required for the shortcode to work?

classes/PodsAdmin.php
```diff
@@ -3741,6 +3741,10 @@ public function add_debug_information( $info ) {
 					'label' => __( 'Pods API Cache Activated' ),
 					'value' => ( pods_api_cache() ) ? __( 'Yes', 'pods' ) : __( 'No', 'pods' ),
 				),
+				'pods-shortcode-allow-evaluate-tags'                  => array(
```
tr1b0t: :warning: Array double arrow not aligned correctly; expected 1 space(s) between "'pods-shortcode-allow-evaluate-tags'" and double arrow, but found 18.

<sub>hash: c4e4274379c3436ca550a7cad8cb6b12</sub>

includes/data.php
```diff
@@ -1469,15 +1469,16 @@ function pods_mb_substr( $string, $start, $length = null, $encoding = null ) {
  * Evaluate tags like magic tags but through pods_v
  *
  * @param string|array|object $tags     String to be evaluated
- * @param bool                $sanitize Whether to sanitize tags
+ * @param bool                $sanitize Whether to sanitize
```
tr1b0t: :no_entry_sign: Parameter comment must end with a full stop

<sub>hash: d139d38053044c0116efc6bf64900c3d</sub>

includes/data.php
```diff
@@ -1497,13 +1498,9 @@ function pods_evaluate_tags( $tags, $sanitize = false ) {
...
 	}
 
-	$callback = 'pods_evaluate_tag';
-
-	if ( true === $sanitize ) {
-		$callback = 'pods_evaluate_tag_sanitized';
-	}
-
-	return preg_replace_callback( '/({@(.*?)})/m', $callback, (string) $tags );
+	return preg_replace_callback( '/({@(.*?)})/m', function ( $tag ) use ( $sanitize, $fallback ) {
```
tr1b0t: * :no_entry_sign: Opening parenthesis of a multi-line function call must be the last content on the line
* :no_entry_sign: Only one argument is allowed per line in a multi-line function call

<sub>hash: bf5d58b62929713734109281b0896c26</sub>

includes/data.php
```diff
@@ -1497,13 +1498,9 @@ function pods_evaluate_tags( $tags, $sanitize = false ) {
...
+		return pods_evaluate_tag( $tag, $sanitize, $fallback );
+	}, (string) $tags );
```
tr1b0t: * :no_entry_sign: Only one argument is allowed per line in a multi-line function call
* :no_entry_sign: Closing parenthesis of a multi-line function call must be on a line by itself

<sub>hash: 94872091deaa30cc31a470869522392c</sub>

sc0ttkclark:
> I have a test commit ready that prevents reverting to the current post but that will return the error `Please provide a Pods name`. Is that correct? Or in other words, is a Pods name **always** required for the shortcode to work?

@JoryHogeveen Can you provide the commit diff or make a new branch off of this one and PR into this branch so it can be looked at/tweaked?

JoryHogeveen:
@sc0ttkclark See #5520

JoryHogeveen:
@sc0ttkclark 
This question is still on the table:
> ~~I have a test commit ready that prevents reverting to the current post~~ but that will return the error Please provide a Pods name. Is that correct? Or in other words, is a Pods name always required for the shortcode to work?

sc0ttkclark:
Pod name is always required unless it’s able to detect current post type / taxonomy from the page it’s on

JoryHogeveen:
Check, but when the shortcode has query params, is the name still required then?

sc0ttkclark:
Name is always required, it’s ID that isn’t required in some cases (like List, which will list items but doesn’t need ID)

Issue #4656, opened 2017-12-11
DFV: Number field type


PodsBot:
DFV: Number field type submitted via Slack by sc0ttkclark

quasel:
![bildschirmfoto 2018-08-30 um 12 23 01 pm](https://user-images.githubusercontent.com/9464217/44862190-6f69ea80-ac7a-11e8-8a75-dacd8400a40e.png)

should add the option to "style" your number yourself e.g leading zeroes `00001`

PR #5334, opened 2019-04-03
Skeleton for number field

pglewis:
Issue: #4656


ui/js/pods-dfv/src/fields/number/number.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { PodsDFVBaseInput } from 'pods-dfv/src/components/base-input';
+
+export const PodsDFVNumber = ( props ) => {
+
+	return (
+		<PodsDFVBaseInput
+			type="text"
```
sc0ttkclark: What about html5 mode?

ui/js/pods-dfv/src/components/base-input.js
```diff
@@ -20,6 +20,8 @@ export const PodsDFVBaseInput = ( props ) => {
 			readOnly={ !!props.fieldConfig.readonly }
 			onChange={ props.onChange || handleChange }
 			onBlur={ props.onBlur }
+			min={ props.min }
```
sc0ttkclark: What if prop not defined?

pglewis: React will omit any attributes that are undefined or null

PR #5375, opened 2019-05-09
Try and get single select to work with [each] templating

### LINKED ISSUE #4507: Single Select vs. Multi Select in Files Fields or Relationships can't handle EACH
Single Select vs. Multi Select in Files Fields or Relationships can't handle EACH submitted via Slack by jimtrue

sc0ttkclark:
## Description
Single select relationships don't work for `[each]` when templating.

Fixes #4507

## ChangeLog
Support single select relationships in templates when using `[each]`

## Checklist:
- [x] My code is tested.


sc0ttkclark:
That's progress :)

OK so the only remaining issue here is special tags like _img while in `[each]`?

jimtrue:
Yep, seems to be only the {@_img} and {@_src} which means you can't get the images to show in a single image setup inside the each. You do have access to the Image ID with {@ID} so that should work as a fallback.

JoryHogeveen:
@jimtrue If you have the time, my latest comment fixes file types, could you test as well?

sc0ttkclark:
Looks good, thanks @JoryHogeveen

PR #5397, opened 2019-06-25
Wrap shortcodes with exception handling

### LINKED ISSUE #5279: Bug: Shortcode with error makes page uneditable in the Block Editor
**Describe the bug**

If you use a shortcode in the Block Editor (Gutenberg) with an error in the shortcode, the error causes the block editor to automatically preview so the error takes focus and makes the page uneditable.

**To Reproduce**
1. Go to a page and use the Block Editor (instead of the Classic Editor)
2. Create a shortcode with an error (I used an orderby non-existent field)
3. The AutoSave will fail, you can try to preview, but editing in the block editor will no longer work.
4. You can edit it with the Classic Editor

**Expected behavior**
It should provide the SQL Error and fail gracefully

## Pods Version
Please provide the version of Pods you are running: 
Pods 2.7.12

**Possible Workaround**
Discussed with @sc0ttkclark: should add a shortcode_handler that wraps around the shortcode output, checks for error and provides a more useful message.

sc0ttkclark:
## Description
Shortcodes can cause pages to die when they encounter SQL errors. This PR addresses that and introduces exception message returning in those cases.

## ChangeLog
Fix - Shortcodes no longer stop the page from loading when they encounter SQL errors

## Checklist:
- [x] My code is tested.

Fixes #5279


includes/general.php
```diff
@@ -684,6 +684,23 @@ function pods_access( $privs, $method = 'OR' ) {
...
+/**
+ * Shortcode support for use anywhere that support WP Shortcodes
+ *
+ * @param array  $tags    An associative array of shortcode properties
+ * @param string $content A string that represents a template override
+ *
+ * @return string
+ * @since 2.7.13
+ */
+function pods_shortcode_run( $tags, $content = null ) {
```
JoryHogeveen: @sc0ttkclark  I assume this is a private function? Maybe prepend it with an underscore?

includes/general.php
```diff
@@ -1024,6 +1041,23 @@ function pods_shortcode( $tags, $content = null ) {
...
+/**
+ * Form Shortcode support for use anywhere that support WP Shortcodes
+ *
+ * @param array  $tags    An associative array of shortcode properties
+ * @param string $content Not currently used
+ *
+ * @return string
+ * @since 2.7.13
+ */
+function pods_shortcode_form_run( $tags, $content = null ) {
```
JoryHogeveen: @sc0ttkclark I assume this is a private function? Maybe prepend it with an underscore?

includes/general.php
pglewis: Not a fan of underscores for private functions, personally, it's a convention that isn't always clear and doesn't enforce anything.  Comments in the function's header is a better idea for cases like this where it can't just be made private, IMO.  Just my $.02, not a huge deal either way.

sc0ttkclark: We'll refactor this in the future when we expand shortcodes/blocks more. Not anything to worry about right now though.

includes/general.php
sc0ttkclark: We'll refactor this in the future when we expand shortcodes/blocks more. Not anything to worry about right now though.

sc0ttkclark: Future: Classes with private methods

sc0ttkclark:
This change only returns the error, it's supposed to prevent the die() itself. Let me check deeper into why the die() might still be occurring.

JoryHogeveen:
@sc0ttkclark 
Found an issue with `PodsData::query()`. In the beginning you check `$wpdb->show_errors` but the if statement only passes if `show_errors` is true. It doesn't get the actual status and uses that.
Also see `pods_error()` in general.php.

With my latest commit it does. So if debug is disabled it will return an exception, if debug is enabled it will exit.

includes/general.php
```diff
@@ -1027,7 +1090,7 @@ function pods_shortcode_form( $tags, $content = null ) {
 
 	$tags['form'] = 1;
 
-	return pods_shortcode( $tags );
+	return pods_shortcode( $tags, $content );
```
sc0ttkclark: Good fix!

includes/general.php
```diff
@@ -255,6 +265,10 @@ function pods_debug( $debug = '_null', $die = false, $prefix = '_null' ) {
 
 	$pods_debug ++;
 
+	if ( ! pods_is_debug_display() ) {
+		return;
```
sc0ttkclark:
```suggestion
		// Log errors if we do not display them.
		error_log( 'Pods error: ' . $error );

		return;
```

PR #5399, opened 2019-06-26
Add Pods debug information to Site Info area

sc0ttkclark:
## Description
WP now has a great new Site Info area, it provides more details we could use when debugging so this adds our unique debug info to that and we can simplify our debug process going forward with that.

## ChangeLog
* Added: Add debug information for Pods to Site Health Info area

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/PodsAdmin.php
+++ b/classes/PodsAdmin.php
@@ -3717,6 +3717,10 @@ class PodsAdmin {
 					'label' => __( 'Pods Install Location', 'pods' ),
 					'value' => PODS_DIR,
 				),
+				'pods-developer'                  => array(
+					'label' => __( 'Pods Developer Activated' ),
+					'value' => ( pods_developer() ) ? __( 'Yes', 'pods' ) : __( 'No', 'pods' ),
+				),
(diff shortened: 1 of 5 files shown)


classes/PodsAdmin.php
```diff
@@ -57,6 +57,9 @@ public function __construct() {
 
 		add_action( 'admin_head-media-upload-popup', array( $this, 'register_media_assets' ) );
 
+		// Add our debug to Site Info.
+		add_filter( 'debug_information', array( $this, 'add_debug_information' );
```
sc0ttkclark:
```suggestion
		add_filter( 'debug_information', array( $this, 'add_debug_information' ) );
```

includes/general.php
```diff
@@ -388,6 +388,22 @@ function pods_tableless() {
...
 
+/**
+ * Determine if Light Mode is enabled
+ *
+ * @return bool Whether Light Mode is enabled
+ *
+ * @since 2.7.13
+ */
+function pods_light() {
+
```
sc0ttkclark:
```suggestion
```

Issue #5403, opened 2019-07-01
Date/Time field is displaying UTC instead of local time.


tgspacelabs:
**Describe the bug**

I use Pods to create a list of date/times for today's date to display a status value.  The display used to work perfectly for some time now.  Today I noticed that the time appears to be displaying UTC rather than local time.  

[https://www.skydivesnohomish.com/contact/tandem-class-status/](https://www.skydivesnohomish.com/contact/tandem-class-status/)

**To Reproduce**

**Expected behavior**
In the example page URL it used to display 12pm to 7:30pm and now appears to be UTC times which are +7 hours.

**Screenshots**
![image](https://user-images.githubusercontent.com/24533761/60455166-9a9ace80-9bea-11e9-8458-63ede2a21d81.png)

## Pods Version
Please provide the version of Pods you are running: 
Version: 2.7.13

## WordPress Environment

<details>

Copy and paste your System Details Tools, Site Health, Info. Use the 'Copy Site Info to Clipboard' and paste the results over this line.

</details>

## Pods Package Export (helpful!)

```
Copy and Paste the JSON Export from **Pods Admin, Migrate: Packages, Export** in WordPress admin here
```

**Additional context**

**Possible Workaround**
If you have discovered a workaround, please include it below.

tgspacelabs:
Rolling back to Version 2.7.12 fixed the problem.

![image](https://user-images.githubusercontent.com/24533761/60455702-ffa2f400-9beb-11e9-823b-9462a2990856.png)

pglewis:
Most likely the same as #5402, though this one probably should be the canonical issue because I do believe it's directly related to UTC.  This is reproduced locally and I'm looking into it.

pglewis:
Appears to be introduced here: https://github.com/pods-framework/pods/pull/5297/files

kengmick:
I believe this same problem in Pods Version 2.7.13 is showing up in admin and front end of a development site, delaying launch, https://www.mokshayoga.kenmick.com/ where I'm using local time DST Chicago = -5 UTC.  

Class start time of noon displays as 5pm, on edit displays as 5pm, can adjust to noon, and on save reverts to 5pm.  
see video:  https://cl.ly/eff91a75e298
I think this problem is causing multiple reports yet to be noted as duplicates, including:  

- this issue [Date/Time field is displaying UTC instead of local time. #5403](https://github.com/pods-framework/pods/issues/5403) status: Open; type: bug
[...]

PR #5409, opened 2019-07-05
Always convert with UTC timezone

### LINKED ISSUE #5382: UTC time issue on custom pods date time fields
Hi there, 
I have an issue with hours display for custom pods date/time fields. 

My WP should be in UTC+2 (Belgian hour) but if I do so, when adding a date into one of my pods, it actually remove 3 hours to the time added. Let's I write 10 am, after saving it will show 7 am. The only way to have a correct display of the website is to define it at UTC-1, but then the saving time is incorrect... 

In WP general settings, the time stamp under the UTC settings is showing the correct time if I choose UTC+2. Also this issue doesn't seem to affect the time of the post (it's not changing after saving).

Any idea of what can be wrong? Thanks in advance!

Brigitte

---

EDIT BY JORY:

## WordPress Environment

<details>
```
Copy and paste your System Details from **Pods Admin > Settings > Debug Information** in WordPress admin here.
```
</details>

## Pods Package Export (helpful!)

```
Copy and Paste the JSON Export from **Pods Admin, Migrate: Packages, Export** in WordPress admin here
```

JoryHogeveen:
Fixes #5382, #5402, #5403 and probably more

## Description

Currently after date/time values are stored into the database the date/time field convert them using the WordPress timezone setting (in cases where DateTime is available). This conversion is incorrect since it returns a different value than the user's input.

## ChangeLog
* Always convert database value for date/time fields with UTC timezone to maintain the actual value.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/fields/datetime.php
+++ b/classes/fields/datetime.php
@@ -394,14 +394,14 @@ class PodsField_DateTime extends PodsField {
 		if ( ! empty( $value ) && ! in_array( $value, array( '0000-00-00', '0000-00-00 00:00:00', '00:00:00' ), true ) ) {
 			// Try default storage format.
 			$date = $this->createFromFormat( static::$storage_format, (string) $value );
-			
+
 			// Convert to timestamp.
 			if ( $date instanceof DateTime ) {
 				$timestamp = $date->getTimestamp();
 			} else {
 				// Try field format.
 				$date_local = $this->createFromFormat( $format, (string) $value );
-				
+
 				if ( $date_local instanceof DateTime ) {
 					$timestamp = $date_local->getTimestamp();
 				} else {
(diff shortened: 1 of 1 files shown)


pglewis:
Works for me

brigbe:
Works for me too.

triclub:
Thanks, all. We're having this issue in our wordpress instance. Any idea when it will be available in an update?

pglewis:
> Thanks, all. We're having this issue in our wordpress instance. Any idea when it will be available in an update?

This fix is now merged into 2.x and a hotfix release primarily for this issue should go out this week.

edellingham:
I've updated to the latest version and have cleared cache, but I'm still experiencing this issue.  The time always reverts to UTC when making a change to a Date/Time Field.  I tried to use the hotfix referenced here, but the file I guess has been deleted since the official fix was added.

JoryHogeveen:
@edellingham 
Could you test #5418 ?
I've omitted the timezone fully in that PR so it will use the server's timezone.

pglewis:
> I've updated to the latest version and have cleared cache, but I'm still experiencing this issue. The time always reverts to UTC when making a change to a Date/Time Field. I tried to use the hotfix referenced here, but the file I guess has been deleted since the official fix was added.

@edellingham: If you have the time to create a new issue with direct steps to reproduce the problem as minimally as possible it would help a lot.  I was able to recreate it before the hotfix and verified that this PR fixed the issue that I had reproduced locally, so I don't yet know what the circumstances are to reproduce it.  Plus we'd have a clear thread of conversation for the remaining problem, this a closed PR.

jeiseman:
This update causes incorrect behavior (sometimes). It will now (sometimes) assume the stored time is UTC time and then the time gets reset to the current timezone. So, if the time stored is 9:00am, it gets displayed as 5:00am (in the New York timezone). This problem is resolved by reverting to Pods version 2.7.13. The weird thing is that it doesn't always happen. Sometimes when I do $pod->display('time') to display the time field it will work correctly but in a different context it displays the incorrect time. I haven't figured out why it happens in one context but not in others.
Changing the code to the following fixes the problem:
`       // $time = $rpod->display('time');
        $strtime = $rpod->field('time');
        $t = strtotime($strtime);
        $time = date("g:ia", $t);
`

JoryHogeveen:
@jeiseman See #5418 !

PR #5428, opened 2019-07-30
Do not allow reserved keywords do be used for pods

JoryHogeveen:
Fixes #4122 (and probably more)

## Todo:
- [x] Validate pod names (PHP)
- Validate field names (PUNT TO 2.8)
- JS validation (PUNT TO 2.8)

## How Has This Been Tested?

Returns an notice (AJAX) if a reserved keyword is used.

## Types of changes

Use the full list of reserved keywords when validating Pod names.

## ChangeLog

* Prevent creation of Pods using reserved keywords from WordPress Codex and Pods.

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style.
- [x] My code follows has proper inline documentation.

PR #5441, opened 2019-08-16
Fix issue with extending post types and reserved keywords.

JoryHogeveen:
## Description

Fix an issue related to #5428 (reserved keywords). When extending CPT or taxonomies it gave errors on (for example) `page` post type since that is also a reserved keyword.

## Types of changes

I've brought the `create_extend` parameter into `save_pod`. If that is set and is set to `extend` it will validate against existing CPT/Tax objects. If that name exists the name will still be valid.

## ChangeLog

No need for an extra changelog.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -1664,7 +1664,7 @@ class PodsAPI {
 	 * $params['object'] string The object being extended (if any)
 	 * $params['storage'] string The Pod storage
 	 * $params['options'] array Options
-	 * $params['extend'] bool Extending an existing content type?
+	 * $params['create_extend'] string Create or Extend a Content Type
 	 *
 	 * @param array    $params    An associative array of parameters
 	 * @param bool     $sanitized (optional) Decides whether the params have been sanitized before being passed, will
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -1679,7 +1679,7 @@ class PodsAPI {
 		$tableless_field_types    = PodsForm::tableless_field_types();
 		$simple_tableless_objects = PodsForm::simple_tableless_objects();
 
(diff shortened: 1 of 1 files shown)


classes/PodsAPI.php
```diff
@@ -1753,12 +1760,28 @@ public function save_pod( $params, $sanitized = false, $db = true ) {
...
 		} else {
+			// New pod (create).
+
+			if (
+				in_array( $params->name, pods_reserved_keywords(), true )
+				&& in_array( pods_v( 'type', $params ), array( 'post_type', 'taxonomy' ), true )
+			) {
+				$valid_name = false;
+
+				if ( $extend ) {
```
sc0ttkclark:
```suggestion
				if ( ! $extend ) {
```

@JoryHogeveen It seems like we only want to validate on creating new ones, extending existing ones is not under our control and the plugin/theme that created the post type / taxonomy would have to mitigate their own risks.

JoryHogeveen: Correct, but this code works the other way around.
If rejects all reserved keywords (see line 1767).
When if finds a reserved keyword THEN it checks if `$extend` is true. If it's true then existing CPT/Tax are allowed.

JoryHogeveen: I'll add inline documentation

JoryHogeveen:
With #5453 merged the errors are shown again.

- Tested adding Pods with reserved keywords. Correctly returns error message.
- Tested extending post types that are reserved keywords (page). Extending works.
- Tested creating exiting post type. Returns error message.
- Tested with WP_DEBUG on and off. Both return error messages (#5453) 

IMO, Q&A is passed.

Issue #5452, opened 2019-08-23
Proper usage of `pods_error()`


JoryHogeveen:
**Describe the bug**

Currently `pods_error()` is both used for debug and validation. This causes confusion in usage.
In https://github.com/pods-framework/pods/commit/3339eaa486dc0e0fd8eb14ed20ad4700dbef59ce I added a check for debug before enabling the display of errors, which is logical in a debug situation.
But in PodsAPI (adding Pods and Fields) the same function is also used for valitaion, so in this case `pods_error()` should always return a message.

Probably related: #5450 

**Possible solutions:**
- Separate functions for error handling in debug vs validation situations. Both are errors but different kind..
- Enhance the code in pods_error. Maybe add a third parameter `$debug = true` that enables/disables the check for `WP_DEBUG` and `WP_DEBUG_DISPLAY`. If set to false it skips the debug validation.

Issue #5450, opened 2019-08-22
JS validation for file upload field


nicdford:
**Describe the bug**
It looks like if you have a file upload field that is required and isn't given a value, it isn't erroring out as it should and instead is acting like its saving, but doesn't save any of the new/updated values in the entire for,.

**To Reproduce**
1. add file up load field, required
2. add other field, required
3. enter values for second field, but not file upload, hit save

**Expected behavior**
File upload field should error out making the user know that it is a required field.

JoryHogeveen:
Hi @nicdford 
Could you test with #5453?

Issue #5440, opened 2019-08-15
Relationship field that is labeled 'status' will not save.


raoulunger:
**Describe the bug**
When a relationship field is labeled 'status', the field will disappear.

**To Reproduce**
1. Edit a pod
2. Add relationship field
3. label it 'status'
4. Save the field (not the pod!). The field will save as expected, and can be moved around to change the order of fields (provided there are already more fields in the pod.)
5. But: when the entire pod is saved, the relationship field will have disappeared. It is also not present when adding a new pod item, or when adding magical tags in a pod template.

**Expected behavior**
Relationship fields with any other label than 'status' do not show this problem. So it is a specific problem with the word 'status'. It is not in the list of WPs 'Reserved Key Words', so we should be able to use it. It is also a word that is fairly likely to be used in a taxonomy context, so it would be good if this bug were fixed. I took me a day to find out it was a semantic issue.

**Screenshots**
n.a.

## Pods Version
Version 2.7.14

### wp-core ###

version: 5.2.2
site_language: en_US
user_language: en_US
permalink: undefined
https_status: true
user_registration: 0
default_comment_status: closed
multisite: false
user_count: 1
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: /var/www/vhosts/ulrikkereedfindenegg.nl/httpdocs
wordpress_size: 96.75 MB (101454912 bytes)
uploads_path: /var/www/vhosts/ulrikkereedfindenegg.nl/httpdocs/wp-content/uploads
uploads_size: 1.35 MB (1420098 bytes)
themes_path: /var/www/vhosts/ulrikkereedfindenegg.nl/httpdocs/wp-content/themes
themes_size: 8.41 MB (8817125 bytes)
plugins_path: /var/www/vhosts/ulrikkereedfindenegg.nl/httpdocs/wp-content/plugins
plugins_size: 17.24 MB (18075106 bytes)
database_size: 4.98 MB (5226496 bytes)
total_size: 128.74 MB (134993737 bytes)

### wp-active-theme ###

name: WP Bootstrap Starter Child
version: 3.2.3.1565018185
author: Afterimage Designs
author_website: https://afterimagedesigns.com/
parent_theme: WP Bootstrap Starter
[...]

Issue #5435, opened 2019-08-08
requirement field doesn't give error


justaguy84:
## Issue Overview
image required field doesn't give error when saving the post and the value of the field is missing.
post is saved (title and content) but other pods fields are not updated.

## Expected Behavior
When saving a post but missing value in a required field, an error should be displayed and the post shouldn't be saved before filling all required fields.

## Current Behavior
post is saved but custom fields are not.

## Steps to Reproduce (for bugs)
1. unfortunately, as the website is now live, I can't provide links and accounts to test this.
2. I do have this in my function: add_filter( 'pods_pod_form_success_message', 'pods_submit' ) , but even when commenting this out, issue is the same.

## Possible Solution
error message on submit post

## WordPress Environment

<details>
`
WordPress Version: 5.2.2

PHP Version: 7.1.30

MySQL Version: 5.6.40

Server Software: Apache

Your User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36

Session Save Path: /tmp

Session Save Path Exists: Yes

Session Save Path Writeable: Yes

Session Max Lifetime: 1440

Opcode Cache:

Apc: No
Memcached: No
OPcache: Yes
Redis: No
Object Cache:

APC: No
APCu: No
Memcache: No
Memcached: Yes
Redis: No
WPDB Prefix: wpfv_

WP Multisite Mode: No

WP Memory Limit: 40M

Current Memory Usage: 19.686M

Current Memory Usage (real): 22.000M

Pods Network-Wide Activated: No

Pods Install Location: /wp-content/plugins/pods/

Pods Tableless Mode Activated: No

Pods Light Mode Activated: No

Currently Active Theme: Findme Child

Currently Active Plugins:

CIO Custom Fields Importer Pro: 1.3.1
Classic Editor: 1.5
Contact Form 7: 5.1.3
Contact Form 7 - Dynamic Text Extension: 2.0.2.1
Elated Booking: 1.1
Elated Core: 1.4.1
Elated Instagram Feed: 1.0
Elated Listing: 1.4
Elated Membership: 1.1
Elated Twitter Feed: 1.0
Envato Market: 2.0.1
FacetWP: 3.3.9
FacetWP - Load More: 0.4.3
FacetWP - Pods integration: 1.2.2
FacetWP - Relevanssi integration: 0.5.1
Layered Popups: 6.54
[...]

PR #5453, opened 2019-08-26
Improve error handling in pods_error()

JoryHogeveen:
Fixes: #5452
Fixes: #5450
Fixes: #5440
Fixes: #5419
Fixes: #5435

## Description

- Revert changes in pods_error related to pods_is_debug_display() from #5397.
- Add pods_is_debug_display() check in pods_shortcode() before returning exception message. Use `pods_debug()` as fallback if debug display is disabled (logs error)
- Fix log error in `pods_debug()`
- Fixed missing translation calls

## ChangeLog

* Fix & improve error handling & debug logs

## Checklist:
- [ ] My code is tested.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -261,7 +261,7 @@ class PodsAPI {
 			 */
 			$post_error = $post_data['ID'];
 
-			return pods_display_error( $post_error->get_error_message(), $this );
+			return pods_error( $post_error->get_error_message(), $this );
 		}
 
 		$this->save_post_meta( $post_data['ID'], $post_meta, $strict, $fields );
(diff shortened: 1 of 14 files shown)


JoryHogeveen:
@sc0ttkclark 

I'm thinking.. the usage of `pods_error()` is mostly for UI (validation) purposes. Debugging with pods_error() was only in a few cases.
What do you think about turning this PR around.

Current:
- `pods_error()` has 3 params with `WP_DEBUG` check `true` by default.
- `pods_display_error()` is the same but with calls `pods_error()` with the third param as `false`.

Proposal:
- `pods_error()` still has 3 params but with `WP_DEBUG` check `false` by default.
- We could keep `pods_display_error()` as an API option..
- Add `pods_debug_error()` which sets the third parameter (check debug) to `true`.

I think the latter has some upsides:
- Less code modifications
- Better backwards compatibility since `pods_error()` only recently checked for WP_DEBUG. That is the reason for all the bug reports.
- Clearer API to use

Let me know your thoughts, I have a commit ready to push.

JoryHogeveen:
Ready for review

Issue #5468, opened 2019-09-13
Tests have been broken since reserved names were added


jamesgol:
Tests have been broken.

PR #5474, opened 2019-09-19
UnitTests: Extend objects when using reserved keywords

JoryHogeveen:
Fixes: #5468 
This will then accept `post`, `page` etc. as Pods object names.

Fixes: #5475 
- Deprecation warning for uppercase + package wimg/php-compatibility

## Checklist:
- [x] My code is tested.

### DIFF
--- a/tests/phpunit/includes/testcase.php
+++ b/tests/phpunit/includes/testcase.php
@@ -649,6 +649,12 @@ class Pods_UnitTestCase extends \WP_UnitTestCase {
 						self::$builds[ $pod_type ][ $object ][ $storage_type ]['fields'][ $field['name'] ] = $field;
 					}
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/includes/testcase.php
```diff
@@ -649,6 +649,12 @@ public static function _initialize_config() {
 						self::$builds[ $pod_type ][ $object ][ $storage_type ]['fields'][ $field['name'] ] = $field;
 					}
 
+					if ( in_array( $pod['name'], pods_reserved_keywords() ) ) {
```
sc0ttkclark:
```suggestion
					if ( in_array( $pod['name'], pods_reserved_keywords(), true ) ) {
```

Issue #5481, opened 2019-09-24
Required checkbox on Pods Form not working on user extended


bunglegrind:
**Describe the bug**
I've extended wp users with a yes/no field type (Input type: checkbox) - required field and text field - required too. 
Users update their profiles in the frontend within a custom page where pods->form is embedded. Even if a user does not flag the checkbox profile is updated. This contradicts the backend behavior, where the other field is not updated if the checkbox isn't flagged

**To Reproduce**
1. extend user with a text field and yes/no field - input checkbox - both required.
2. create a custom template for a page where the following snippet is embedded:

```
if (is_user_logged_in()) {
    $user_id = get_current_user_id();
    $userpod = pods('user', $user_id);
    $fields = array('my_required_text_field', 'my_required_checkbox');
    echo $userpod->form($fields, 'Update');
} else {
   wp_die("forbidden", 403);
}
```

3. write something in the text field but do not flag the checkbox
4. Message "Form submitted successfully" appears...

4bis. Try the very same steps in the backend (wp-admin/profile.php). The error message "my_required_checkbox is empty" appears. 

**Expected behavior**
the Backend behavior is right. An error message must appear. 

## Pods Version
2.7.15

## WordPress Environment
5.2.3
`
### wp-core ###

version: 5.2.3
site_language: en_US
user_language: en_US
permalink: /%year%/%monthnum%/%day%/%postname%/
https_status: false
user_registration: 0
default_comment_status: open
multisite: false
user_count: 2
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: C:\wamp64\www\wordpress
wordpress_size: 37.54 MB (39367723 bytes)
uploads_path: C:\wamp64\www\wordpress/wp-content/uploads
uploads_size: 365.25 KB (374011 bytes)
themes_path: C:\wamp64\www\wordpress/wp-content/themes
themes_size: 27.48 MB (28816544 bytes)
plugins_path: C:\wamp64\www\wordpress/wp-content/plugins
plugins_size: 5.13 MB (5379044 bytes)
database_size: 1.09 MB (1141880 bytes)
total_size: 71.60 MB (75079202 bytes)

### wp-active-theme ###

name: Design Italia
version: 0.28
author: Marco Gargano
author_website: http://marcogargano.com/
parent_theme: none
[...]

PR #5498, opened 2019-10-14
Fix Required checkbox field value

bunglegrind:
## Description
the field is now filled with the same value inserted in the backend.
Fixes or at least mitigates #5481

This patch fixes the UI, but the same problem may arise again when the post request is performed manually via curl or whatever with user inserted data (See my last post in #5481)

 ## How Has This Been Tested?
By inspecting the data transmitted by the browser in both cases, and how the BE is managing null value with respect to the 0 value.

### DIFF
--- a/ui/js/jquery.pods.js
+++ b/ui/js/jquery.pods.js
@@ -188,7 +188,7 @@
 
                             if ( $el.is( 'input[type=checkbox]' ) && !$el.is( ':checked' ) ) {
                                 if ( $el.is( '.pods-boolean' ) || $el.is( '.pods-form-ui-field-type-boolean') )
-                                    val = 0;
+                                    val = null;
                                 else
                                     return true; // This input isn't a boolean, continue the loop
                             }


JoryHogeveen:
Closing in favor of #5508

bunglegrind:
Well, it still fixes the asimmetry between FE in javascript and BE in php.

JoryHogeveen:
PHP code suggests it prefers integers (`0` or `1`), not `null`, even though it accepts any scalar value.
Correct me if I'm wrong but it looks like this PR would actually create asymmetry between JS and PHP.
Let me know and I'll revisit your findings, thanks!

bunglegrind:
This is not what I'm seeing when I'm inspecting the POST request from user-edit.php

The pods field is missing when the checkbox is not flagged!  

Whereas from pods->form (javascript, front-end) is set to 0.  My pull request fixes this asymmetry.

Issue #5479, opened 2019-09-23
Relationship Field with Simple Custom Defined List doesn't save value in Pod item if it contains an apostrophe


TheDude12x:
**Describe the bug**
Relationship Field with Simple Custom Defined List doesn't save value if it contains an apostrophe.

**To Reproduce**
1. Add Field > Select Field Type 'Relationship'
2. Select Field Type 'Relationship' related to 'Simple (custom defined list)'
3. Enter a value that contains an apostrophe (ie - Do's and Don'ts)
4. When trying to save to Pod item it doesn't save but will display in the drop-down

**Expected behavior**
Expect the value to be saved.

## Pods Version
2.7.15

## WordPress Environment

<details>

`
### wp-core ###

version: 5.2.3
site_language: en_US
user_language: en_US
permalink: /%postname%
https_status: true
user_registration: 0
default_comment_status: closed
multisite: false
user_count: 17
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: /nas/content/live/advocatecap
wordpress_size: 84.73 MB (88845022 bytes)
uploads_path: /nas/content/live/advocatecap/wp-content/uploads
uploads_size: 938.67 MB (984271320 bytes)
themes_path: /nas/content/live/advocatecap/wp-content/themes
themes_size: 4.26 MB (4465628 bytes)
plugins_path: /nas/content/live/advocatecap/wp-content/plugins
plugins_size: 81.85 MB (85821640 bytes)
database_size: 31.50 MB (33027745 bytes)
total_size: 1.11 GB (1196431355 bytes)

### wp-dropins (1) ###

advanced-cache.php: true

### wp-active-theme ###

name: Advocate Capital Inc.
version: 1
author: Altrumedia LLC
author_website: (undefined)
parent_theme: Genesis
theme_features: menus, post-thumbnails, title-tag, automatic-feed-links, body-open, genesis-inpost-layouts, genesis-archive-layouts, genesis-admin-menu, genesis-import-export-menu, genesis-readme-menu, genesis-customizer-theme-settings, genesis-customizer-seo-settings, genesis-auto-updates, genesis-breadcrumbs, genesis-menus, genesis-structural-wraps, widgets, html5, genesis-accessibility, genesis-responsive-viewport, custom-header, genesis-after-entry-widget-area, genesis-footer-widgets
theme_path: /nas/content/live/advocatecap/wp-content/themes/genesis
[...]

Issue #1943, opened 2014-01-07
Simple Relationship data does not sanitize.


SalocinDotTEN:
Using the latest version 2.3.18 and the data entered in Simple Relationship appears as one line separated by "n".

Using Wordpress 3.8.

Thanks.

jimtrue:
@SalocinDotTEN Can you please share the simple relationship list you created, as well as what content type you are working with and how you are displaying this field? That would be very helpful in determining if this is a bug or an issue with your use that we can help you with.

[--- Linked with ASANA.com using Unito.io
#commenter Jim True
---[aa]

jimtrue:
Hi @Shelob9,

The content type I was working with renders a drop down list.

Relationship [type: pick] 
› Simple (custom defined list) (Custom)

In the Pods editor, there is a text area "Custom Defined Options" where we put the list items in separate lines. Eg..

One
Two
Three

After saving, the list comes out in the editor as...

OnenTwonThree

We correct it and save, only to have it appear like above again.

We have not had any issue with this prior to Wordpress 3.8.
Is there a patch already available for this?

[--- Linked with ASANA.com using Unito.io
#commenter Jim True
---[aa]

jimtrue:
They should be in the form of value|Label, like this:
one|One
two|Two
three|Three

[--- Linked with ASANA.com using Unito.io
#commenter Jim True
---[aa]

jimtrue:
Thanks @Shelob9. It seems to be working just fine now. I shall notify my web dev team. :)

[--- Linked with ASANA.com using Unito.io
#commenter Jim True
---[aa]

jimtrue:
If it's not working unless you use keys here, then there's still a bug to fix I think.

[--- Linked with ASANA.com using Unito.io
#commenter Jim True
---[aa]

PR #5517, opened 2019-11-05
Unslash simple relationship values to support saving quoted values

JoryHogeveen:
Fixes: #2360
Fixes: #5479
Closes: #3616 (Sorry @jamesgol, I think this might be a better and simpler fix)
Closes: #1943 (Old related issue)

## Description

This PR makes sure simple custom relationship options are unslashed before comparing to input data.
This needs to be done because input data is unslashed as well. If we want to correctly compare input data to available options both sides need to be the same.

## Checklist:
- [x] My code is tested.


sc0ttkclark:
@jamesgol can you have a look to see if this helps resolve your problems?

Issue #5534, opened 2019-11-15
DateTime field forces default value of time()


nobleclem:
**Describe the bug**
When a `Date/Time` field used with no "Default Value" specified the field defaults to the current date/time (GMT if HTML5 and localtime if not).

**To Reproduce**
Assuming you extend the "Post" content type and have a `Date/Time` field with no default value and its not required.
1. Posts -> Add New
2. Save Draft
3. Refresh Edit Page
4. Value should be present in Date/Time field when none was entered.

This issue is more obvious if you enable html5 input for the Date/Time field but this happens on both input types.  For the non-HTML5 input type the value immediately shows in the "hidden" input that follows the actual input of the field.

**Expected behavior**
If a Date/Time field is NOT required and no value is entered into the field then the value saved should either be `0000-00-00 00:00:00` or `''` or `NOT EXISTS`.

## Pods Version
2.7.16.2

**Additional context**
It seems as that the core of this bug has existed for some time but with two recent changes #5460 _(yes my issue and I must have missed this test case when testing the fix for it)_ & #5488 / #5528 which seems to have brought this to light.  Until these two changes were made the raw issue never presented itself as a noticeable bug.

**Possible Workaround**
The only work around I have found was this:
```
add_filter( 'pods_form_ui_field_datetime_formatter', function( $datetime, $format, $date ){
    if( !$date ) {
        return false;
    }

    return $datetime;
}, 10, 3 );
```
along commenting out line 126 in ui/fields/datetime.php
`$mysql_value = date_i18n( $mysql_format );`

## Possible Fix
The issues seems to be two files attempting to default the value to the current date time if no value exists.  I have tested this fix in a dev environment and that seems to work.  The workaround I am using in production while this bug exists.

**remove lines 125-126 in ui/fields/datetime.php**
```
} else {
	$mysql_value = date_i18n( $mysql_format );
```

**change lines 719-721 in classes/fields/datetime.php**
from
```
if ( empty( $date ) ) {
	$timestamp = time();
} else {
```
to
```
$timestamp = null;

if( !empty( $date ) ) {
```

JoryHogeveen:
Hello @nobleclem 
Could you please make sure your date/time fields allow empty values?

> For the non-HTML5 input type the value immediately shows in the "hidden" input that follows the actual input of the field.
What do you mean?

nobleclem:
Hi @JoryHogeveen 

[...]

PR #5538, opened 2019-11-19
DateTime: Use is_empty() to centralize validation

JoryHogeveen:
Fixes: #5534 (verified by user)
Fixes: #5544

## Description

See issue.

## Types of changes

Everywhere we check for empty date values I've replace the array search with the is_empty() method. This centralizes the empty validation checks.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/fields/datetime.php
+++ b/classes/fields/datetime.php
@@ -308,13 +308,7 @@ class PodsField_DateTime extends PodsField {
 	 */
 	public function validate( $value, $name = null, $options = null, $fields = null, $pod = null, $id = null, $params = null ) {
 
-		if ( ! empty( $value ) && ( 0 === (int) pods_v( static::$type . '_allow_empty', $options, 1 ) || ! in_array(
-			$value, array(
-				'0000-00-00',
-				'0000-00-00 00:00:00',
-				'00:00:00',
-			), true
-		) ) ) {
+		if ( ! $this->is_empty( $value ) ) {
 			$js = true;
 
 			if ( 'custom' !== pods_v( static::$type . '_type', $options, 'format' ) ) {
@@ -331,10 +325,9 @@ class PodsField_DateTime extends PodsField {
 
 				return sprintf( esc_html__( '%1$s was not provided in a recognizable format: "%2$s"', 'pods' ), $label, $value );
 			}
-		}//end if
+		}
 
 		return true;
-
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


classes/fields/datetime.php
```diff
@@ -345,15 +338,9 @@ public function pre_save( $value, $id = null, $name = null, $options = null, $fi
...
-			$value, array(
-				'0000-00-00',
-				'0000-00-00 00:00:00',
-				'00:00:00',
-			), true
-		) ) ) {
+		if ( ! $this->is_empty( $value ) ) {
 			$value = $this->convert_date( $value, static::$storage_format, $format );
-		} elseif ( 1 === (int) pods_v( static::$type . '_allow_empty', $options, 1 ) ) {
+		} elseif ( pods_v( static::$type . '_allow_empty', $options, 1 ) ) {
```
sc0ttkclark: The option could come as a string and evaluate as true though, right?

classes/fields/datetime.php
```diff
@@ -369,13 +356,7 @@ public function ui( $id, $value, $name = null, $options = null, $fields = null,
...
 		$value = $this->display( $value, $name, $options, $pod, $id );
 
-		if ( 1 === (int) pods_v( static::$type . '_allow_empty', $options, 1 ) && ( empty( $value ) || in_array(
-			$value, array(
-				'0000-00-00',
-				'0000-00-00 00:00:00',
-				'00:00:00',
-			), true
-		) ) ) {
+		if ( $this->is_empty( $value ) && pods_v( static::$type . '_allow_empty', $options, 1 ) ) {
```
sc0ttkclark: Same here

classes/fields/datetime.php
```diff
@@ -416,7 +397,7 @@ public function format_value_display( $value, $options, $js = false ) {
 			}
 
 			$value = date_i18n( $format, $timestamp );
-		} elseif ( 0 === (int) pods_v( static::$type . '_allow_empty', $options, 1 ) ) {
+		} elseif ( ! pods_v( static::$type . '_allow_empty', $options, 1 ) ) {
```
sc0ttkclark: Same here

classes/fields/datetime.php
JoryHogeveen: Why would it come as a string? And even if it did come as `'0'` it would still be `false`.
I think it's good to simplify the way we validate boolean types.

classes/fields/datetime.php
JoryHogeveen: See first comment

classes/fields/datetime.php
JoryHogeveen: See first comment

classes/fields/datetime.php
sc0ttkclark: I double checked '0' evaluates to false, nevermind!

bodhi1970:
Empty value still shows a date in DATE FIELD, how to avoid that.

JoryHogeveen:
@bodhi1970 
Please open a new issue with a more detailed explanation. For example, HTML5 or regular field view?

Issue #5539, opened 2019-11-21
Number/Currency field forces 0 as empty value


JoryHogeveen:
**Describe the bug**
Reported on Slack.
Numeric fields do not allow an actual empty value (`''`) so deleting a value results in `0` instead of `null`.

## Pods Version
2.7.16.2

vosykapavel:
I have the same issue:
- in default wordpress backend admin metabox
- in the database, after I use `$pod->add( $data )` with code
- in frontend after I show the pods form (when user delete value, the zero is back immediately)
```
$fields = [
        // default null not helping
        'field' => ['default' => null],
        // default '' null not helping
        'field' => ['default' => ''],
];

$my_pod->form($fields);
```

Even when the number is not required. 🤔 

![image](https://user-images.githubusercontent.com/3764279/141795035-0d0ec5d3-6666-4f99-b004-52f78b457d9a.png)

JoryHogeveen:
Hi @vosykapavel Please check on the latest Pods version: 2.8.4 which has been release yesterday.

PR #5540, opened 2019-11-21
Number/Currency: Use is_empty() to centralize validation

JoryHogeveen:
Fixes: #5539 

NOTE: Punted to 2.8 in favor of #5628 

## Types of changes

Everywhere we check for empty number values I've used the is_empty() method. This centralizes the empty validation checks.

This PR still works with required options etc.

@sc0ttkclark I'm only wondering what to return if a value is empty.
Currently `pre_save` returns `null` in that case which removes the meta entry in the database. I'd say we should return an empty string to keep the meta row in the database. What do you think? (Also, how will this respond with repeatable fields in 2.9?)

## Checklist:
- [x] My code is tested.


classes/fields/currency.php
```diff
@@ -285,6 +285,11 @@ public function validate( $value, $name = null, $options = null, $fields = null,
 	 */
 	public function pre_save( $value, $id = null, $name = null, $options = null, $fields = null, $pod = null, $params = null ) {
 
+		if ( $this->is_empty( $value ) ) {
+			// Don't enforce a default value here
```
tr1b0t: :no_entry_sign: Inline comments must end in full-stops, exclamation marks, or question marks

<sub>hash: eb40e2b8a7fb30232e0dde51e4509213</sub>

classes/fields/number.php
```diff
@@ -250,11 +249,9 @@ public function validate( $value, $name = null, $options = null, $fields = null,
 		$dot         = $format_args['dot'];
 
 		$check = str_replace(
-			array( $thousands, $dot, html_entity_decode( $thousands ) ), array(
-				'',
-				'.',
-				'',
-			), $value
+			array( $thousands, $dot, html_entity_decode( $thousands ) ),
+			array( '', '.', '', ),
```
tr1b0t: * :no_entry_sign: Comma not allowed after last value in single-line array declaration
* :no_entry_sign: Comma not allowed after last value in single-line array declaration

<sub>hash: 360ed25b00088f5e0370af2acdb62f07</sub>

classes/fields/number.php
```diff
@@ -303,7 +305,7 @@ public function pre_save( $value, $id = null, $name = null, $options = null, $fi
 	 */
 	public function format( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
 
-		if ( null === $value ) {
+		if ( $this->is_empty( $value ) ) {
```
sc0ttkclark: `is_empty()` will catch `0` and `0.00` values too, not sure we want that applied here.

classes/fields/number.php
```diff
@@ -274,6 +271,11 @@ public function validate( $value, $name = null, $options = null, $fields = null,
 	 */
 	public function pre_save( $value, $id = null, $name = null, $options = null, $fields = null, $pod = null, $params = null ) {
 
+		if ( $this->is_empty( $value ) ) {
```
sc0ttkclark: `is_empty()` will catch `0` and `0.00` values too, not sure we want that applied here.

classes/fields/currency.php
```diff
@@ -327,8 +332,8 @@ public function pre_save( $value, $id = null, $name = null, $options = null, $fi
 	 */
 	public function format( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
 
-		if ( null === $value ) {
-			// Don't enforce a default value here
+		if ( $this->is_empty( $value ) ) {
```
sc0ttkclark: `is_empty()` will catch `0` and `0.00` values too, not sure we want that applied here.

classes/fields/currency.php
```diff
@@ -285,6 +285,11 @@ public function validate( $value, $name = null, $options = null, $fields = null,
 	 */
 	public function pre_save( $value, $id = null, $name = null, $options = null, $fields = null, $pod = null, $params = null ) {
 
+		if ( $this->is_empty( $value ) ) {
```
sc0ttkclark: `is_empty()` will catch `0` and `0.00` values too, not sure we want that applied here.

classes/fields/number.php
JoryHogeveen: `0` and `0.00` will not be seen as numeric values:
https://github.com/pods-framework/pods/blob/feature/5539-number-empty/classes/fields/number.php#L167-L177

sc0ttkclark:
This one is causing a test fail, need to dig into it more to see what's causing that.

JoryHogeveen:
@sc0ttkclark See commit https://github.com/pods-framework/pods/pull/5540/commits/fa4813c7f29114f82a1ab49acf582c00073eff9a
`0` wasn't considered a valid value before but this PR fixed that. `0` is now a valid numeric value. Only `''`, `false` or `null` are invalid values

sc0ttkclark:
This PR introduces a backwards compatibility issue which is why the test was failing for the `if` conditional shortcode. I think there's a case to be made that 0 is empty for is_empty() checks, while still allowing 0 as an acceptable non-empty value in the areas it checks the logic.

PR #5550, opened 2019-12-06
Allow No (0) as an answer for radio and select boolean fields

### LINKED ISSUE #5549: Allow "No" as an answer for required boolean fields
Reported by @quasel  on Slack
Currently No returns an error when a required boolean field isn't set to "yes".
For the Checkbox type this is logical but for radio and select it isn't since "No" is a valid (selectable) answer.

JoryHogeveen:
Fixes #5549 

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/fields/boolean.php
+++ b/classes/fields/boolean.php
@@ -178,7 +178,9 @@ class PodsField_Boolean extends PodsField {
 
 		$yes_required = ( 'checkbox' === pods_v( static::$type . '_format_type', $options ) );
 
-		if ( pods_v( 'required', $options ) ) {
+		$required = (int) pods_v( 'required', $options );
+
+		if ( 1 === $required ) {
 			if ( $yes_required ) {
 				if ( 0 === $check ) {
 					$errors[] = __( 'This field is required.', 'pods' );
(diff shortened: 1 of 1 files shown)


quasel:
Thank you! PR Background - from an API or in my case a GF form submit I should be able to set a field to 0/No because required in this case means a "valid value" as it does for all other field types… 

I do understand that for a checkbox consent field's point of view required should mean "yes" but that makes it hard e.g in an api case to distinguish between "UNSET, not provided, provided a wrong value,…) but yeah that's an edge case ^^

classes/fields/boolean.php
```diff
@@ -173,12 +172,20 @@ public function data( $name, $value = null, $options = null, $pod = null, $id =
...
-	
+
 		$errors = array();
 		$check  = $this->pre_save( $value, $id, $name, $options, $fields, $pod, $params );
 
-		if ( 1 === (int) pods_v( 'required', $options ) && 0 === $check ) {
-			$errors[] = __( 'This field is required.', 'pods' );
+		$yes_required = ( 'checkbox' === pods_v( static::$type . '_format_type', $options ) );
+
+		if ( pods_v( 'required', $options ) ) {
```
sc0ttkclark:
```suggestion
		$required = (int) pods_v( 'required', $options );

		if ( 1 === $required ) {
```

Issue #5577, opened 2020-01-23
PHP Warning: Illegal offset type in isset or empty in /classes/fields/pick.php on line 1734


JoryHogeveen:
**Describe the bug**

It looks like in some cases `$value` values are non-scalar:
https://github.com/pods-framework/pods/blob/2.x/classes/fields/pick.php#L1721-L1749
```php
			if ( null === $data ) {
				$data = $this->get_object_data( $object_params );
			}

			$data = (array) $data;

			$key = 0;

			if ( is_array( $value ) ) {
				if ( ! empty( $data ) ) {
					$val = array();

					foreach ( $value as $k => $v ) {
						if ( isset( $data[ $v ] ) ) {
							if ( false === $raw ) {
								$k = $v;
								$v = $data[ $v ];
							}

							$val[ $k ] = $v;
						}
					}

					$value = $val;
				}
			} elseif ( isset( $data[ $value ] ) && false === $raw ) {
				$key   = $value;
				$value = $data[ $value ];
			}//end if
```

Stack trace:
```
wp-content/plugins/pods/classes/fields/pick.php:1734
PodsField_Pick->simple_value()
	wp-content/plugins/pods/classes/fields/pick.php:1625
PodsField_Pick->ui()
	wp-content/plugins/pods/classes/PodsForm.php:1510
PodsForm::field_method()
	wp-content/plugins/pods/classes/PodsMeta.php:655
PodsMeta->cpac_meta_value()
	wp-includes/class-wp-hook.php:288
apply_filters('ac/column/value')
	wp-content/plugins/codepress-admin-columns/classes/ListScreen.php:908
AC\ListScreen->get_display_value_by_column_name()
	wp-content/plugins/codepress-admin-columns/classes/ListScreen/Post.php:63
AC\L\Post->manage_value()
	wp-includes/class-wp-hook.php:288
do_action('manage_keralayer_posts_custom_column')
	wp-admin/includes/class-wp-posts-list-table.php:1265
WP_Posts_List_Table->column_default()
	wp-admin/includes/class-wp-list-table.php:1361
WP_List_Table->single_row_columns()
	wp-admin/includes/class-wp-posts-list-table.php:1298
WP_Posts_List_Table->single_row()
	wp-admin/includes/class-wp-posts-list-table.php:741
WP_Posts_List_Table->_display_rows()
	wp-admin/includes/class-wp-posts-list-table.php:718
WP_Posts_List_Table->display_rows()
[...]

PR #5578, opened 2020-01-23
Fix get_meta() cache loop

JoryHogeveen:
Fixes: #5577 

## Description

The if conditions didn't make sense and returns nested empty arrays (`array( [0] => array() )`) in some occasions (not possible for metadata).

## Types of changes

This PR also refactors that part of code to remove duplicate lines and make it more readable. The whole code block is now half the size it was before.

## How Has This Been Tested?

Needs proper testing!!
Testers can do this by using magic tags and pods shortcode. Preferably with traversing to related objects.

## Checklist:
- [x] My code is tested.


JoryHogeveen:
My tests worked for this. Might be good to have a second/third contributor to test this PR.

PR #5571, opened 2020-01-09
Add Friends of Pods callout

sc0ttkclark:
## Description
This PR adds a new Friends of Pods callout so it can let people know we're seeking more funding.

## Screenshots
image

## ChangeLog
* Added: New callout for our Friends of Pods program will show up on the Pods Admin > Pods list so we can let people know how to donate (props @nicdford)

PR #5611, opened 2020-03-05
Enhance Friends of Pods callout

JoryHogeveen:
Related: #5571

This PR fixes some CSS issues with the FoP callout.

I did not make any major changes, just a hotfix to make it "act" the same as WP core.

@sc0ttkclark @nicdford Personally I'd prefer not to use flexbox and just rely on WordPress core CSS to keep it all clean. Why the choice of flexbox? `.pods-admin` is used in various different screens alongside with WP core classes.

## Checklist:
- [ ] My code is tested.
- [ ] My code follows the WordPress code style.


ui/admin/settings.php
```diff
@@ -1,5 +1,5 @@
 <div class="wrap pods-admin">
-	<div class="pods-admin__content-container">
+	<div class="pods-admin-container">
```
sc0ttkclark: The `__` is BEM I believe https://en.bem.info/methodology/naming-convention/ and I see that used more and more alongside newer React/JS projects. I think this might be fine as it was.

JoryHogeveen: Ah, that explains!

JoryHogeveen: Maybe PodsUI should be updated as well then because it still uses `pods-admin-container`.
I'm all for using a naming convention, but if we do we should update everything.. not just the newly added parts because that creates a mess..

PR #5617, opened 2020-03-09
Set type detection based on classes instead of parameters

JoryHogeveen:
Currently we detect the current queried object based on the object parameters. Using instanceof would be more reliable.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/classes/Pods.php
+++ b/classes/Pods.php
@@ -288,16 +288,16 @@ class Pods implements Iterator {
 			if ( $queried_object ) {
 				$id_lookup = true;
 
-				if ( isset( $queried_object->post_type ) ) {
+				if ( $queried_object instanceof WP_Post ) {
 					// Post Type Singular.
 					$pod = $queried_object->post_type;
-				} elseif ( isset( $queried_object->taxonomy ) ) {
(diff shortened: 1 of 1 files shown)


classes/Pods.php
```diff
@@ -288,16 +288,16 @@ public function __construct( $pod = null, $id = null ) {
...
-				} elseif ( isset( $queried_object->taxonomy ) ) {
+				} elseif ( $queried_object instanceof WP_Term ) {
 					// Term Archive.
 					$pod = $queried_object->taxonomy;
-				} elseif ( isset( $queried_object->user_login ) ) {
+				} elseif ( $queried_object instanceof WP_User ) {
 					// Author Archive.
 					$pod = 'user';
-				} elseif ( isset( $queried_object->public, $queried_object->name ) ) {
+				} elseif ( $queried_object instanceof WP_Post_Type && $queried_object->public ) {
```
sc0ttkclark:
```suggestion
				} elseif ( $queried_object instanceof WP_Post_Type ) {
```

We didn't previously check if it was public

JoryHogeveen: Oh right, it was only an `isset`, not `! empty`

Issue #2823, opened 2015-04-03
Private visibility settings are not recognized under Advanced Options, Default Status 'Private'


PodsBot:
Private visibility settings are not recognized under Advanced Options, Default Status 'Private' submitted via Slack by jimtrue

jimtrue:
Verified by @jimtrue: Created a fresh Pod CPT in 2.5.1.2. Those are actually my screenshots above showing the verification process. Advanced options on the Pod, set Default Status to 'Private'. When saving pods, they are created as Default 'Draft' instead.

This may be related to #1832

sc0ttkclark:
How are you adding the post? How are you saving the post?

jimtrue:
The standard expected WordPress way: Add New 'Post Type' from the Admin Menu. Clicking the publish on the screen. It would be 'expected' the Visibility above would be flipped to 'Private' (that was the original poster's expectation, and mine).

sc0ttkclark:
Clicking "Publish" will publish the post, clicking "Save Draft" will save a draft of the post. There's no default handling for WP core's post editor that we've built in. This setting is explicitly for new posts saved through the Pods API or through a custom `$pod->form()`

jimtrue:
Hmmm, that might be the general confusion then. People are expecting that to 'override' the setting of the Editor, I believe. We just need to clarify that in the Help text for that section then and that will alleviate that particular issue and let them know if they want these settings recognized they need to call their entry with a form shortcode, $pod->form() or Pods API.

sc0ttkclark:
We should update the help text, or find a way to default the post status to Private, which may be difficult to do without JS and that's hacky.

jimtrue:
I'd say just the help text. Overriding the editor is outside the scope, isn't it? I know these settings are 'configurable' in register_post_type(), correct? Is there an official way to override them in the WP editor?

sc0ttkclark:
Some people do it on post saving hook that lets you filter values to be saved -- http://wordpress.stackexchange.com/questions/113540/how-to-set-default-status-to-custom-post-type

But that's still confusing to the user because it doesn't change UI and reflect to the user what they're actually about to do on save.

sc0ttkclark:
WP creates the post here: https://github.com/WordPress/WordPress/blob/60c26865f94e302f18548564324f78f2b2bbad5f/wp-admin/includes/post.php#L579

There's no filter currently..

sc0ttkclark:
We could:

```php
[...]

PR #5622, opened 2020-03-17
Fix dbx_post_advanced deprecated notices and error on page

sc0ttkclark:
## Description
Admin area shows deprecated notices when using Pods templates/helpers/pages components. Also sometimes bad pod configurations can fatal.

## ChangeLog
Fixed: Replaced hook used for meta boxes in admin for custom component integrations to avoid deprecated notices from WordPress.
Fixed: Prevented potential fatal errors on Pod Template editor screen when pod configuration is broken.


JoryHogeveen:
@sc0ttkclark  Might be related: #2823

JoryHogeveen:
And @sc0ttkclark should this PR include a deprecated/removed notice in the changelog for `dbx_post_advanced`?

sc0ttkclark:
@JoryHogeveen WP itself deprecated the hook so we're just changing where we're hooking in for those parts. This does not change Pods itself, it just fixes usages by the Pods components editors.

Issue #5600, opened 2020-02-23
Default sessions to off


sc0ttkclark:
We should default the sessions to off for Pods. See #5626

JoryHogeveen:
Open for discussion (see PR #5626 )

Not sure whether a setting is required for this.
The `on` feature seems a bit redundant since `auto` would cover all situations where we need sessions.
Any thoughts are welcome.

sc0ttkclark:
Fixed via #5626

PR #5626, opened 2020-03-19
Disable sessions by default

JoryHogeveen:
Fixes: #5600
Fixes: #5769 

We should default the sessions to off for Pods. This would involve a few things:

## ChangeLog
Fixed - We now default sessions to being off by default, to turn them back on again go to the Pods Admin > Settings screen to turn them on. Sessions are used to secure public Pods forms through `[pods form]` or `Pods::form()` when someone is not logged in.

## Checklist:
- [x] If `PODS_SESSION_AUTO_START` is not defined, it should assume false (current default: true)
- [x] Introduce notice text for public forms (when user not logged in) that they are required to log in (with login link) to access the form (if we determine the session will not be started)
- [x] Introduce new Pods setting to turn on public forms sessions (auto | off | on)
- [x] Auto (default): Disable sessions, if `Pods::form()`is called when `! is_user_logged_in()`, auto set option to `on` so future loads have sessions on
- [x] On: Enable sessions
- [x] Off: Disable sessions, if `Pods::form()`is called when `! is_user_logged_in()`, show notice
- [x] Validate on installations where the PHP Session extension isn't installed (reported on Slack)
- [x] My code is tested.


JoryHogeveen:
@sc0ttkclark 

> Introduce notice text for public forms (when user not logged in) that they are required to log in (with login link) to access the form (if we determine the session will not be started)

Why are sessions required for public forms?

sc0ttkclark:
Since with Pods, you can use our shortcode, PHP, or block (2.8+) to output a form with a varied number of fields entirely dependent on your use case -- we use sessions when people are logged out. The intention is not to use sessions if they are logged in, but that's tricky to have a fallback on when we don't have access to that at the time we set sessions up I think << Worth double checking!

Here's the use case we are protecting against.

As a developer
who has Pods installed
and has a Pod set up with fields

I want to allow people to submit a form
and they do not have to be logged in
and if they are logged in the fields need to be different
and if they are logged in they should be able to edit specific records

So that they can securely add and save details
and those details get saved to a Pod item
and they cannot modify or get around the fields I define
and they cannot add or edit for any pod without them encountering security checks

sc0ttkclark (approved):

Issue #5675, opened 2020-04-10
Pods @post_date doesn't honour WP's date setting (and is also not user-configurable)


patrikhuber:
Hi,

In WP's Settings => General, my WP date format is set to `j F Y`. This works fine everywhere on the website.

However when adding a Pod and outputting `{@post_date}`, it shows (for example) `2020-11-27 11:00`.

Users have noted this in 2014 and following - here an old Google Cache link to the old Pods support forum: https://webcache.googleusercontent.com/search?q=cache:v17EuboNpB4J:https://pods.io/forums/topic/date-formatting-using-magic-tags/+&cd=1&hl=en&ct=clnk&gl=uk

This is **not** a duplicate of #5038. #5038 is about custom fields - on these, the date is user-configurable in the Pods custom fields settings.
However the `{@post_date}` fields is not user-configurable.

The workaround in above link is to add a PHP function that converts the date. However adding a PHP function is an obstacle to many. It would be great if:
1) Pods honoured WP's date setting (bugfix)
2) There was a Pods way of setting the date output format without using PHP - ideally this would even be at the time of outputting it so that I could output the date of the same Pods type *with* the time in one Pods template and *without* the time in another Pods template. (feature request)

patrikhuber:
I think the best solution to all this, if technically feasible, would be if we could just access the individual elements of `{@post_date}`'s date. It would be fantastic to be able to do, for example, `{@post_date.month}/{@post_date.year}`. Some people in above forum link also suggested this and it would be a very intuitive way and solve all problems.

JoryHogeveen:
Hi @patrikhuber 

Please see #5676 as a concept for such a feature.

patrikhuber:
Hi @JoryHogeveen,

Cool, that's awesome! I'll leave a comment there. :-)

Issue #4525, opened 2017-10-19
Post_date Magic Tag displaying in default yyyy-mm-dd hh:mm:ss instead of WordPress Default


PodsBot:
Post_date Magic Tag displaying in default yyyy-mm-dd hh:mm:ss instead of WordPress Default submitted via Slack by jimtrue

jimtrue:
`{@post_date}` in a template is defaulting to the default field display instead of the WordPress, Settings, General Date format

jamesgol:
This really comes down to Pods->display() not returning a properly formatted date, which is ultimately the responsibility of Pods->field() at the current time.  

So at this time I wouldn't be adding any more code to that function until it gets broken up with the PodsObject type stuff.  

I created a phpunit test for this and I'll create a PR for it but the tests will be failing until the issue gets resolved.

pglewis:
@guidobras: can you open a new issue for this in case they're from separate things?

guidobras:
created #4925
Thanks!

quasel:
Yep still an issue ;)

PR #4530, opened 2017-10-25
Properly format post_date for display

jamesgol:
This PR is incomplete but adds a branch with a test that we can use when we finally can fix it properly.

PR #5676, opened 2020-04-10
Fix default post_date format + add date format magic tag options

JoryHogeveen:
Fixes: #5675
Fixes: #4525

## Description

THIS PR IS NOT DONE.

The relates issue noted that using `{@post_date}` always displays the mysql format (Y-m-d h:i).
This PR sets this to the default formatting settings from WordPress.

This PR also adds a format param to the magic tag: `{@post_date._format.Y-m-d}`, `{@post_date._format.Y m d}` and even `{@post_date._format.Y.m.d}` (note the dots).
Currently this PR only supports this for core-field dates (`post_date`). But might be a cool feature for other date fields as well.

**Would like to have feedback/ideas on how to best implement this.**

## Checklist:
- [ ] My code is tested.
- [ ] My code includes PHP Unit Tests (if applicable)


patrikhuber:
Hi @JoryHogeveen,

Cool, this would be great, thank you for moving this forward! Big thumbs up from me on the feature, and by just googling "wordpress pods date format", there are dozens of users looking for something like it.

As a programmer myself, I would perhaps suggest to put the format string in parentheses? I would say that it gets a bit dodgy with whitespaces otherwise, like `{@post_date._format.Y m d}`. For example `{@post_date._format("Y m d")}`. But since everything is inside curly braces `{...}`, even your whitespace example isn't too unclear.

JoryHogeveen:
@patrikhuber 
Thanks. And about the parentheses, that isn't compatible with the Pods magic tag formats, so even if it would be enclosed with quotes, there won't be any function-like formatting here.

classes/Pods.php
```diff
@@ -1143,9 +1137,52 @@ public function field( $name, $single = null, $raw = false ) {
...
+									if ( $field ) {
+										$format = implode( '.', $field );
+										$value  = date_i18n( $format, strtotime( $value ) );
+
+										$field_data['options']['datetime_type']          = 'custom';
+										$field_data['options']['datetime_time_type']     = 'custom';
+										$field_data['options']['datetime_format_custom'] = $format;
+									} else {
+										$field_data['options']['datetime_type']          = 'wp';
+										$field_data['options']['datetime_time_type']     = 'wp';
```
sc0ttkclark, line 1161:
```suggestion
										$field_data['options']['datetime_type']      = 'wp';
										$field_data['options']['datetime_time_type'] = 'wp';
```

classes/Pods.php
```diff
@@ -1143,9 +1137,52 @@ public function field( $name, $single = null, $raw = false ) {
...
+				foreach ( $date_fields as $date_field ) {
+					if (
+						0 === strpos( $params->name, $date_field . '.' )
+					) {
+						$field = explode( '.', $params->name );
+
+						if ( 1 < count( $field ) ) {
+							array_shift( $field ); // Remove field name.
+							$type = array_shift( $field );
+							switch ( $type ) {
```
sc0ttkclark, line 1150:
```suggestion

							switch ( $type ) {
```

classes/Pods.php
```diff
@@ -1143,9 +1137,52 @@ public function field( $name, $single = null, $raw = false ) {
...
 				} //end foreach
+
+				foreach ( $date_fields as $date_field ) {
+					if (
+						0 === strpos( $params->name, $date_field . '.' )
+					) {
+						$field = explode( '.', $params->name );
+
+						if ( 1 < count( $field ) ) {
+							array_shift( $field ); // Remove field name.
```
sc0ttkclark, line 1148:
```suggestion
							// Remove field name.
							array_shift( $field );

```

sc0ttkclark:
Looks good, just needs tests

JoryHogeveen:
@sc0ttkclark 
For tests we might want to merge https://github.com/pods-framework/pods/pull/4530 first. Maybe merge it in this PR?

classes/Pods.php
```diff
@@ -1676,6 +1722,7 @@ public function field( $name, $single = null, $raw = false ) {
 									foreach ( $data as $item_id => $item ) {
 										// $field is 123x123, needs to be _src.123x123
 										$full_field = implode( '.', array_splice( $params->traverse, $key ) );
+										var_dump( $full_field );
```
sc0ttkclark, line 1725:
```suggestion
```

JoryHogeveen:
@sc0ttkclark
I found a little glitch in our usage of dates.
We set the date fields to be `datetime`, which is logical. But, as the tests from @jamesgol compared the `post_date` tag with `get_the_date()` this will not match since `get_the_date()` only returns the date, not the time.

I can think of various options to add other magic tags:
- `{@post_date}` > Only date in default WP format
- `{@post_time}` > Only time in default WP format
- `{@post_datetime}` > Date & time in default WP format
etc.

Adding these tags would mean I'd need to extend the WP object fields for posts in PodsAPI.

But we can also leave it like this since it now supports custom formats: `{@post_date._format.##`

What do you think would be best?

sc0ttkclark:
We probably need to choose something different because post_date (data-wise) is really the post date+time. Maybe something like the_date or the_datetime or the_time?

JoryHogeveen:
@sc0ttkclark 
Sounds good to me as well. So with this `pods_date` remains unchanged?
I can refactor this PR to do that. But we'll need to create docs for this so people know `post_date` is the actual database value and `the_##` is the display value.

sc0ttkclark:
@JoryHogeveen Yes, I think you meant `post_date` not `pods_date`, but yes that value remains unchanged. `the_*` variations would do display.

sc0ttkclark:
All of the logic will need to be reworked here in the context of the new field value mapping added in Pods 2.8. Also overriding options in the field config means that it modifies the object and that could be a problem for tests etc.

Overall, I think we need to look at a better solution for modifiers outside of the regular field mapping in 2.8, maybe something that can run after the value itself is retrieved and can change the resulting value/display.

dmkjr:
Any movement on this?  Looking for a method to only include the Month/Year in our PODS template.

Issue #5699, opened 2020-04-29
`edit_category_form` is deprecated since WP 3.0


JoryHogeveen:
**Describe the bug**
https://wordpress.org/support/topic/error-trying-to-access-categories-labels/

Deprecated actions:
- `edit_category_form` => `{$taxonomy}_add_form`
- `edit_link_category_form` => `{$taxonomy}_edit_form`
- `edit_tag_form` => `{$taxonomy}_edit_form`
- `add_category_form` => `{$taxonomy}_add_form`

PR #5700, opened 2020-04-29
Fix deprecated tax form actions

JoryHogeveen:
Fixes: #5699

## How Has This Been Tested?

If you edit/add a taxonomy relationship the modal popup it should automatically close after edit/add.

## Checklist:
- [x] My code is tested.


JoryHogeveen:
@sc0ttkclark From what I can tell looping through all taxonomies is the only way to do this. If you have another idea then let me know!

Issue #4815, opened 2018-03-30
unable to remove admin menu items after using pods_is_admin


SethTurin:
I've killed a few days on this now. Please help!

## Issue Overview
I'm using Wordpress multisite. After applying the suggestion from #1629, the menu shows up as expected. However, I'd like to remove admin abilities to change components and templates, since this presents a security risk. 

my first step is to remove the menu items with 
```
    remove_submenu_page('pods', 'pods-components');
    remove_submenu_page('pods', 'edit.php?post_type=_pods_template');
```

This code works fine if I'm a super admin and I don't have the pods_is_admin code running. But as soon as I add the pods_is_admin code (the fix from #1629), the menu items reappear, and I haven't found any way to get rid of them.

Of course, after I get rid of them, I also need to restrict access to the pages themselves (since removing the menu items doesn't do this). The security risk still exists because the user can manually type in the url. Interestingly, despite the menu item for templates being shown, by default, the site admin is unable to access the templates page. I'm not sure if this is expected behavior, since the menu item is shown, but it is the desired behavior, assuming I can kill the menu item. 

That just leaves the components page to remove. I looked for capabilities I could set to remove this, but didn't find any. Is there any other simple way to remove site admin access to this page?

## Expected Behavior
1. menu items should be removed when using remove_submenu_page on pods submenus
2. need a way to restrict access to components page once the menu item is removed

## Current Behavior
1. menu items are shown even after being removed
2. admins already do not have access to templates page, but they do have access to components page

## Steps to Reproduce (for bugs)
For the menu item removal behavior, you should be able to reproduce it with the code below. This is on a multisite installation, and the code is placed in the functions.php of the default theme.
```
add_filter( 'pods_is_admin', 'slug_let_admins_admin', 10, 3 );
function slug_let_admins_admin( $has_access, $cap, $capability ) {
	$capability = 'admin';
	return $capability;
}
function custom_remove_admin_menus (){
    remove_submenu_page('pods', 'pods-components'); //pods components
    remove_submenu_page('pods', 'edit.php?post_type=_pods_template'); //pods templates

    return true;
}
add_action('admin_menu', 'custom_remove_admin_menus', 999);
```

## Related Issues and/or PRs
#1629

## WordPress Environment
<details>
```
Debug Information
WordPress Version: 4.9.4

PHP Version: 7.0.28

MySQL Version: 5.6.36

Server Software: Apache

[...]

Issue #3985, opened 2017-02-03
Unable to disable Pods Admin access for roles that have edit_users capability


Digitale-Substanz:
Hi,

I have a custom user role (_my_role_) defined on plugin initialisation. This role has no 'Pods' capabilities. If I double-check the cap-settings for this role, all 'Pods' capabilities are unchecked as they should be.

```
register_activation_hook(__FILE__, function(){
  add_role( 'my_role', 'My Role', array(
    'delete_others_pages'			=> true,
    'delete_pages'						=> true,
    'delete_private_pages'		=> true,
    'delete_published_pages'	=> true,
    'edit_others_pages'				=> true,
    'edit_pages'							=> true,
    'edit_private_pages'			=> true,
    'edit_published_pages'		=> true,
    'read_private_pages'			=> true,
    'publish_pages'						=> true,
    'delete_others_posts'			=> true,
    'delete_posts'						=> true,
    'delete_private_posts'		=> true,
    'delete_published_posts'	=> true,
    'edit_others_posts'				=> true,
    'edit_posts'							=> true,
    'edit_private_posts'			=> true,
    'edit_published_posts'		=> true,
    'read_private_posts'			=> true,
    'publish_posts'						=> true,
    'add_users'				 => true,
    'create_users'		 => true,
    'edit_users'			 => true,
    'delete_users'		 => true,
    'remove_users'  	 => true,
    'list_users'			 => true,
    'read' 						 => true,
    'level_0' 				 => true,
    'access_userarea'  => true,
    'use_sxd_tools'		 => true,
    'delete_others_tribe_events'	=> true,
    'delete_tribe_events'					=> true,
    'edit_others_tribe_events'		=> true,
    'publish_tribe_events'			  => true,
    'read_private_tribe_events'		=> true,
    'delete_private_tribe_events' => true,
    'delete_published_tribe_events' => true,
    'edit_private_tribe_events'		=> true,
    'edit_published_tribe_events'	=> true,
    'edit_tribe_events'						=> true,
  });
```

However, the Pods Admin menu including all sub menu items appear in the WP-Admin of any user with only the 'my_role' role. Every feature of Pods can be accessed by this role without proper capabilites.

Thanks.

sc0ttkclark:
Any user with the `edit_users` capability means they can change anyone's role. Anyone with that capability is specifically treated as a super admin within WordPress, and is what we fallback to when our capabilities are not present.

JoryHogeveen:
[...]

Issue #2311, opened 2014-08-05
Capabilities: delete_users


adleviton:
It looks like access to pods is solely based on one wordpress capability: delete_users.

What should I do if I want a client to be able to delete users and NOT have access to pods?

quasel:
from @JoryHogeveen #3625

>  Currently the admin capabilities check for the delete_users capability in pods_is_admin() (includes / general.php).
> I've would like to see this changed into manage_options or administrator since Pods is a options page...
> 
> An other idea is to just use the is_super_admin() function since this function also works for non-multisite installations.

is this still an issue or just a doc's issue?

quasel:
and from @sc0ttkclark  #3985 

> There's a filter where you can override all access checks via `pods_is_admin()`, the filter is named `pods_is_admin`.
> 
> For single site installs, you can disable access from delete_users capability by using the filter:

```php
add_filter( 'pods_admin_capabilities', '__return_empty_array' );
```

PR #5712, opened 2020-05-13
Only check `delete_users` for single installations

JoryHogeveen:
Related: #2311
Related: #3985
Related: #4815

## Description

This PR will enhance the network check for Pods access.
Currently Pods validates the `delete_users`, just like WordPress core.
However, this is not entirely correct.

For multisite installations this check isn't valid as only actual super admins should always have access.
This PR only adds the `delete_users` cap if it's a single installation. Multisites do not check this cap.

Doing so improves the way we validate capabilities and makes it behave more similar to WP core.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -328,6 +328,7 @@ function pods_is_admin( $cap = null ) {
 		}
 
 		$pods_admin_capabilities = array();
+
 		if ( ! is_multisite() ) {
 			// Default is_super_admin() checks against this capability.
 			$pods_admin_capabilities[] = 'delete_users';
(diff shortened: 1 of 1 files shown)


includes/general.php
```diff
@@ -322,17 +322,19 @@ function pods_is_debug_display() {
...
-		);
-
-		$pods_admin_capabilities = apply_filters( 'pods_admin_capabilities', $pods_admin_capabilities, $cap );
 
 		if ( is_multisite() && is_super_admin() ) {
 			return apply_filters( 'pods_is_admin', true, $cap, '_super_admin' );
 		}
 
+		$pods_admin_capabilities = array();
+		if ( ! is_multisite() ) {
```
sc0ttkclark, line 331:
```suggestion

		if ( ! is_multisite() ) {
```

Issue #5705, opened 2020-05-05
2.7.20: Pod field editor REST API tab should show "Depth" field only for "Response Type: Full"


lkraav:
**Describe the bug**
It turns out "Depth" is irrelevant for any other "Response Type" like "ID", or "Name", and is probably confusing for any user that might be wondering whether it should have an effect or not.

![image](https://user-images.githubusercontent.com/147228/81107887-89180680-8f20-11ea-858d-08efa9be4c99.png)

Confusing :arrow_up: 

**To Reproduce**
1. Set any field REST API response type to "ID"
2. Modify "Depth" value: 0, 1, or 2
3. Make REST API requests for all Depth values
4. Observe no difference in REST API responses

**Expected behavior**
"Depth" field should only be shown for relevant "Response Type".

## Pods Version
2.7.20

## WordPress Environment
5.4.1

JoryHogeveen:
Good find, thanks!

PR #5714, opened 2020-05-14
REST field options: Only display depth for array response type

JoryHogeveen:
Fixes: #5705 

Hide the "Depth" option for single response types.

## Checklist:
- [x] My code is tested.


lkraav:
Wow, wasn't even aware y'all already have a dependency system built-in :tada:

PR #5802, opened 2020-08-11
Shortcode field template

### LINKED ISSUE #5801: Shortcode enhancement: Use template on relationship field
Currently if you add a `field` parameter to the shortcode you just output that field value.
It would be awesome to have the option to use the `field` and `template` parameters in combination for relationship fields.
Instead of a complicated `where` kind of shortcode you can just call the related field and it will output the template based archive HTML.

JoryHogeveen:
Fixes #5801 

## Description

Currently if you add a `field` parameter to the shortcode you just output that field value.
It would be awesome to have the option to use the `field` and `template` parameters in combination for relationship fields.
Instead of a complicated `where` kind of shortcode you can just call the related field and it will output the template based archive HTML.

## Usage examples

This will output the template for each relationship context if used within the loop.
`[pods field="relationship_field" template="Template Name"]`

This will output the template for each relationship based on the provided Pod.
`[pods name="pod" id="#" field="relationship_field" template="Template Name"]`

Nested shortcode content is also supported:
`[pods field="relationship_field"] TEMPLATE CONTENT [/pods]`
`[pods name="pod" id="#" field="relationship_field"] TEMPLATE CONTENT [/pods]`

## Checklist:
- [x] My code is tested.
- [ ] My code includes PHP Unit Tests (if applicable)


includes/general.php
```diff
@@ -1052,7 +1052,16 @@ function pods_shortcode_run( $tags, $content = null ) {
 
 		return $pod->form( $tags['fields'], $tags['label'], $tags['thank_you'] );
 	} elseif ( ! empty( $tags['field'] ) ) {
-		if ( empty( $tags['helper'] ) ) {
+		if ( ! empty( $tags['template'] ) ) {
+			$return  = '';
```
tr1b0t, line 1056: :warning: Equals sign not aligned correctly; expected 1 space but found 2 spaces

<sub>hash: 3ad5a47bde8150a0b6c85eb02e86e30e</sub>

includes/general.php
```diff
@@ -1052,7 +1052,16 @@ function pods_shortcode_run( $tags, $content = null ) {
...
+			// @todo New Pods query based on relationship field.
```
tr1b0t, line 1057: :warning: This comment is 47% valid code; is this commented out code?

<sub>hash: 5d702fc7856cce018653f3da4bd128dc</sub>

includes/general.php
```diff
@@ -1052,7 +1052,16 @@ function pods_shortcode_run( $tags, $content = null ) {
...
+			//$field = $pod->fields( $tags['field'] );
```
tr1b0t, line 1058: :no_entry_sign: No space found before comment text; expected "// $field = $pod->fields( $tags['field'] );" but found "//$field = $pod->fields( $tags['field'] );"

<sub>hash: e4ecea80e19af387db61041dc80391ac</sub>

includes/general.php
```diff
@@ -1052,7 +1052,16 @@ function pods_shortcode_run( $tags, $content = null ) {
...
+			//$rel_pod = $field[ $tags['field'] ];
```
tr1b0t, line 1059: :no_entry_sign: No space found before comment text; expected "// $rel_pod = $field[ $tags['field'] ];" but found "//$rel_pod = $field[ $tags['field'] ];"

<sub>hash: 49ae555ba7582b56ba426afcd320d233</sub>

includes/general.php
```diff
@@ -1052,7 +1052,15 @@ function pods_shortcode_run( $tags, $content = null ) {
 
 		return $pod->form( $tags['fields'], $tags['label'], $tags['thank_you'] );
 	} elseif ( ! empty( $tags['field'] ) ) {
-		if ( empty( $tags['helper'] ) ) {
+		if ( ! empty( $tags['template'] ) ) {
+			// @todo Single Pods query based on relationship field.
+			$return  = '';
+			$related = $pod->field( $tags['field'], array( 'output' => 'pods' ) );
```
tr1b0t, line 1058: :no_entry_sign: Short array syntax must be used to define arrays

<sub>hash: 29d054e3102146a1da4b5c0fc77b220f</sub>

tr1b0t (commented):
:no_entry_sign: 1 error(s)

includes/general.php
JoryHogeveen: @sc0ttkclark Is this already a requirement for 2.7.23?

sc0ttkclark: Sorry, this is for 2.8, you can ignore

sc0ttkclark, line 1058: This would only work on a related object that is also a pod itself.

But if that was the case, you'd probably want to use `'output' => 'find'` instead and just do:

```php
if ( $related instanceof Pods ) {
	$return = $related->template( $tags['template'], $content );
}
```

sc0ttkclark (approved):
I really like this idea, I don't think it's problematic as long as there's some checks to make sure there are no unforeseen errors when they use a wrong field name or non-pick field.

This makes a lot of sense to me, this could also let us expand the field widget/block for that use case too.

includes/general.php
sc0ttkclark: I want to note that calling it like this won't need a foreach or while, since template does that when it detects it's in a list context (without passing a direct ID to the Pods object).

JoryHogeveen: True, but in this case (since we use `field`) we get a list of separate Pods objects.

JoryHogeveen:
I've added two more changes the prevent possible errors.

includes/general.php
```diff
@@ -1052,7 +1052,17 @@ function pods_shortcode_run( $tags, $content = null ) {
 
 		return $pod->form( $tags['fields'], $tags['label'], $tags['thank_you'] );
 	} elseif ( ! empty( $tags['field'] ) ) {
-		if ( empty( $tags['helper'] ) ) {
+		if ( ! empty( $tags['template'] ) ) {
+			// @todo Single Pods query based on relationship field.
+			$return  = '';
+			$related = (array) $pod->field( $tags['field'], array( 'output' => 'pods', 'single' => false ) );
```
tr1b0t, line 1058: * :no_entry_sign: Short array syntax must be used to define arrays
* :no_entry_sign: When a multi-item array uses associative keys, each value should start on a new line.

<sub>hash: 18997f6c3ffebb35482ff2e9ad51df3d</sub>

tr1b0t (commented):
:no_entry_sign: 2 error(s)

includes/general.php
sc0ttkclark: If we use `'output' => 'find'` it would be more performant. Multiple Pods objects ends up being more intensive.

sc0ttkclark (changes requested):
Let's look at using `'output' => 'find'`

includes/general.php
```diff
@@ -1052,7 +1052,13 @@ function pods_shortcode_run( $tags, $content = null ) {
 
 		return $pod->form( $tags['fields'], $tags['label'], $tags['thank_you'] );
 	} elseif ( ! empty( $tags['field'] ) ) {
-		if ( empty( $tags['helper'] ) ) {
+		if ( ! empty( $tags['template'] ) ) {
+			$return  = '';
+			$related = $pod->field( $tags['field'], array( 'output' => 'find' ) );
```
tr1b0t, line 1057: :no_entry_sign: Short array syntax must be used to define arrays

<sub>hash: 1d0149ad9aad33436ad8ca94d338441b</sub>

tr1b0t (commented):
:no_entry_sign: 1 error(s)

includes/general.php
```diff
@@ -1052,7 +1052,14 @@ function pods_shortcode_run( $tags, $content = null ) {
 
 		return $pod->form( $tags['fields'], $tags['label'], $tags['thank_you'] );
 	} elseif ( ! empty( $tags['field'] ) ) {
-		if ( empty( $tags['helper'] ) ) {
+		$template = ! empty( $tags['template'] ) ? $tags['template'] : $content;
+		if ( $template ) {
+			$return  = '';
+			$related = $pod->field( $tags['field'], array( 'output' => 'find' ) );
```
tr1b0t, line 1058: :no_entry_sign: Short array syntax must be used to define arrays

<sub>hash: 29d054e3102146a1da4b5c0fc77b220f</sub>

tr1b0t (commented):
:no_entry_sign: 1 error(s)

JoryHogeveen:
I've refactored it to make use of the `find` output.
Another addition is that the PR now also supports template content in a nested shortcode!

includes/general.php
```diff
@@ -1052,7 +1052,14 @@ function pods_shortcode_run( $tags, $content = null ) {
...
+			if ( $related instanceof Pods && $related->valid() ) {
+				$return .= $related->template( $tags['template'], $template );
+			}
```
sc0ttkclark, line 1061:
```suggestion

		if ( $template ) {
			$return  = '';
			$related = $pod->field( $tags['field'], array( 'output' => 'find' ) );

			if ( $related instanceof Pods && $related->valid() ) {
				$return .= $related->template( $tags['template'], $template );
			}
```

includes/general.php
```diff
@@ -1052,7 +1052,16 @@ function pods_shortcode_run( $tags, $content = null ) {
 
 		return $pod->form( $tags['fields'], $tags['label'], $tags['thank_you'] );
 	} elseif ( ! empty( $tags['field'] ) ) {
-		if ( empty( $tags['helper'] ) ) {
+		$template = ! empty( $tags['template'] ) ? $tags['template'] : $content;
+
+		if ( $template ) {
+			$return  = '';
+			$related = $pod->field( $tags['field'], array( 'output' => 'find' ) );
```
tr1b0t, line 1059: :no_entry_sign: Short array syntax must be used to define arrays

<sub>hash: cde034aecbcba999f530fdb5a82c9f75</sub>

tr1b0t (commented):
:no_entry_sign: 1 error(s)

includes/general.php
```diff
@@ -1052,7 +1052,16 @@ function pods_shortcode_run( $tags, $content = null ) {
...
+
+			if ( $related instanceof Pods && $related->valid() ) {
+				$return .= $related->template( $tags['template'], $template );
```
sc0ttkclark, line 1062: @JoryHogeveen Can you describe the usage here, if `$tags['template']` is set then it sets `$template` too.

sc0ttkclark (commented):
Q

includes/general.php
JoryHogeveen: Good one, that isn't correct.. Weird that it worked locally though..

sc0ttkclark (approved):

PR #5825, opened 2020-09-02
Pods 2.8: WP doesn't use border radius

JoryHogeveen:
The 9px border radios isn't in line with the WP and Pods UI.

### DIFF
--- a/ui/js/dfv/src/admin/edit-pod/main-tabs/field-groups.scss
+++ b/ui/js/dfv/src/admin/edit-pod/main-tabs/field-groups.scss
@@ -9,15 +9,14 @@
 
 .pods-field-group-wrapper {
 	margin-bottom: 10px;
-	border-radius: $border-radius;
 	border: 1px solid $color--light-grey;
 	background-color: #fff;
 	position: relative;
-    transition: border 200ms ease-in-out;
+	transition: border 200ms ease-in-out;
 
 	&.pods-unsaved-data  {
-        border-left-color: #95BF3B;
-        border-left-width: 5px;
+		border-left-color: #95BF3B;
+		border-left-width: 5px;
 	}
 
 	&>.pods-field_wrapper {
@@ -71,7 +70,6 @@
 	border: 2px dashed $color--light-grey;
 	background: none;
 	width: 100%;
-	border-radius: $border-radius;
 	padding: .7em 1em;
 	transition: 300ms ease background-color;
 }
--- a/ui/js/dfv/src/admin/edit-pod/main-tabs/field-list-item.scss
+++ b/ui/js/dfv/src/admin/edit-pod/main-tabs/field-list-item.scss
@@ -147,7 +147,6 @@
 
 .pods-field_wrapper-items {
 	border: 1px solid $color--light-grey;
-	border-radius: $border-radius;
 	margin: 1em 0;
 	overflow: hidden;
 }
@@ -180,10 +179,10 @@
 }
 
 .pods-field_button.pods-field_delete {
-    color: $color--red;
+	color: $color--red;
 
-    &:hover {
-        color: $color--red-dark;
+	&:hover {
+		color: $color--red-dark;
 	}
 
 	&::after {
--- a/ui/js/dfv/src/admin/edit-pod/_edit-pod-variables.scss
+++ b/ui/js/dfv/src/admin/edit-pod/_edit-pod-variables.scss
@@ -1,4 +1,4 @@
-$border-radius: 0;
+$border-radius: 2px;
 
 $color--grey: #82878c;
 $color--light-grey: #ccd0d4;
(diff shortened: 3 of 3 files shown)


sc0ttkclark:
Those border radius usages were intentional for our group design aesthetic.

JoryHogeveen:
I understand that they were added intentionally but IMO it's better to keep the UI in line with WP core

nicdford:
Do we need to take a larger look at the entire UI and make sure it's all consistent? My understanding was that we were going to work through each of the different UI elements to make it matches the new "look and feel" (see figma https://www.figma.com/file/KTj2ZgskEcxxkwwaDI6S9A/Pods-Admin-Edit-Pod-Groups?node-id=0%3A1). I agree that the border-radius doesn't really make sense in the context of the old UI, but I imagined we'll be using more border-radius, much like Gutenberg has to provide specific UI details to help the user understand where they are at in the flow the editing experience. I'm always open to design feedback as well, I did a bit of digging and I don't see much border-radius usage on container elements in the new Gutenberg "look", so maybe we just limit it to input fields and buttons?

sc0ttkclark:
We wanted to differentiate the group containers from everything else, I think that makes sense for what we're trying to do there. I'm open to removing them but I'd like that to be something we think about.

JoryHogeveen:
> I'm always open to design feedback as well, I did a bit of digging and I don't see much border-radius usage on container elements in the new Gutenberg "look", so maybe we just limit it to input fields and buttons?

That was my conclusion as well. Borders are only applied to form elements.

> We wanted to differentiate the group containers from everything else, I think that makes sense for what we're trying to do there.

Understandable but I don't think that rounded borders will enhance differentiation. The regular WP UI is (imo) already sufficient.

Personally I'd just replicate the default metabox UI from the latest WP update.
Removing the border radius already does that for the most part, though there are some more elements I could fix if you want.

JoryHogeveen:
Proposal:

image

nicdford:
If we do decide to circle back to the design (I'd like to make sure @sc0ttkclark is ok with that first), I'd like to update the UI to better match the WP 5.5 Block UI updates they recently implemented. These specifically address the ordering controls and following how they handle their nested columns display to apply to our fields, would make sense to me as well. 

see: 
image

If you look at the [Gutenberg Phase 2 design updates](https://make.wordpress.org/design/2020/08/28/gutenberg-phase-2-friday-design-update-60/), it indicates many of the WP UI is being overhauled to match the Gutenberg block editor experience, so it'd be a pretty safe bet for us to do this as well.

sc0ttkclark:
@nicdford can you post a wireframe example of what you're talking about? I want to be careful about time spent on this because we're so far into things and we're already over budget for this feature.

nicdford:
The idea here is that we try to follow suit with what they're doing with blocks. I think its a solid direction, but I get the sense that the Gutenberg design hasn't fully reached maturity, so adopting some of this look could result in us trying to "keep up" with what they're doing aesthetically. Specifically looking at how the order/drag functionality looks, it has dramatically changed in recent updates. All in a good direction, but it's hard to decipher what might also change. 

image

After going down this road a bit, I'm leaning towards @JoryHogeveen recommendation, as it doesn't match much else in the Pods UI or WP UI at the moment. So removing the border-radius and moving forward with the design as-is otherwise. It would be nice to add sorting controls that don't require a mouse drag, as this has accessibility implications, but I don't think its a fatal flaw we need to address at the moment.

nicdford:
I went ahead and updated the border-radius variable to match wp core, but it's not currently being used by anything. It's there if we need it in the future though. I then removed the border-radius value entirely, instead of applying a zero value.

nicdford (approved):
Setting this to a 2px value to match the rest of WP Core.

ui/js/dfv/src/admin/edit-pod/main-tabs/field-groups.scss
```diff
@@ -9,15 +9,14 @@
...
 	background-color: #fff;
 	position: relative;
-    transition: border 200ms ease-in-out;
+	transition: border 200ms ease-in-out;
 
 	&.pods-unsaved-data  {
-        border-left-color: #95BF3B;
-        border-left-width: 5px;
+		border-left-color: #95BF3B;
+		border-left-width: 5px;
```
nicdford, line 19: Fixing space space misformatting

ui/js/dfv/src/admin/edit-pod/main-tabs/field-list-item.scss
```diff
@@ -180,10 +179,10 @@
 }
 
 .pods-field_button.pods-field_delete {
-    color: $color--red;
+	color: $color--red;
 
-    &:hover {
-        color: $color--red-dark;
+	&:hover {
+		color: $color--red-dark;
```
nicdford, line 185: Fixing white space misformatting

nicdford (approved):

sc0ttkclark (approved):

JoryHogeveen:
Awesome @nicdford !
I also think it would be even better to follow the Block UI since you are right, WP will patch that over to all parts of WP core UI.

@sc0ttkclark  How can we be best proceed in getting this into 2.8? Separate PR? Multiple smaller PR's?

sc0ttkclark:
@JoryHogeveen This is good, was able to merge it earlier. We can look at implementing the newer Block UI across our interfaces as some separate work maybe in 2.9 or 3.0

PR #5846, opened 2020-10-02
Prevent storing autoloaded transients in the DB

sc0ttkclark:
## Description

Many Pods transients have no expiration on them, which is fine in the case of using Pods Alternative Cache, but when stored in the database with large configs, this can be problematic as they are autoloaded in bulk.

Fixed: Prevent Pods from saving transients to the options table and having them autoloaded.

## PR Checklist

- [x] My code is tested.


JoryHogeveen (dismissed):

JoryHogeveen:
Looks good to me.
The only thing I'm wondering is whether this would actually improve loading.
I can imagine transients for registering Pods should always be loaded and thus would be faster if autoload was enabled since it doesn't require an extra query then.

sc0ttkclark:
The primary problem is autoload gets bloated and then has problems from MySQL limits. It's a trade off :(

JoryHogeveen:
Then maybe we should take a look at how to prevent bloating autoload but still add autoloading to transients that always need to be loaded.
This feels like an incomplete fix.

JoryHogeveen:
@sc0ttkclark  Do we have some form of overview of what transients are made throughout Pods?

jaycollier:
For what it's worth, I described the large autoloaded transient tables in our database in this ticket:

https://wordpress.org/support/topic/autoloading-database-tables/#post-13553380

jaycollier:
Hi!

Does the most recent note mean this is being considered for an upcoming release? I'm not experienced with GitHub to know what this means...

sc0ttkclark force-pushed the main branch from 220e6ac to a2aeac9 26 days ago

...  and I didn't see 5846 in the changelog.

Thanks for considering this.

-Jay

JoryHogeveen:
@sc0ttkclark 
Why was this punted to Backlog?

sc0ttkclark:
I'm not sure why this went to backlog, it should have just been moved to the next milestone. This PR needs further review on how we can separate things we do want auto-loaded versus ones that we don't.

jaycollier:
Thank you so very much for the autoload improvements in 2.8. I see that there was a lot of work involved, but it has reduced our autoload significantly! Very much appreciated.

PR #5847, opened 2020-10-06
Add Primary Branch header for GitHub Updater

afragen:
## Description

Since Pods' primary branch is now `main` in order for GitHub Updater to properly work the `Primary Branch` header needs to be added.

https://github.com/afragen/github-updater/wiki/Settings#primary-branch

Unfortunately users may need to use the branch switching to reinstall or use a webhook to force a branch switch. Otherwise they may not _see_ an update.

### DIFF
--- a/init.php
+++ b/init.php
@@ -8,6 +8,7 @@ Author: Pods Framework Team
 Author URI: https://pods.io/about/
 Text Domain: pods
 GitHub Plugin URI: https://github.com/pods-framework/pods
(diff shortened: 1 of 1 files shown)


JoryHogeveen (approved):

sc0ttkclark (approved):

sc0ttkclark:
Merging into the main branch but it will be included with 2.7.23 when it goes out

Issue #5677, opened 2020-04-11
Users with user_status not set to 0 are being excluded from relationships and lists using Pods code


ridinghoodmedia:
## Issue Overview

There appear to be a few bugs with bidirectional relationship, multiselect fields.

## Expected Behavior
I expect the bidirectional multiselect field and autocomplete field to work correctly.

## Current Behavior
I have a bidirectional relationship field Members (User single) and Organization (Organization multi).

From the organization page I can add users and it works and saves to the database. However the newly added users don’t show up in the *autocomplete* input field on reload, even though the database entries are made correctly and on the User page itself, it shows the organization.

If I use a multiselect field instead of an autocomplete, I can see the Users that belong to that Organization.

Also I noticed in the Advanced section in Organizations Pods settings under visibility it lists “Restrict access by role”. If I check a role and leave Administrator unchecked, I can still see that field information in the edit Organization page.

In addition, under “Additional Field Options” > “Limit list to role”, if I check two valid User roles I created “Assessor” and “Organization Admin”, if I visit the edit Organization page, I can add Assessors and Organization Admins, but when I reload the page, they don’t show up in the list. If I go back and deselect “Limit list to role”, the previously added Users magically appear.

The “Limit list to role” bug doesn’t exist with the checkbox AND multi select field but does exist with the List View AND Autocomplete fields. Strangely, with the Autocomplete field, it shows 2 of the 3 Users belonging to the Organization, whereas the List View shows none (and the checkbox and multi select show all 3 selected = correct).

## Steps to Reproduce (for bugs)
1. Create a bidirectional relationship between two custom post types, in my case Organization > "users" field (multiple) AND User > "organization" field (single)
2. Set the organization's users field to be a multiple selection autocomplete field
3. Create an organization in the backend, add multiple users to the organization
4. Reload the page - at this point, the "users" previously added to that field disappear

The other variations of this bug are described above.

## Possible Solution
It seems like whatever filter is determining the currently selected fields is off. Don't know enough about how this is done.

## WordPress Environment
<details>
Debug Information
WordPress Version: 5.4

PHP Version: 7.3.16

MySQL Version: 5.6.40

Server Software: Apache

Your User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.87 Safari/537.36

Session Save Path: /tmp

Session Save Path Exists: Yes

[...]

PR #5849, opened 2020-10-07
Do not limit users by user_status=0 anymore

sc0ttkclark:
## Description

WordPress core does not do this kind of limitation on WP User queries and the column has seen total disuse over the years since WordPress has been developed. Plugins are using this column for their own purposes which causes unexpected conflicts with how Pods relationships and lists using `Pods::find()` queries for users.

## How Has This Been Tested?

Tested on client site and confirmed as working.

* Fixed: Stop limiting users being queried using the deprecated `user_status` column in the WP Users table to prevent conflicts from other plugin usage of the disused column. #5677

## PR Checklist

- [x] My code is tested.

### DIFF
--- a/classes/PodsAPI.php
+++ b/classes/PodsAPI.php
@@ -9390,9 +9390,7 @@ class PodsAPI {
 			$info['meta_field_index'] = 'meta_key';
 			$info['meta_field_value'] = 'meta_value';
 
-			$info['where'] = array(
-				'user_status' => '`t`.`user_status` = 0'
-			);
+			$info['where'] = [];
 
 			$info['object_fields'] = $this->get_wp_object_fields( $object_type, $info['pod'] );
 		} elseif ( 'comment' === $object_type || 'comment' === pods_v( 'type', $info['pod'] ) ) {


classes/PodsAPI.php
```diff
@@ -9390,9 +9390,7 @@ public function get_table_info( $object_type, $object, $name = null, $pod = null
 			$info['meta_field_index'] = 'meta_key';
 			$info['meta_field_value'] = 'meta_value';
 
-			$info['where'] = array(
-				'user_status' => '`t`.`user_status` = 0'
-			);
+			$info['where'] = [];
```
tr1b0t, line 9393: :no_entry_sign: Short array syntax is not allowed

<sub>hash: db78b4a6c536a904c6f9655f989839e1</sub>

tr1b0t (commented):
:no_entry_sign: 1 error(s)

JoryHogeveen (approved):
Looks good to me, but might be better to do regular array syntax and add it to 2.7.23 maintenance release

sc0ttkclark:
@JoryHogeveen I'm primarily against removing this for 2.7.x because of the range of WP version support we have there. Pods 2.8 has more flexibility there since it's upping the required WP version by a lot, beyond the realm of any potential core usage of user_status

PR #5854, opened 2020-10-13
Fix PDF src in each loop.

### LINKED ISSUE #4040: Bug with WP 4.7.1 PDF Thumbnails {@_src} handlers goes through pods_image_url appends image suffix
I have a Pod setup with a File / Image / Video field.

When a user is adding content in the CPT, and they upload a PDF into that field, the URL generated shows a .jpg extension on the file instead of a .PDF (thus displaying the jpg thumbnail in the browser when clicked, rather than linking to the PDF)

Ex: the file I want to display is 
DMB-170119_17-000154-01-08.pdf

but the URL generated ends with 
DMB-170119_17-000154-01-08-pdf.jpg

This only happens for PDF’s that have been uploaded within WP 4.7, where a thumbnail has been generated. PDF’s uploaded prior to 4.7 (that still have the generic WP gray icon for a PDF file) have the correct .pdf file extension in the URL.

JoryHogeveen:
Fixes #4040 

## Description

PDF's tags like `{@id}` and `{@_src}` in each loops are converted to `{@image_attachment_url}` tags. This makes the PDF handled as an image.
I've chosen to fix this in the `field()` method instead of in frontier since a different solution would require an extra global tag like `pdf_attachment` or something which I didn't think was needed.
This makes sure if someone uses the `{@image_attachment_url}` tag the URL's get handled correctly.

Of course, if someone uses the `{@image_attachment}` tag this will always return an actual image (HTML).

This PR also removes duplicate code in `pods_image_url` for default value handling.

## PR Checklist

- [x] My code is tested.

### DIFF
--- a/tests/phpunit/includes/tests-media.php
+++ b/tests/phpunit/includes/tests-media.php
@@ -28,7 +28,7 @@ class Test_Media extends Pods_UnitTestCase {
 		// Default.
 		$this->assertNotEmpty( pods_image( null, 'thumbnail', self::$images[0] ) );
 
-		// Default (make sure it does not loop.
+		// Default (make sure it does not loop).
 		$this->assertEmpty( pods_image( null, 'thumbnail', 123456789 ) );
 	}
 
@@ -39,7 +39,7 @@ class Test_Media extends Pods_UnitTestCase {
 		// Default.
 		$this->assertNotEmpty( pods_image_url( null, 'thumbnail', self::$images[0] ) );
 
-		// Default (make sure it does not loop.
+		// Default (make sure it does not loop).
 		$this->assertEmpty( pods_image_url( null, 'thumbnail', 123456789 ) );
 	}
 
--- a/tests/phpunit/includes/tests-media.php
+++ b/tests/phpunit/includes/tests-media.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Pods_Unit_Tests;
+
+/**
+ * @group pods_acceptance_tests
+ */
+class Test_Media extends Pods_UnitTestCase {
+
+	/** @var array */
+	public static $images = array();
+
+	public function setUp() {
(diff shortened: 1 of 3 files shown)


sc0ttkclark (approved):

JoryHogeveen:
@sc0ttkclark 
Do you know of a way to add unit tests for this? I doubt PHPUnit supports Imageick and therefore the creation of PDF thumbnails...

sc0ttkclark:
Probably not much we can do there I think, at first. There's probably some stuff we'll unlock when we start using Codeception and some of the other tooling.

@JoryHogeveen Can you do a QA pass on this?

tests/phpunit/includes/tests-media.php
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	public function test_pods_image() {
+		// Simple fetch.
+		$this->assertNotEmpty( pods_image( self::$images[0] ) );
+
+		// Default.
+		$this->assertNotEmpty( pods_image( null, 'thumbnail', self::$images[0] ) );
+
+		// Default (make sure it does not loop.
```
sc0ttkclark, line 31:
```suggestion
		// Default (make sure it does not loop).
```

tests/phpunit/includes/tests-media.php
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	public function test_pods_image_url() {
+		// Simple fetch.
+		$this->assertNotEmpty( pods_image_url( self::$images[0] ) );
+
+		// Default.
+		$this->assertNotEmpty( pods_image_url( null, 'thumbnail', self::$images[0] ) );
+
+		// Default (make sure it does not loop.
```
sc0ttkclark, line 42:
```suggestion
		// Default (make sure it does not loop).
```

sc0ttkclark (approved):

PR #5533, opened 2019-11-14
Redo #5289: Enhance Pods auto-templates in general (especially Taxonomies and Comments)

JoryHogeveen:
From: #4768 & #5289 
Fixes: #3962 (Taxonomy auto templates)
Fixes: #5832 (Comment auto templates)

This PR has greatly changed the code for auto templates. Restructured and refactored to make more sense and be more aimed at future enhancements.

Major enhancement for this PR is that it allows auto templates for taxonomies, users & comments (eg. description or other locations).

**Text below is copied from previous PRs:**

<details>
> Via @gwhitney
> 
> I am submitting this pull request as suggested in the discussion to Issue #3962, where I requested that auto templates for taxonomy pods not be turned off. Let me start by saying that although I have considerable coding experience, I am relatively inexperienced with git workflows, so my apologies if I do not have the forking/branching structure correct, and if so, I am happy to close this pull request and create a new better one, with some guidance as to what should be changed.
> 
> Next, I should say that this pull request is organized as a series of successive recommendations regarding auto template handling for taxonomy types; you may decide to take almost any initial sequence of these commits and leave the remainder for further deliberation or reject the remainder.
> 
> Finally, let me explain the contents/motivation of each commit further, so you can better decide which to take, despite the risk of tl;dr. (Of course, I personally use and recommend all of these changes.) This list is in the order of the commits in the request:
> 
> 59d2e7e -- this is just the suggested patch from the discussion concerning #3962. Just by itself it makes auto templates with taxonomies more usable and understandable, in that it changes the default type of the pod created to run the template against from the pod type of the query that generated the post(s) to the post type of the post that is actually currently being displayed.
> 
> f201889 -- Another simple change that extends the usefulness of auto templates with taxonomies to those taxonomies which span multiple post types.. By including a magic code like {@post_type} in the auto template name, one can use different templates for different post types that might show up in the taxonomy archive.
> 
> 90a5f88 -- I noticed that 'the_content' was hard-coded into the temporary hook removal and re-installation in load_template, even though the actual filter used may have been set to something else, and that the priority of the hook was changing from 10.5 to 10. So instead, store the filters which have been hooked and their priorities in a member variable so that they can be maintained through load_template. Does not provide any new functionality, but is necessary for following commits, since they are just slightly more intensive in terms of setting/removing hooks.
> 
[...]

PR #5908, opened 2020-12-31
Fixes for Auto Template filter checking and usage

sc0ttkclark:
## Description

This ensures the right filters get hooked at the start of the setup.

* Fixed: Auto Template filters are set up correctly for anything other than the_content.

## PR Checklist

- [x] My code is tested.

### DIFF
--- a/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
+++ b/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
@@ -49,25 +49,12 @@ class Pods_Templates_Auto_Template_Front_End {
 	 * @since 2.6.6
 	 */
 	public function hook_content() {
-		/**
-		 * Allows plugin to append/replace the_excerpt.
-		 *
-		 * Default is false, set to true to enable.
-		 */
-		if ( ! defined( 'PFAT_USE_ON_EXCERPT' ) ) {
-			define( 'PFAT_USE_ON_EXCERPT', false );
-		}
-
 		$possible_pods = $this->auto_pods();
 
 		foreach ( $possible_pods as $pod_name => $pod ) {
 			$filter = $this->get_pod_filter( $pod_name );
 
 			$this->filtered_content[ $filter ] = 10.5;
-
-			if ( PFAT_USE_ON_EXCERPT && ! empty( $possible_pods[ $pod_name ]['type'] ) && 'post_type' === $possible_pods[ $pod_name ]['type'] ) {
-				$this->filtered_content['the_excerpt'] = 10;
-			}
 		}
 
 		$this->install_hooks();
@@ -301,9 +288,9 @@ class Pods_Templates_Auto_Template_Front_End {
 		if ( isset( $possible_pods[ $pod_name ] ) ) {
 			$this_pod = $possible_pods[ $pod_name ];
 
-			if ( in_the_loop() && ! is_singular() ) {
+			if ( is_archive() || is_post_type_archive() || is_tax() ) {
 				$filter = pods_v( 'archive_filter', $this_pod, $filter, true );
-			} else {
+			} elseif ( is_singular() || is_single() ) {
 				$filter = pods_v( 'single_filter', $this_pod, $filter, true );
 			}
 		}
@@ -453,7 +440,7 @@ class Pods_Templates_Auto_Template_Front_End {
 				}
 			}
 
-			if ( $is_single && ( ! $in_the_loop || is_singular() ) ) {
+			if ( $is_single && ( ! $in_the_loop || is_singular() || is_single() ) ) {
 				$type        = 'single';
 				$type_filter = 'single_filter';
 				$type_append = 'single_append';


components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -301,9 +288,9 @@ public function get_pod_filter( $pod_name = '' ) {
 		if ( isset( $possible_pods[ $pod_name ] ) ) {
 			$this_pod = $possible_pods[ $pod_name ];
 
-			if ( in_the_loop() && ! is_singular() ) {
+			if ( is_archive() || is_post_type_archive() || is_tax() ) {
 				$filter = pods_v( 'archive_filter', $this_pod, $filter, true );
-			} else {
+			} elseif ( is_singular() || is_single() ) {
```
JoryHogeveen, line 293: Why add an `elseif` here? The single filters might be needed on other locations.
Also, current fix doesn't check for `in_the_loop()` and it's very much possible to add archive-like loops in singular pages.

sc0ttkclark: Adjusted

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -49,25 +49,12 @@ public function __construct() {
...
-		if ( ! defined( 'PFAT_USE_ON_EXCERPT' ) ) {
-			define( 'PFAT_USE_ON_EXCERPT', false );
-		}
-
 		$possible_pods = $this->auto_pods();
 
 		foreach ( $possible_pods as $pod_name => $pod ) {
 			$filter = $this->get_pod_filter( $pod_name );
 
 			$this->filtered_content[ $filter ] = 10.5;
```
JoryHogeveen, line 57: Since the changes on getting the single or archive filter name it's possible that no filter is returned, this should be checked.
The alternative is to set a default like before this PR (single).

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -301,10 +288,10 @@ public function get_pod_filter( $pod_name = '' ) {
 		if ( isset( $possible_pods[ $pod_name ] ) ) {
 			$this_pod = $possible_pods[ $pod_name ];
 
-			if ( in_the_loop() && ! is_singular() ) {
-				$filter = pods_v( 'archive_filter', $this_pod, $filter, true );
-			} else {
+			if ( is_singular() || is_single() ) {
 				$filter = pods_v( 'single_filter', $this_pod, $filter, true );
+			} elseif ( is_archive() || is_post_type_archive() || is_tax() || in_the_loop() ) {
```
JoryHogeveen, line 293: This would still change the way it currently works dramatically.
By doing this, all taxonomies will always use the archive filter, even outside the loop.
This is inconsistent with the following overview:
https://github.com/pods-framework/pods/pull/5533#issuecomment-699581377

JoryHogeveen: Adding to this.. even inside a loop it could be that a single filter for posts should be used.
A lot of themes (including Genesis) use the loop even for singular pages (just the main content)

JoryHogeveen:
@sc0ttkclark 
Been doing some more testing with this PR and trying different solutions.
What is your idea on the following:

- We register the archive hooks for all Pods that have them enabled.
- We only register the single hooks for the current queried object

On the front filter we check the context we are in at that moment. If it's the current object we run the single template and otherwise we run the archive filter.

I get the feeling that the current solution is simple too general. Especially if there are multiple loops on one page of the same content type you'll definitely have problems finding the right template to show.

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -49,25 +49,12 @@ public function __construct() {
...
-		 * Default is false, set to true to enable.
-		 */
-		if ( ! defined( 'PFAT_USE_ON_EXCERPT' ) ) {
-			define( 'PFAT_USE_ON_EXCERPT', false );
-		}
-
 		$possible_pods = $this->auto_pods();
 
 		foreach ( $possible_pods as $pod_name => $pod ) {
 			$filter = $this->get_pod_filter( $pod_name );
```
sc0ttkclark, line 55:
```suggestion
			$filter = $this->get_pod_filter( $pod_name );

			if ( empty( $filter ) ) {
				continue;
			}
```

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
sc0ttkclark: Updated

JoryHogeveen:
Explanation for my commits.

Previously either the archive/list or the singular filter was loaded for a Pod. With this update, archive/list filters are always added and only singular filters are dependent on the current object.
My doing this the loading part of auto templates gets a lot simpler and in the hooked method it will check for the actual object and filter context.

JoryHogeveen (approved):

sc0ttkclark:
I've got some testing to do on my side to be sure this is right

sc0ttkclark:
@digicentral I need to do more testing on this PR to make sure it's totally safe and won't introduce new issues, but for now, you can use the ZIP from the branch which has been updated with everything from Pods 2.7.30.

https://github.com/pods-framework/pods/archive/refs/heads/fix/auto-template-add-filter.zip

PR #5990, opened 2021-04-05
WIP: Boolean Group fields

zrothauser:
## Description

Starts to implement Boolean Group fields.

Still needs some work:
- I'm not sure if the way it's implemented is ideal, adds too much logic to wrappers/components that are higher up in the app.
- The PHP code that converted boolean groups to checkbox fields is commented out, and should be handled correctly. 
- Performance issues, maybe due to extra logic for calculating dependencies. [Update: noticing similar performance issues in the release/2.8 branch, it's not due to changes here. Should be handled separately.]

- ~~Needs tests for the field~~
- ~~Unsure about the specification and additional options.~~
- ~~Need to implement the "help" text attributes.~~
- ~~Need to handle 'depends-on' subfields.~~

## Testing instructions

- There's a new entry in the Kitchen Sink which can be seen when editing a pod. It's at the top of the config file for convenience, but doesn't need to go there. 

- Fix: Fixes an issue where truthy values are not being handled correctly in Boolean fields. 
- Feature: Implements boolean group fields. 

## PR checklist


classes/PodsAdmin.php
```diff
@@ -1674,33 +1674,33 @@ protected function backcompat_convert_tabs_to_groups( array $tabs, array $option
 					$boolean_group = [];
 					$boolean_extra = [];
 
-					if ( ! empty( $field_options['boolean_group'] ) ) {
-						$boolean_group = $field_options['boolean_group'];
+					// if ( ! empty( $field_options['boolean_group'] ) ) {
```
zrothauser, line 1677: @sc0ttkclark Leaving this part to you to remove and clean up the right way :)

ui/js/dfv/src/fields/boolean/index.js
```diff
@@ -30,7 +34,7 @@ const Boolean = ( {
 				pick_format_single: formatType,
 				data: options,
 			} }
-			value={ value }
+			value={ toNumericBool( value ) }
```
zrothauser, line 37: Fixes an unrelated bug that I noticed while working on this.

PR #5995, opened 2021-04-16
Support taxonomy filters in Pods::find() and other object fields

### LINKED ISSUE #5994: Support taxonomies in Pods::find() filters
## Problem

Got this question from Slack.

Currently only pick fields are supported as filters in Pods::find() (and this in the shortcode).
We should also support taxonomies as pick types.

JoryHogeveen:
## Description

Support taxonomies as filters in Pods::find()

## Related GitHub issue(s)

Fixes #5994 

## Testing instructions

- Create some posts with various categories
- Add the following shortcode anywhere: `[pods name="post" filters="category" field="title"]`
- It should add a selectbox with all available categories
- Select a category and only the posts related to that category should show.

## PR checklist

### DIFF
--- a/classes/PodsData.php
+++ b/classes/PodsData.php
@@ -1236,14 +1236,16 @@ class PodsData {
 				$where  = array();
 				$having = array();
 
-				if ( ! isset( $params->fields[ $filter ] ) && ! isset( $params->object_fields[ $filter ] ) ) {
-					continue;
-				}
-
 				$attributes = pods_v( $filter, $params->fields, null );
 				if ( ! $attributes ) {
 					$attributes = pods_v( $filter, $params->object_fields, null );
 				}
+
+				if ( null === $attributes ) {
+					// Field not found.
+					continue;
+				}
+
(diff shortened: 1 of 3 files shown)


classes/PodsData.php
```diff
@@ -1236,12 +1236,17 @@ public function build( $params ) {
 				$where  = array();
 				$having = array();
 
-				if ( ! isset( $params->fields[ $filter ] ) ) {
+				$attributes = pods_v( $filter, $params->fields, null );
+				if ( ! $attributes ) {
+					$attributes = pods_v( $filter, $params->object_fields, null );
+				}
```
sc0ttkclark, line 1242:
```suggestion
				// Check if we have a matching field.
				$attributes = pods_v( $filter, $params->fields, null );

				// If we do not have a matching field, check for a matching object field.
				if ( ! $attributes ) {
					$attributes = pods_v( $filter, $params->object_fields, null );
				}
```

sc0ttkclark (approved):

PR #5999, opened 2021-04-20
Allow ID as parameter for audio and video

### LINKED ISSUE #5983: Output multiple Videos and Video URLs with Magic Tags
Hi,
I'm having the same issue I had with images (https://github.com/pods-framework/pods/issues/5982), now I'm having with videos. I need the video URL, and in this case, also the placeholder poster to output it in a gallery with other images from the post. I'm using a Video field because I need it to be able to select multiple files.
I tried:
`{@videos}` outputs nothing.
`{@videos,pods_video}` if there is only one video, it outputs a full HTML tagged vídeo, but it won't work for me, as I need a specific markup for the gallery. BUT, if there is more than one video on the field, it doesn't work at all.
`{@videos._src}` outputs nothing.
Even though I'm not using an oembed file, I also tried `{@videos,wp_oembed_get}`, but got nothing from it.

Am I missing something, or this is behaving as expected?
Thank you

JoryHogeveen:
## Description

Allow ID to be passed as param for `pods_audio` and `pods_video`

Will also enable the following magic tag usage:

```
[each videos]
{@ID,pods_video}
[/each]
```

## Related GitHub issue(s)

Fixes #5983 

## PR checklist

### DIFF
--- a/includes/media.php
+++ b/includes/media.php
@@ -410,20 +410,25 @@ function pods_image_resize( $attachment_id, $size ) {
  *
  * @since 2.5.0
  *
- * @param string|array $url  Can be a URL of the source file, or a Pods audio field.
- * @param bool|array   $args Optional. Additional arguments to pass to wp_audio_shortcode().
+ * @param string|array|int $url  Can be a URL of the source file, or a Pods audio field.
+ * @param bool|array       $args Optional. Additional arguments to pass to wp_audio_shortcode().
  *
  * @return string
  */
 function pods_audio( $url, $args = false ) {
 
-	if ( is_array( $url ) ) {
-		if ( ! is_null( pods_v( 'ID', $url ) ) ) {
-			$id  = pods_v( 'ID', $url );
-			$url = wp_get_attachment_url( $id );
-		} else {
-			return;
+	if ( ! is_string( $url ) ) {
+		$id = $url;
+		if ( is_array( $id ) ) {
+			$id = pods_v( 'ID', $url );
 		}
+		if ( is_numeric( $id ) ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark (dismissed):

sc0ttkclark:
@JoryHogeveen I modified the logic a little to simplify it, can you confirm it's still good?

includes/media.php
```diff
@@ -410,26 +410,30 @@ function pods_image_resize( $attachment_id, $size ) {
...
-			$id = pods_v( 'ID', $url );
-		}
-		$url = wp_get_attachment_url( $id );
+	// Support arrays.
+	if ( is_array( $url ) ) {
+		$url = pods_v( 'ID', $url );
+	}
+
+	// Support IDs.
+	if ( false !== strpos( $url, '://' ) && is_numeric( $url ) ) {
```
JoryHogeveen, line 426: @sc0ttkclark  Why not just check for `is_numeric`? It's faster and makes the strpos redundant.

sc0ttkclark: that was supposed to be a ||, will fix

```suggestion
	if ( is_numeric( $url ) || false !== strpos( $url, '://' ) ) {
```

includes/media.php
```diff
@@ -446,26 +450,30 @@ function pods_audio( $url, $args = false ) {
...
```
JoryHogeveen, line 466: @sc0ttkclark  Why not just check for `is_numeric`? It's faster and makes the `strpos` redundant.

```suggestion
	if ( is_numeric( $url ) ) {
```

JoryHogeveen:
@sc0ttkclark Looks good to me!

PR #6030, opened 2021-05-24
Use Tippy for help tooltips

zrothauser:
## Description

Replaces help tooltip implementation to fix overflow issues. 

## Testing instructions

1. Check help tooltips in the "Kitchen Sink" for a pod.
2. Also check tooltips in an "edit" modal. 

Enhancement: Replaces help tooltips with a library to better calculate positions. 

## PR checklist


lkraav:
So WP core doesn't ship any React tooltip library with block editor?

zrothauser:
@lkraav There is one, but the tooltips here need to work a bit differently: 1. this package handles positioning automatically better, 2. better works with HTML content (mainly links) inside the tooltip, and 3. should be shown when a "help" icon is clicked, whereas the WP component is more for helper text to explain UI icons.

lkraav:
> There is one, but the tooltips here need to work a bit differently: 1. this package handles positioning automatically better, 2. better works with HTML content (mainly links) inside the tooltip, and 3. should be shown when a "help" icon is clicked, whereas the WP component is more for helper text to explain UI icons.

:( it sure is unfortunate everyone has to ship their own..

Issue #6048, opened 2021-06-18
PHP Notice "Indirect modification of overloaded property PodsArray::$fields has no effect" on Custom Settings Page


remcokalf:
I have a Pods Custom Settings Page with a few custom settings.

I noticed a PHP Notice on the settings page itself:

Indirect modification of overloaded property PodsArray::$fields has no effect

See attached screenshot for more info from Query Monitor plugin.

![Schermafbeelding 2021-06-18 om 12 19 44](https://user-images.githubusercontent.com/4609920/122549081-64062a80-d032-11eb-8fbc-a1434c306705.png)

Wordpress 5.7.2
Pods 2.7.28
PHP 7.4.12

PR #6051, opened 2021-06-18
PHP Notice "Indirect modification of overloaded property PodsArray::$fields has no effect" on Custom Settings Page

unknownnf:
## Description

Update the way $options->fields is modified by first making a local copy then replacing the one in the object with the modified local variable.

The old method would only update the local copy of fields['manage'] and the object reference would remain untouched with an empty manage key.

## Related GitHub issue(s)

#6048

### DIFF
--- a/classes/PodsUI.php
+++ b/classes/PodsUI.php
@@ -1134,9 +1134,11 @@ class PodsUI {
 
 		// Set up default manage field.
 		if ( empty( $options->fields['manage'] ) ) {
-			$options->fields['manage'][ $options->sql['field_index'] ] = [
-				'label' => __( 'Name', 'pods' ),
-			];
+			$fields = $options->fields; // make a local copy of the fields
+			$fields['manage'][ $options->sql['field_index'] ] = array(
+					'label' => __( 'Name', 'pods' ),
+			);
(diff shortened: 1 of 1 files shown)


classes/PodsUI.php
```diff
@@ -1134,9 +1134,11 @@ public function setup( $options ) {
...
 		// Set up default manage field.
 		if ( empty( $options->fields['manage'] ) ) {
-			$options->fields['manage'][ $options->sql['field_index'] ] = [
-				'label' => __( 'Name', 'pods' ),
-			];
+			$fields = $options->fields; // make a local copy of the fields
+			$fields['manage'][ $options->sql['field_index'] ] = array(
+					'label' => __( 'Name', 'pods' ),
+			);
+			$options->fields = $fields;
```
sc0ttkclark, line 1141:
```suggestion
			// Make a local copy of the fields.
			$fields = $options->fields;

			// Change the info.
			$fields['manage'][ $options->sql['field_index'] ] = [
					'label' => __( 'Name', 'pods' ),
			];

			// Set the object fields property.
			$options->fields = $fields;
```

sc0ttkclark (approved):

PR #6055, opened 2021-06-24
Fix meta_user do_action not using the right object.

unknownnf:
## Description

Fix warnings on admin when editing a user with a custom pod for non-existent $post variable, do_action would also not pass the right object over to hooks.

## Testing instructions

1. Create a user pod, add some fields to it.
2. Go to edit the user.

## PR checklist

### DIFF
--- a/classes/PodsMeta.php
+++ b/classes/PodsMeta.php
@@ -1944,6 +1944,7 @@ class PodsMeta {
 		if ( is_object( $user_id ) ) {
 			$user_id = $user_id->ID;
 		}
+		$user = get_userdata($user_id);
 
 		$id  = $user_id;
 		$pod = null;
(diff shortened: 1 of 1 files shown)


classes/PodsMeta.php
```diff
@@ -1944,6 +1944,7 @@ public function meta_user( $user_id ) {
 		if ( is_object( $user_id ) ) {
 			$user_id = $user_id->ID;
 		}
+		$user = get_userdata($user_id);
```
sc0ttkclark, line 1947:
```suggestion
		if ( is_object( $user_id ) ) {
			$user    = $user_id;
			$user_id = $user_id->ID;
		} else {
			$user = get_userdata( $user_id );
		}
```

sc0ttkclark (approved):

PR #6062, opened 2021-07-05
Resolve `strlen()` warning

pd-cm:
## Description
Resolve `Warning: strlen() expects parameter 1 to be string, array given in wp-content/plugins/pods/classes/fields/text.php on line 168`

## Related GitHub issue(s)

None.

## Testing instructions

I was getting this error output. Casting the value to `(string)` resolves the Warning and still serves the purpose of the function.

## Screenshots / screencast

None.

* Fix a PHP error output.

## PR checklist

### DIFF
--- a/classes/fields/text.php
+++ b/classes/fields/text.php
@@ -108,6 +108,21 @@ class PodsField_Text extends PodsField {
 	}
 
 	/**
+	 * @inheritDoc
+	 */
+	public function is_empty( $value ) {
+		$is_empty = false;
+
+		$value = (string) $value;
+
+		if ( '' === $value ) {
+			$is_empty = true;
+		}
+
+		return $is_empty;
+	}
+
+	/**
 	 * {@inheritdoc}
 	 */
 	public function display( $value = null, $name = null, $options = null, $pod = null, $id = null ) {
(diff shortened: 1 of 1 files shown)


JoryHogeveen:
Hi @pd-cm 
This would fix the warning, though I'm curious why this happens at all..
As this is a text field it shouldn't get an array in the first place..

JoryHogeveen (approved):

classes/fields/text.php
```diff
@@ -165,7 +180,7 @@ public function validate( $value, $name = null, $options = null, $fields = null,
 		if ( is_array( $check ) ) {
 			$errors = $check;
 		} else {
-			if ( 0 < strlen( $value ) && '' === $check ) {
+			if ( $this->is_empty( $check ) ) {
```
sc0ttkclark, line 183: This could simplify things, we only care about strings here and it's either empty or not.

```suggestion
			if ( '' !== $value && '' === $check ) {
```

classes/fields/text.php
```diff
@@ -107,6 +107,21 @@ public function schema( $options = null ) {
...
+
+		$value = (string) $value;
+
+		if ( '' === $value ) {
+			$is_empty = true;
+		}
+
+		return $is_empty;
+	}
+
```
sc0ttkclark, line 124:
```suggestion
```

Issue #6146, opened 2021-09-22
[Pods 2.8 Bug] Post ID isn't passed using Plupload AJAX uploader


JoryHogeveen:
Might also be a 2.7 issue. Going to test a.s.a.p.:
Initial report: https://wordpress.org/support/topic/magic-tags-in-custom-upload-directory/

## Pods 2.8 version

2.8.0-rc.1

## Description

When using Plupload we also send the current post. This is used for uploading context.
This isn't send anymore in Pods 2.8
https://github.com/pods-framework/pods/blob/main/classes/fields/file.php#L535-L540

## Testing instructions

4. See error message '....'

## Other super helpful information

<details>
<summary>WordPress Environment</summary>
  
```
Copy and paste your Tools > Site Health > Info. Use the 'Copy Site Info to Clipboard' and paste the results over this line.
```

</details>

<details>
<summary>Pods Package Export</summary>

```json
Copy and paste your Pods Admin > Migrate: Packages > Export and paste the code over this line.
```

</details>

sc0ttkclark:
Great find!

Fixed via 39166daf2454297f55f8c8606fad0faf3975b31d

PR #6147, opened 2021-09-23
Improve Plupload item context

JoryHogeveen:
## Description

Allows all Pod types for plupload context and enhances support for Posts.

## Related GitHub issue(s)

Fixes: #6146 

## PR checklist

### DIFF
--- a/classes/fields/file.php
+++ b/classes/fields/file.php
@@ -544,6 +544,7 @@ class PodsField_File extends PodsField {
 					'pod'      => $pod_id,
 					'field'    => $field_id,
 					'item_id'  => $item_id,
+					'post_id'  => $item_id,
 					'uri'      => $uri_hash,
 				],
 			];
--- a/classes/fields/file.php
+++ b/classes/fields/file.php
@@ -510,6 +510,7 @@ class PodsField_File extends PodsField {
 
 			$pod_id   = (int) pods_v( 'pod_id', $args->pod, 0 );
 			$field_id = (int) pods_v( 'id', $options, 0 );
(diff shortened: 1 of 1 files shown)


classes/fields/file.php
```diff
@@ -547,6 +543,7 @@ public function build_dfv_field_options( $options, $args ) {
 					'method'   => 'upload',
 					'pod'      => $pod_id,
 					'field'    => $field_id,
+					'item_id'  => $item_id,
```
sc0ttkclark, line 546: For backcompat on any hooks that were looking for `post_id`

```suggestion
					'item_id'  => $item_id,
					'post_id'  => $item_id,
```

sc0ttkclark (approved):

classes/fields/file.php
```diff
@@ -544,6 +544,7 @@ public function build_dfv_field_options( $options, $args ) {
 					'pod'      => $pod_id,
 					'field'    => $field_id,
 					'item_id'  => $item_id,
+					'post_id'  => $item_id,
```
JoryHogeveen, line 547: @sc0ttkclark 
This part is misleading since this could be a tax ID or user ID as well. It's the Pod item ID, not necessarily a Post ID.

JoryHogeveen: Also, I think if this is preferable then it should get a comment about backwards compatibility so we know why this is still here in the future.

JoryHogeveen: Resolved in https://github.com/pods-framework/pods/pull/6151

PR #6190, opened 2021-10-19
Allow Pods fields to be registered properly with REST API

### LINKED ISSUE #6175: App using REST API doesn't work after upgrade to 2.8
## Description

I have a third party app using REST API. It's not working with 2.8. If I revert to 2.7 works again
I check end points and see fields are now removed. I guess there should be some way to get it back again, but after looking over website can't figure how to do it

this is the current url that works with 2.7.31:
https://mysite.com/wp-json/wp/v2/product/

How should it looks now in 2.8?

thanks

jamesgol:
Fixes #6175

## Description

The 'type' that needs to be registered with the REST API is the object name instead of the actual type.

## Related GitHub issue(s)

#6175

### DIFF
--- a/classes/PodsRESTFields.php
+++ b/classes/PodsRESTFields.php
@@ -145,13 +145,11 @@ class PodsRESTFields {
 			$rest_args['pods_update'] = true;
 		}
 
-		$object_type = $this->pod->get_type();
+		// Get the object type for register_rest_field(), this is documented weird and we should just pass the object name.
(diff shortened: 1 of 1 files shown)


jamesgol:
I never trust the docs and prefer to look at the code and debugger.  In 2.7.x when it worked properly the pod name was used since that is the object that is extended.

In WordPress core it just stuffs it into a global array based on the 'object_type'

PR #6411, opened 2022-01-19
Retain backwards compatibility with field name passed for pick fields

jamesgol:
## Description

The change here https://github.com/pods-framework/pods/commit/55d150b631d153a29ae6fa8afca55226dd2727bc broke query generation on pick fields that don't have an explicit reference.  

For years now I've had code that called
```
$where = array( array( 'key' => 'style_revision', 'value' => '2015' ) );
$params = array( 'where' => $where, 'limit' => 16 );
$pod->find($params);

```

This stopped working with 2.8.  I can fix it in my code by changing the key to `style_revision.post_title` which might have always been the proper way to do it but we should retain backwards compatibility.

### DIFF
--- a/classes/PodsData.php
+++ b/classes/PodsData.php
@@ -2789,7 +2789,12 @@ class PodsData {
 								$table = $the_field->get_table_info();
 
 								if ( ! empty( $table ) ) {
-									$field_cast = "`{$field_name}`.`" . $table['field_id'] . '`';
+									if ( is_int( $field_value ) ) {
+										$field_cast = "`{$field_name}`.`" . $table['field_id'] . '`';
+									} else {
+										// Prior to 2.8 this was the default query, retain backwards compatibility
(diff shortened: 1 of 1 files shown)


classes/PodsData.php
```diff
@@ -2789,7 +2789,12 @@ public static function query_field( $field, $q, $pod = null, &$params = null ) {
 								$table = $the_field->get_table_info();
 
 								if ( ! empty( $table ) ) {
-									$field_cast = "`{$field_name}`.`" . $table['field_id'] . '`';
+									if ( is_int( $field_value ) ) {
```
sc0ttkclark, line 2792: Should we check numeric or just int?

```suggestion
									if ( is_numeric( $field_value ) ) {
```

jamesgol: is_numeric wouldn't be backwards compatible.  The query string I'm using is '2015'

sc0ttkclark (approved):

PR #6422, opened 2022-01-30
Error prevention in WP_Query_Integration

petersentner:
It can happen that query_posts incorrectly results in a category query. Then the get_queried_object is empty and PHP throws a notice that it is trying to retrieve a property from an empty object.

So it is better to check that get_queried_object is not empty.


petersentner:
I have just noticed that the problem is solved in version 2.8.9

PR #6437, opened 2022-02-05
Add setting to turn Pods into types-only mode which disables groups/fields hooks and screens

sc0ttkclark:
## Description

Types-only mode is geared towards people who want to use Pods for content types but prefer to use another custom field plugin. Types-only mode will be more performant and have reduced overhead on many areas as a result of no longer running hooks/logic checks on meta handling as well.

## Testing instructions

1. Go to Pods Admin > Settings
2. Enable / Disable Custom Fields handling
3. See the Groups / Fields no longer showing in Pods Admin > Edit Pods and the Fields tab is removed from the Edit Pod screen

## Screenshots / screencast

Screen Shot 2022-02-05 at 6 09 01 PM

> Added: New Types-only mode introduced in the Pods Admin > Settings screen for those wanting to use Pods only for content types. This mode improves performance and reduces hooks/logic where fields are concerned. #TBD (@sc0ttkclark, @zrothauser)

## PR checklist

## Remaining Tasks

- [x] Store the boolean option for podsAdminConfig.global.showFields in the DFV app for reference (@zrothauser)
- [x] If showFields is off, do not add "Fields" tab (@zrothauser)
- [x] If showFields is off, do not add Groups/Fields management to screen (@zrothauser)
- [x] If showFields is off, do not send Groups/Fields data to the REST API (@zrothauser)


lkraav:
DRY and simpler UX would dictate "Manage custom fields" option to be a checkbox.

"Sessions" - options should probably be in the same logical order: Auto, Enable, Disable

sc0ttkclark:
@lkraav I updated the session option ordering to match what you said but the option for the fields enable/disable -- I feel is best to keep as the radio fields for now so we can make sure it's more clear what happens in each case. We can change it if we see the need in the future though, just depends on whether all of the text in the label / description / help is sufficient enough for people when they start using it.

JoryHogeveen:
@sc0ttkclark 
This would effectively render Pods to be a simple CPT/CT creator right? Setting pages become obsolete if fields are disabled.

sc0ttkclark:
@JoryHogeveen Correct, this would effectively reduce the available content types and remove the fields management from Pods administration itself.

sc0ttkclark:
I'll need to fix the tests before merging this

PR #6453, opened 2022-02-16
Fix z-index issue with WP editor

### LINKED ISSUE #6452: z-index issue with WYSIWYG editor after a relationship field.
### Description

https://wordpress.org/support/topic/custom-field-dropdown-list-z-index-issue/

### Version

2.8.10

### Testing Instructions

- Create/extend Pod
- Add relationship field (list view) and wysiwyg field (after relationship)
- Make sure there are several relationship options
- The select box is under the editor when opening.

### Screenshots / Screencast

image

JoryHogeveen:
## Related GitHub issue(s)

Fixes #6452

## Testing instructions

- Create/extend Pod
- Add relationship field (list view) and wysiwyg field (after relationship)
- Make sure there are several relationship options
- The select box is under the editor when opening. When applying this patch it's visible!

## PR checklist


ui/styles/src/base/_form.scss
```diff
@@ -136,6 +140,8 @@ tr.pods-form-ui-row-type-heading td {
 	padding: 0;
 }
 
+/* Pick */
```
sc0ttkclark, line 143:
```suggestion
/* Pick field */
```

ui/styles/src/base/_form.scss
```diff
@@ -85,6 +85,10 @@ div.pods-textfield input,
 }
 
 /* textarea/wysiwyg */
+.pods-dfv-field .wp-editor-tools {
+	z-index: auto;
```
nicdford, line 89: My only thought about this would be to check the select field and adjust the z-index to pull it up higher than the WYSIWYG; instead of pulling the WYSIWYG editor down.

I looked into the select field a bit, but couldn't track down exactly where that CSS is coming from. It seems like its loading from JS somewhere, but I can't find it.

nicdford (approved):

ui/styles/src/base/_form.scss
sc0ttkclark: @nicdford Check the DFV src folder for the field type specific JS, there are some embedded styles there. The aim was to keep those newer things tied directly to each field type for organization purposes.

JoryHogeveen: @nicdford  I wanted to do that as well but that didn't really work very well due to the layering of elements.
After trial and error I didn't really understand why the editor had a z-index in the first place so I simply removed it :)

nicdford: Yeah, I'm not sure about that either. It looks like it's coming from WP core from what I can tell. I was able to find the react-select component though and add a style there if you agree that method is less intrusive, I'd say we run with that.

sc0ttkclark (approved):

PR #6541, opened 2022-06-23
Allow language relationships for Polylang

### LINKED ISSUE #6540: Relationship to languages not working anymore (Polylang)
### Description

Since languages are simply stored as taxonomies our integration should support it as relationship objects.

### Version

2.8.0 - 2.8.20

JoryHogeveen:
## Description

Adds a relationship type for PLL languages.

@sc0ttkclark Should this be defined as a simple relationship?

## Related GitHub issue(s)

Fixes #6540

## Testing instructions

1. Install Polylang and create some languages
2. Add a relationship field
3. You should see a new group "Polylang" with a "languages" relationship type.
4. In terms of storage it should simply store as a taxonomy relationship.

## PR checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pods",
-  "version": "2.8.19",
+  "version": "2.8.20",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "pods",
-      "version": "2.8.19",
+      "version": "2.8.20",
       "license": "GPL-2.0+",
       "dependencies": {
         "@babel/runtime": "^7.17.9",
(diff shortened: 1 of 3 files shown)


sc0ttkclark:
Yes, use a simple relationship with a callback. That way it's not loaded up every page load when it gets the related objects.

JoryHogeveen:
@sc0ttkclark 
Why would we need a separate callback? These are simply taxonomies so would work the same right?
By adding the `taxonomy-` prefix it seems it's already working great :)

sc0ttkclark:
Sorry, I got confused about what was happening here.

We exclude the taxonomy here: https://github.com/pods-framework/pods/blob/b5882d6c956adc26a0d8a8461007fd1946cbc6bd/src/Pods/Integrations/Polylang.php#L94

So we just add it back under it's own group in this code now. I'm fine with that, it makes it easier to see too.

sc0ttkclark (approved):

Issue #6577, opened 2022-08-04
Support WebP as default image format


JoryHogeveen:
### Description

https://wordpress.org/support/topic/webp-support-39/#post-15886383

Currently our default list of image types doesn't include WebP.

### Version

2.8.23

sc0ttkclark:
Merged into Pods 2.9

PR #6578, opened 2022-08-04
Support webp as default image extension

JoryHogeveen:
## Description

This PR adds the webp extension to the default image extensions for the file field.

## Related GitHub issue(s)

Fixes: https://github.com/pods-framework/pods/issues/6577 (verified)

## PR checklist


sc0ttkclark:
We should technically be supporting any registered audio/ mime type for audio, and video/ mime type for videos, and image/ mime type for images.

See this for the logic we should really be using here. https://github.com/pods-framework/pods/blob/release/2.9/classes/fields/file.php#L1157

That gives full support for whatever WP and other plugins enable for image/video/audio.

The "etc" does a lot of heavy lifting here so we could describe that as " and any others registered with WP"

sc0ttkclark:
Updated PR with the refactor/changes to future-proof our support of the main media types and reduce logic duplication that existed in the handling of it.

@JoryHogeveen If you can put this through a few tests, that'd be helpful -- otherwise I will do that in the next few days before merging.

JoryHogeveen:
`images-any` > webp doesn't show in the available list when selecting files. File filter not correct or WP doesn't include webp by default?

Other than that it seems good. WP indeed adds several other MIME types if the any option is selected.

sc0ttkclark:
Merging and we can sort out whatever edge case there is with images-any later. It all works when I run through the tests on my side so there may be something else going on that we can figure out later.

Issue #6895, opened 2022-09-07
 Pick field autocomplete returns `Invalid field request` for the fields configured via code


naveen17797:
When  i register the pick field with code by following this tutorial https://docs.pods.io/code/registering-configurations/#Registering_Pods_with_Code, it returns invalid field request on the ajax call.

when i checked the code, i found this is happening due to this snippet

https://github.com/pods-framework/pods/blob/cb8b515e513976195ac45961d383fe50ccffa083/classes/fields/pick.php#L2969

initially $pod['id'] is 0, so i added 'id' to pods registration, also added 'id', 'pod_id' to field registration

```
    $pod = [
        'id'    => 100,
        'name'           => 'book_author',
        'label'          => 'Authors',

        'type'           => 'post_type',
        'storage'        => 'meta',
        'label_singular' => 'Author',
        'public'         => '1',
    ];
```
Then i noticed the code tries to load the field from the database which usually returns null

https://github.com/pods-framework/pods/blob/main/classes/fields/pick.php#L2942

naveen17797:
i created a PR here 
https://github.com/pods-framework/pods/pull/6896
which tries to load the field from store initially, then queries the database if its not found. i am not sure if its the correct way to fix the issue. Can you take a look at the issue and guide me for the right way to fix this issue ?

PR #6896, opened 2022-09-07
see #6895: Fix pick field autocomplete

naveen17797:


classes/fields/pick.php
```diff
@@ -2939,12 +2940,18 @@ public function admin_ajax_relationship() {
...
 
 		$pod   = self::$api->load_pod( array( 'id' => (int) $params->pod ) );
-		$field = self::$api->load_field(
-			array(
-				'id'         => (int) $params->field,
-				'table_info' => true,
-			)
-		);
+
+		$field = Store::get_instance()->get_object( $params->field );
```
sc0ttkclark, line 2944: Unfortunately this store call is not the correct fix even though it solves it for your specific case. Store calls are all limited across Pods to PodsAPI so that it can properly handle multiple stores and prevent conflicts with fields from other pods too.

The correct fix would be to adjust the code so that it's not looking up by ID but by the field name while also passing the pod name. Pods is still transitioning from reliance on IDs towards using names/slugs in more places for better support for code/file based configs like this.

There may be other areas of the code that need to be adjusted further too in this way, but the last big area is definitely sorting out the wp_podsrel table and writing a migration script that can deal with lots of data.

That's all a big undertaking and not part of this fix but I wanted to give you the big picture of what we're doing and where we're going with things like this.

I'll push some updates after I get some fixes in, but I'm also going to write a test for this method as well to make sure it's totally solved and won't break again.

sc0ttkclark (changes requested):

PR #6909, opened 2022-09-19
Fixes #6907 -- display_filter always set to the_title for `Display Field in Selection List`

### LINKED ISSUE #6907: Bug with $display_filter in pods/classes/fields/pick.php
### Description

The value of `$display_filter` is computed using `$search_data->field_index` but a `$field_index` variable is computed below and used in `$filter_args`:

```PHP
              $filter_args = array(
                $display_filter,
                $result[ $field_index ],
              );
```

I think `$display_filter` ought to be computed as follows in the `$results` foreach loop:

```PHP
            // COMPUTE $display_filter HERE with the new $field_index
            $display_filter = pods_v( 'display_filter', pods_v( $field_index, $search_data->object_fields ) );

            if ( 0 < strlen( $display_filter ) ) {
              $display_filter_args = pods_v( 'display_filter_args', pods_v( $field_index, $search_data->pod_data['object_fields'] ) );

              $filter_args = array(
                $display_filter,
                $result[ $field_index ],
              );
```

### Version

2.9.3

### Testing Instructions

1) Set `Display Field in Selection List` to `{@permalink}` for a custom field added to a (custom?) post type.
2) Add a `the_title` filter (to functions.php for instance).

`{@permalink}` will set `$field_index` to `post_name` and the original `$display_filter` will be empty as there are no filters for slugs. But the original `$display_filter` will be computed with the default field index for post type, which is the post title, so `$display_filter` below will not be empty:

```PHP
$display_filter = pods_v( 'display_filter', pods_v( $search_data->field_index, $search_data->object_fields ) );
```

This will result in if `$display_filter` if condition been true and `$filter_args` being evaluated as:

```
Array
(
    [0] => the_title
    [1] => ...post name (slug) here...
)
```

which is twice wrong, i.e. it's using the post slug instead of the post title, and it's missing the post ID as 2nd argument, which is necessary to run `the_title` filter.

### Screenshots / Screencast

_No response_

### Possible Workaround

_No response_

### Site Health Information

_No response_

### Pods Package

_No response_

therealgilles:
## Description

This fixes the filter applied to the `Display Field in Selection List` field when applicable. Before this fix, the default `the_title` filter was applied to the field, even though it may not have been applicable. An example is if the field value is set to `{@permalink}` to display the field as a slug (i.e. post_name).

## Related GitHub issue(s)

Fixes #6907.

## Testing instructions

See #6907.

Bug:`$display_filter` needs to be computed with the updated `$field_index` value. #6907 

## PR checklist


therealgilles:
@JoryHogeveen In the code below:

```PHP
              $display_filter_args = pods_v( 'display_filter_args', pods_v( $field_index, $search_data->pod_data['object_fields'] ) );
```

Is `$search_data->pod_data['object_fields']` the same as `$search_data->object_fields`? If so, this line could be replaced with:

```PHP
              $display_filter_args = pods_v( 'display_filter_args', $display_filter ) );
```

which would make most sense.

sc0ttkclark (approved):

PR #6930, opened 2022-09-29
Traverse field formatting

### LINKED ISSUE #6929: Magic tags in pods template not displaying proper number formatting when pulling from a related post type
### Description

I have two custom post types in wordpress - 'credit cards' and 'rebate offers'

I am using  a pods template for rebate offer single pages. Some of the data on the page is pulled in using magic tags, from rebate offer custom fields, and other data from credit card custom fields, using relationship field (each rebate offer is related to a credit card). 

In this example - I have a field in credit cards called 'current bonus' , and the value is 60,000. When I use current_bonus elsewhere on the site it displays the default number formatting with the comma. However when I pull it into this template using the magic tag from the relationship field: {@offer_credit_card.current_bonus}, it loses this formatting.

I also noticed when I use the relationship for currencies, it loses the '$', but if I use a currency field directly eg. {currency} from rebate_offer instead of {@offer_credit_card.currency}  from credit_card, it keeps the '$' formatting as set in pods field configuration.

See attached images. 

### Version

2.9.8

### Testing Instructions

_No response_

### Screenshots / Screencast

pods-capture1
pods-capture2
pods-capture3
pods-capture4

### Possible Workaround

_No response_

### Site Health Information

_No response_

### Pods Package

_No response_

JoryHogeveen:
## Description

Field formatting isn't working when traversing through fields. This PR aims to fix this.

## Related GitHub issue(s)

Fixes #6929

## PR checklist


JoryHogeveen:
@sc0ttkclark This PR is a beginning.

The only issue I couldn't find yet is how to fix repeatable fields with multiple values.

https://github.com/pods-framework/pods/blob/2.9.8/classes/Pods.php#L1396-L1411
Because of the data loop the repeatable fields (which are already an array) adds an array to the value array here:
https://github.com/pods-framework/pods/blob/2.9.8/classes/Pods.php#L1574

Because of this a simple repeatable field will result in the following value:

```
array(
  0 => array(
    0 => 'value 1',
    1 => 'value 2',
  )
)
```

This should be 

```
array(
  0 => 'value 1',
  1 => 'value 2',
)
```

The only thing is that I'm not sure at what point we should tackle this issue...

sc0ttkclark:
Updated compatibility for `Pods::field()` to handle related repeatable fields to prevent nested arrays of values and to always return the array of values together. Also added the formatting checks to the traversal tests.

sc0ttkclark (approved):

PR #6972, opened 2022-12-14
Pods 2.9.10.2

sc0ttkclark:


lkraav (commented):
I patched my live instance with 693dbfbdf5f7b0dab89a75669405aba8deed161c and at least things aren't crashing instantly, but it might take a bit for that specific flow to occur, to verify whether this patch the original problem.

PR #7041, opened 2023-03-27
Date format mapper - redo

JoryHogeveen:
Related: #4525
Related: #5675

## Description

This is a redo from an earlier PR: #5676
With this PR you can set your own formats when using date fields.

Text from prev PR:

> The relates issue noted that using {@post_date} always displays the mysql format (Y-m-d h:i).
> This PR sets this to the default formatting settings from WordPress.
> 
> This PR also adds a format param to the magic tag: {@post_date._format.Y-m-d}, {@post_date._format.Y m d} and even {@post_date._format.Y.m.d} (note the dots).
> Currently this PR only supports this for core-field dates (post_date). But might be a cool feature for other date fields as well.
>
> Would like to have feedback/ideas on how to best implement this.

This PR does NOT change the default behavior. It only adds the option to define a format (or select WP default format).

## Testing instructions

4. See '....'

## PR checklist

### DIFF
--- a/src/Pods/Data/Map_Field_Values.php
+++ b/src/Pods/Data/Map_Field_Values.php
@@ -754,7 +754,6 @@ class Map_Field_Values {
 		switch ( $traverse[0] ) {
 			case '_format':
 				$format    = $traverse[1];
-				$wp_format = in_array( strtolower( $format ), array( 'wp', 'wordpress' ), true );
 
 				switch ( $format ) {
 					case 'date':
--- a/src/Pods/Data/Map_Field_Values.php
+++ b/src/Pods/Data/Map_Field_Values.php
@@ -746,13 +746,30 @@ class Map_Field_Values {
 
 		$value = $obj->field( $field, array( 'raw' => true ) );
 
+		// No fields modifiers found.
 		if ( empty( $traverse[0] ) || empty( $traverse[1] ) ) {
 			return null;
 		}
 
 		switch ( $traverse[0] ) {
 			case '_format':
-				$format = $traverse[1];
+				$format    = $traverse[1];
+				$wp_format = in_array( strtolower( $format ), array( 'wp', 'wordpress' ), true );
+
+				if ( $wp_format ) {
+					switch ( $field ) {
+						case 'date':
+							$format = get_option( 'date_format' );
+							break;
+						case 'time':
+							$format = get_option( 'time_format' );
+							break;
+						default:
+							$format = get_option( 'date_format' ) . ' ' . get_option( 'time_format' );
+							break;
+					}
+				}
+
 				$value = date_i18n( $format, strtotime( $value ) );
 				break;
 			default:
--- a/src/Pods/Data/Map_Field_Values.php
+++ b/src/Pods/Data/Map_Field_Values.php
@@ -58,6 +58,7 @@ class Map_Field_Values {
 			'context_info',
 			'calculation',
 			'image_fields',
+			'date_fields',
 			'avatar',
 		];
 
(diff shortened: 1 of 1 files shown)


src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+						$format = get_option( 'time_format' );
+						break;
+					case 'datetime':
+					case 'wp':
+					case 'wordpress':
+						$format = get_option( 'date_format' ) . ' ' . get_option( 'time_format' );
+						break;
+				}
+
+				$value = date_i18n( $format, strtotime( $value ) );
```
sc0ttkclark, line 776: We may need to do something for the time field in `$date_fields`.

src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+	 *
+	 * @return null|mixed The date field value or null if there was no match.
+	 */
+	public function date_fields( $field, $traverse, $field_data, $obj = null ) {
+
+		// Default date field handlers.
+		$date_fields = array(
+			'date',
+			'time',
+			'datetime',
```
sc0ttkclark, line 725: Are these just special fields that will output the current date/time/datetime? If so, we should check whether the field exists before calling `->field()` on it (only if it's date/time/datetime) and then use the current date+time in that case.

src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+		);
+
+		// Skip if the field exists.
+		if ( $field_data && ! in_array( $field_data->get_type(), $date_fields, true ) ) {
```
sc0ttkclark, line 729: If we are dealing with post_date etc, those fields will match an object field and be skipped.

src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+			$date_fields[] = 'post_modified';
+			$date_fields[] = 'post_modified_gmt';
+		}
+
+		// Handle special field tags.
+		if ( ! in_array( $field, $date_fields, true ) ) {
+			return null;
+		}
+
+		$value = $obj->field( $field, array( 'raw' => true ) );
```
sc0ttkclark, line 747: We need a case for current date/time/datetime values.

```suggestion
		$value = $field_data ? $obj->field( $field, array( 'raw' => true ) ) : current_time( 'mysql' );
```

src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+	public function date_fields( $field, $traverse, $field_data, $obj = null ) {
+
+		// Default date field handlers.
+		$date_fields = array(
+			'date',
+			'time',
+			'datetime',
+		);
+
+		// Skip if the field exists.
```
sc0ttkclark, line 728:
```suggestion
		// Skip if the field exists and is NOT a date/time/datetime field type.
```

src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+		}
+
+		$object_type = $obj ? $obj->pod_data->get_type() : null;
+
+		if ( 'post_type' === $object_type ) {
+			$date_fields[] = 'post_date';
+			$date_fields[] = 'post_date_gmt';
+			$date_fields[] = 'post_modified';
+			$date_fields[] = 'post_modified_gmt';
+		}
```
sc0ttkclark, line 740: I don't think we need this logic

```suggestion
```

src/Pods/Data/Map_Field_Values.php
sc0ttkclark, line 747: We need a case for current date/time/datetime values.

```suggestion
		// Handle getting current field value.
		if ( $field_data ) {
			$value = $obj->field( $field, array( 'raw' => true ) );
		} else {
			$value = current_time( 'mysql' );
		}
```

sc0ttkclark (changes requested):
Some changes here needed but it's very close. Let's be sure to add coverage for all of the available options/logic in the `tests/codeception/wpunit/Pods/MappingTest.php` file.

src/Pods/Data/Map_Field_Values.php
```diff
@@ -703,6 +704,85 @@ public function image_fields( $field, $traverse, $field_data, $obj ) {
...
+	 * @param string[]                $traverse   The list of fields in the path excluding the first field name.
+	 * @param null|Field|Object_Field $field_data The field data or null if not a field.
+	 * @param Pods|null               $obj        The Pods object or null if not set.
+	 *
+	 * @return null|mixed The date field value or null if there was no match.
+	 */
+	public function date_fields( $field, $traverse, $field_data, $obj = null ) {
+
+		// Default date field handlers.
+		$date_fields = array(
```
sc0ttkclark, line 722: FYI we could use `PodsForm::date_field_types()` here but because it makes more sense to only keep this for date/time/datetime as kind of global special tags, I don't think it needs to match the current date field types at all and can remain as this hardcoded list.

This is more of a note for ourselves to reference for why this is here :)

classes/Pods.php
```diff
@@ -1696,7 +1697,7 @@ public function field( $name, $single = null, $raw = false ) {
 			$value = array_merge( ...$value );
 		}
 
-		if ( ! empty( $field_data ) && ( $params->display || ! $params->raw ) && ! $params->in_form && ! $params->raw_display ) {
+		if ( ! $is_mapped && ! empty( $field_data ) && ( $params->display || ! $params->raw ) && ! $params->in_form && ! $params->raw_display ) {
```
sc0ttkclark, line 1700: In all cases, are mapped values handling their own display output? Maybe, just worth confirming

sc0ttkclark (approved):
Looks good to me, we could improve with some assertions on an object field but it's good either way

sc0ttkclark (changes requested):
There's a test failure on the first assertion at https://github.com/pods-framework/pods/pull/7041/files#diff-2c2cd009a6ec8c79443141f820e785b79db4f0d63491de1d5a2bf0d978f271fbR620

https://github.com/pods-framework/pods/actions/runs/4554822589/jobs/8033220382?pr=7041#step:15:44

classes/Pods.php
JoryHogeveen: I did assume they would. Most mappers are related to iteration fields (index, row etc.).
There are two mappers that our unit tests should verify: Avatar and Image fields.

That said, I feel like mappers are and should be short circuit handlers for fields and thus should not have to be processed further.
The `Pods::field()` method should definitely be more abstracted to make this much more easy to follow and debug.

JoryHogeveen:
> There's a test failure on the first assertion at https://github.com/pods-framework/pods/pull/7041/files#diff-2c2cd009a6ec8c79443141f820e785b79db4f0d63491de1d5a2bf0d978f271fbR620
> 
> https://github.com/pods-framework/pods/actions/runs/4554822589/jobs/8033220382?pr=7041#step:15:44

I do not get why the tests are failing, and definitely not why it returns false.
When I test the field locally it works, only the tests (also locally) are failing.

PR #7146, opened 2023-08-31
Fix deprecated PHP

### LINKED ISSUE #7145: PHP 8+ deprecated issues
- [ ] Implicit float to int conversion: https://php.watch/versions/8.1/deprecate-implicit-conversion-incompatible-float-string#array-keys
- [ ] Dynamic properties
- [x] Automatic conversion of false to array is deprecated in \pods\components\Templates\includes\auto-template\Pods_Templates_Auto_Template_Settings.php on line 413

Locations (to be completed):
- https://github.com/pods-framework/pods/blob/main/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php#L58-L81

JoryHogeveen:
## Description

Changed the prio to an int,

## Related GitHub issue(s)

Fixes #7145 

## PR checklist

### DIFF
--- a/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
+++ b/components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
@@ -52,7 +52,7 @@ class Pods_Templates_Auto_Template_Front_End {
 		foreach ( $possible_pods as $pod_name => $pod ) {
 			$filter = pods_v( 'archive_filter', $pod, '', true );
 			if ( $filter ) {
-				$this->filtered_content[ $filter ] = 10.5;
+				$this->filtered_content[ $filter ] = 10;
 			}
 		}
 
@@ -75,7 +75,7 @@ class Pods_Templates_Auto_Template_Front_End {
 			// No need to check for default hooks, this is already done in auto_pods().
 			$filter = pods_v( 'single_filter', $possible_pods[ $pod ], '', true );
 			if ( $filter ) {
-				$this->filtered_content[ $filter ] = 10.5;
+				$this->filtered_content[ $filter ] = 10;
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -52,7 +52,7 @@ public function hook_content() {
 		foreach ( $possible_pods as $pod_name => $pod ) {
 			$filter = pods_v( 'archive_filter', $pod, '', true );
 			if ( $filter ) {
-				$this->filtered_content[ $filter ] = 10.5;
+				$this->filtered_content[ $filter ] = 10;
```
sc0ttkclark, line 55: I wonder if we should set this to 11 or not

components/Templates/includes/auto-template/Pods_Templates_Auto_Template_Front_End.php
```diff
@@ -75,7 +75,7 @@ public function hook_content() {
 			// No need to check for default hooks, this is already done in auto_pods().
 			$filter = pods_v( 'single_filter', $possible_pods[ $pod ], '', true );
 			if ( $filter ) {
-				$this->filtered_content[ $filter ] = 10.5;
+				$this->filtered_content[ $filter ] = 10;
```
sc0ttkclark, line 78: I wonder if we should set this to 11 or not

JoryHogeveen: I choose 10 because when I tested `(int) 10.5` it returned 10. But yeah, I had the same thought myself haha.
However, we allow users to override this value so personally I wouldn't worry about it too much and leave it as the WP default value.

PR #7175, opened 2023-09-21
Support leading zeroes for the number field

sc0ttkclark:
## Description

This PR adds support for the number field having leading zeroes even after formatting.

## Testing instructions

TBD

## Screenshots / screencast

TBD

TBD

## PR checklist


classes/fields/number.php
```diff
@@ -82,6 +82,16 @@ public function options() {
 				'type'        => 'boolean',
 				'excludes-on' => array( static::$type . '_decimals' => 0 ),
 			),
+			static::$type . '_keep_leading_zeroes' => array(
```
JoryHogeveen, line 85: I would choose a `_format` field instead of a specific field for this case just as we dit with the currency field.

https://github.com/pods-framework/pods/blob/3.0.2/classes/fields/currency.php#L100-L133

Note the decimals and decimal_handling fields below the format options.

JoryHogeveen (commented):
~~Would be a nice addition, however the naming could be more general as we did with the currency field.~~

~~What do you think? Or do you consider the number field a totally different case?~~

classes/fields/number.php
sc0ttkclark: Can you clarify what you mean here? do you mean use `_format` as part of the option name?

```suggestion
			static::$type . '_format_keep_leading_zeroes' => array(
```

sc0ttkclark: I think we could add this to currency field too, but I can't think of a legitimate use case for a currency with leading zeroes. Unless that's really a thing?

JoryHogeveen: What I meant is that the currency field already supports this and I think it's better to keep the code the same as these fields inherit from each other. In this case, the currency field inherits from the number field.

JoryHogeveen: Hold on, I totally misunderstood the option hear, I was thinking about trailing zero's, not leading...

JoryHogeveen (commented):
Would it be nice to set the maximum allowed leading zero's?
Or even a minimum amount of digits (excluding decimals)?

sc0ttkclark:
We could probably put more time into this to figure out how many zeroes to limit to or if padding should be done as well.

reedtsutton:
> Would it be nice to set the maximum allowed leading zero's? Or even a minimum amount of digits (excluding decimals)?

Is it possible to just make it so the zeros can be entered? User can enter 0123 or 00123 and it displays as entered?

JoryHogeveen:
@sc0ttkclark @reedtsutton Just an idea for a first solution.
An option to disable formatting all together and allow all numeric type formats with a simple regex?

So basically, within the "_format" selectbox we add an option "none" or "any" witch disabled all formats but only runs the value through a regex that filters out everything except for decimals (if enabled).

reedtsutton:
Hey guys any update on this one?

reedtsutton:
Hey guys wondering if there is going to be any movement on this?

sc0ttkclark:
We just need to sort out exactly what we want from this enhancement. I'll look at this in the upcoming release when I can.

reedtsutton:
There is actually a separate (but probably related?) issue where an entry of 0 in the currency field is being treated to be a null value (see screenshots #1 and #2).

This compounds when I try to pass the value to another post type, and the resulting field ends up being blank (screenshots #3 and #4).

1 - manual entry of 0 in currency
2 - still shows as empty
3 - post feed
4 - offer rebate is empty

Issue #2306, opened 2014-07-28
Non wp_ table prefix -> user relation by role broken


canonex:
Hi,

got two fresh installations of Wordpress.

If prefix table is the standard wp_ everything is working.

If prefix table is something else (ex. bwp_ or wpk_ ...):
->new field
->select relation
->select users in the correlation dropdown
->in the additional field options pick one ore more role to limit the available users

the  filtered list is always empty.

Thank you,
Rickyx

ps. this issue is somewhat related to http://pods.io/forums/topic/user-dropdown-by-role/ but this is a brand new installation and no other plugins are present.

Issue #1415, opened 2013-05-21
Multiple select relating to Users is empty when Relationship is limited by Role


jamesshedden:
Issue noticed using WP 3.5.1 & Pods 2.3.4, replicated after a reset/clean install.

Created additional field for Posts (Field Type: Relationship -> Users) for choosing multiple Users of a certain role (let's say Contributor) that I want relate to a Post.

I have created 3 Contributor roles to test. When there is no Role restriction, everything's fine, but list shows empty with necessary restriction.

See below:

![pods-contributors-issue](https://f.cloud.github.com/assets/883598/543556/d99444dc-c23d-11e2-88a3-554262469a38.jpg)

Thanks!

PR #7185, opened 2023-09-25
Use prefix instead of parsing our own

### LINKED ISSUE #7181: Relationship Field improperly connecting to data taken from the database when WordPress is in multiple network mode
### Description

In the **pods/classes/fields/picks.php** at line **2740**

Replace

```
$prefix = $wpdb->base_prefix;

if ( is_multisite() && ! is_main_site() ) {
    $prefix .= get_current_blog_id() . '_';
}

```
with

```
$prefix = $wpdb->prefix;

// if ( is_multisite() && ! is_main_site() ) {
//     $prefix .= get_current_blog_id() . '_';
// }
```

When we want to get the list of users on the current site, the result returns the list of the original site.

### Version

3.0.3

### Testing Instructions

_No response_

### Screenshots / Screencast

_No response_

### Possible Workaround

_No response_

### Site Health Information

_No response_

### Pods Package

_No response_

JoryHogeveen:
Enhance compatibility with other plugins.

## Related GitHub issue(s)

Fixes #7181

## PR checklist

### DIFF
--- a/classes/fields/pick.php
+++ b/classes/fields/pick.php
@@ -2860,13 +2860,7 @@ class PodsField_Pick extends PodsField {
 								continue;
 							}
 
-							$prefix = $wpdb->base_prefix;
-
-							if ( is_multisite() && ! is_main_site() ) {
-								$prefix .= get_current_blog_id() . '_';
-							}
-
-							$where[] = $prefix . 'capabilities.meta_value LIKE "%\"' . pods_sanitize_like( $role ) . '\"%"';
+							$where[] = $wpdb->prefix . 'capabilities.meta_value LIKE "%\"' . pods_sanitize_like( $role ) . '\"%"';
 						}
 
 						if ( ! empty( $where ) ) {


sc0ttkclark:
There was a reason this was in place, I need to go figure out why that was and how to continue supporting that (if it's still a valid reason).

sc0ttkclark:
Walking backwards from most recent to the original introduction of this usage:

* We switched to `$wpdb->base_prefix` from `$wpdb->prefix` in #2393 to fix subsite issues
* We introduced the `$wpdb->prefix` usage instead of hardcoding `wp_` in #2385 #2306
* We fixed usage so subsites got the current blog ID added to the prefix in #1415
* We introduced the ability to limit by user roles in ab1370d18e0c065adc08cccba40fa0e853f9c1f5

Given that, I think your PR looks good -- EXCEPT -- this is not what WP_User uses (as of WP 4.9+) and that's what we need to follow instead here.

```php
	/**
	 * Sets the site to operate on. Defaults to the current site.
	 *
	 * @since 4.9.0
	 *
	 * @global wpdb $wpdb WordPress database abstraction object.
	 *
	 * @param int $site_id Site ID to initialize user capabilities for. Default is the current site.
	 */
	public function for_site( $site_id = '' ) {
		global $wpdb;

		if ( ! empty( $site_id ) ) {
			$this->site_id = absint( $site_id );
		} else {
			$this->site_id = get_current_blog_id();
		}

		$this->cap_key = $wpdb->get_blog_prefix( $this->site_id ) . 'capabilities';

		$this->caps = $this->get_caps_data();

		$this->get_role_caps();
	}
```

So the solution here needs to actually be calling `$wpdb->get_blog_prefix()` instead.

sc0ttkclark:
Just did more digging into the `wpdb` class and it looks like when switching blogs, it accounts for everything automatically so the code that `WP_User` is using isn't necessary since our context doesn't need to be able to check by site ID -- only the current site context. We can merge this as-is.

sc0ttkclark (approved):

sc0ttkclark (approved):

Issue #7218, opened 2023-11-08
After setting a custom capability access is revoked for single install admins


JoryHogeveen:
### Description

Once you've saved a CPT Pod with a custom capability this capability will keep on being used, even after you switch back to the default Post or Pages capabilities.

If you set a custom capability then the whole access is revoked if you do not assign it to the admin role.
If you have set a custom capability and then select the Post again you will be able to list the CPT but not edit or anything.

The capability value keeps being saves in the CPT, even though it's not set on a custom capability anymore.
This isn't a problem but should be fixed when registering the CPT.

### Version

3.0.8

### Testing Instructions

1. Create CPT in a single WP installation
2. Select custom capability.
3. Notice you'll not be able to to view this CPT anymore since you do not have this capability.
4. Switch back to Post capability.
5. You are able to list but not edit the CPT.

### Possible Workaround

You can manually remove the capability input and it will work again.

sc0ttkclark:
Super admin checks are not utilized unless on multisite -- so even if you are an admin of a single site, you don’t automatically get access to any/all caps. That’s a quirk IMO.

We could perhaps do a check when they set a cap there to warn them that their role doesn’t have that cap yet, but I don’t think there’s anything we could do there besides possibly some workflow improvements like:

1. Notice: we could show a notice next load of the pod edit screen for a CPT/CT when cap is not set on the current user role
2. Notice: we could add HTML field that shows a notice below the cap setting that shows only if the value of that cap is not in the list of the current user role’s caps
3. Action: we could add checkbox field that shows a notice below the cap setting that shows only if the value of that cap is not in the list of the current user role’s caps and allow them to check it to auto-add the cap to the current user role

JoryHogeveen:
While your assessment here is totally valid, it's not directly related to the bug here so probably belongs in it's own issue (related to this one).

The issue here is that when I revert the capability to "post" this change isn't done. For some reason it still not accessible anymore and I cannot find out why/where this goes wrong. The register parameters seems correct.

sc0ttkclark:
I didn't realize resetting there wasn't working. Can you confirm that the value remains post on reload and that the pod config did change over?

JoryHogeveen:
Yes I already checked this, the change does seem to reflect the Pods register options but it looks like it's maybe modified afterwards based on the custom value?
The custom value isn't removed, it's just hidden after you select "post" instead of "custom". The custom value also keeps being saved in the Pod object.

[...]

PR #7250, opened 2024-01-30
Fix custom capability fallback

JoryHogeveen:
## Description

Currently if you leave the custom capability field empty it will fallback to `post` instead of the CPT name.
This PR fixes this :)

## Related GitHub issue(s)

Related #7218

## Testing instructions

1. Create CPT with custom cap
3. Leave custom cap empty
4. Switch to role that includes the custom edit caps (edit_cpt/edit_cpts) but doesn't have the default edit_post and edit_posts caps.
5. Try to edit CPT post
6. You can't! :(
7. Apply patch
8. You can! :)

## PR checklist

### DIFF
--- a/classes/PodsInit.php
+++ b/classes/PodsInit.php
@@ -1232,10 +1232,10 @@ class PodsInit {
 					$cpt_rewrite['slug'] = preg_replace( '/[^a-zA-Z0-9%\-_\/]/', '-', $cpt_rewrite['slug'] );
 				}
 
-				$capability_type = pods_v( 'capability_type', $post_type, 'post' );
+				$capability_type = pods_v( 'capability_type', $post_type, 'post', true );
 
 				if ( 'custom' === $capability_type ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
I believe this was intentional because all CPTs by default use 'post' as the capability base: https://developer.wordpress.org/reference/functions/register_post_type/#:~:text=the%20posts%20icon.-,capability_type,-string%7Carray

I'm going to keep this as it was before except add a strict check on it so that if it was saved as empty then should force 'post' no matter what there.

sc0ttkclark:
Oh wait I see this was `capability_type_custom`, carry on!

classes/PodsInit.php
```diff
@@ -1235,7 +1235,7 @@ public function setup_content_types( $force = false ) {
 				$capability_type = pods_v( 'capability_type', $post_type, 'post' );
```
sc0ttkclark, line 1235:
```suggestion
				$capability_type = pods_v( 'capability_type', $post_type, 'post', true );
```

sc0ttkclark (dismissed):

PR #7341, opened 2024-08-22
Improve REST authentication method when registering fields

### LINKED ISSUE #7340: Using REST API with application passwords not working with Pods fields (unless public)
### Description

When you enable REST API but set the fields are "private" (requires authentication) they will not show when using application passwords.

After testing I've found that during REST initialization (when Pods registers it's fields) the function `is_user_logged_in()` does not work for application passwords.
Only in the `get_callback` of a field will the application password user be authenticated. However, at that point the fields is already disabled.

### Version

3.2.6

### Testing Instructions

See description and it should be clear!
Add a field, enable REST API but only accessible for logged in users, test a call using an application password through (for example) Postman.

### Screenshots / Screencast

_No response_

### Possible Workaround

Non that I've found, other than enabling the fields publicly.

### Site Health Information

_No response_

### Pods Package

_No response_

JoryHogeveen:
Instead of checking if the user is logged in it also checks if the user a basic auth for applications is used.

## Description

Fixes #7340

@sc0ttkclark I'm not 100% sure this will be the best method. You could also already set the current WP user so WP core doesn't revalidate using the same method. I'll leave this up to your judgement.

## Testing instructions

See #7340

Try the same with this patch and see it now works!

## PR checklist

### DIFF
--- a/classes/PodsRESTFields.php
+++ b/classes/PodsRESTFields.php
@@ -27,6 +27,7 @@ class PodsRESTFields {
 
 	/**
 	 * The user ID for the authenticated user.
+	 *
 	 * @var int
 	 */
 	private static $rest_user_id;
--- a/classes/PodsRESTFields.php
+++ b/classes/PodsRESTFields.php
@@ -26,6 +26,12 @@ class PodsRESTFields {
 	protected $pod = null;
 
 	/**
+	 * The user ID for the authenticated user.
+	 * @var int
+	 */
+	private static $rest_user_id;
+
+	/**
 	 * Constructor for class
 	 *
 	 * @since 2.5.6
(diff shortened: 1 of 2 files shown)


classes/PodsRESTFields.php
```diff
@@ -25,6 +25,12 @@ class PodsRESTFields {
 	 */
 	protected $pod = null;
 
+	/**
+	 * The user ID for the authenticated user.
+	 * @var int
```
PodsBot, line 30: **[phpcs]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>`Generic.Commenting.DocComment.SpacingBeforeTags`<br>There must be exactly one blank line before the tags in a doc comment

sc0ttkclark (approved):

sc0ttkclark (approved):

PR #7369, opened 2024-10-15
Block Editor Form Validation

### LINKED ISSUE #7343: Form Validation: Post editor (block editor)
We need to prevent the block editor saving process when there are fields in an invalid state (required, bad format, etc).

- [x] Stop saving of a new post
- [x] Stop saving of a draft
- [x] Stop updating of a published post

Stop saving/updating means:

- [x] Do not save the post
- [x] Do not save the metaboxes

Bonus points:

- [ ] Show the error in the publish flow (when publishing a new post)
- [x] Show a notice on the screen about errors being highlighted in the form below (maybe link to the first metabox?)

JoryHogeveen:
## Description

This PR adds a useBlockEditor React Hook which will initialize by overriding the default savePost function of WordPress core. Based on the code by ACF. The order of overriding should not matter as we are returning the previous value if no validation errors occur, therefore we should be compatible with ACF and other plugins using the same method.

The hook is further used within the field wrapper component which handles validation. Through this component I can set the valid state and pass any error messages.

This PR will not add extra validation on the current system but will be compatible with any extras we add to the useValidation hook as it is merely using it's response messages.

It should create error messages at the top if you click save and will auto-trigger that message to show on the field itself as well.
After validation it will also disable the save button until you've made the required changes.

## Related GitHub issue(s)

Fixes #7343 

## Testing instructions

Just create a Pod with some required fields and test around by trying to save the post.

## PR checklist


ui/js/dfv/src/hooks/useBlockEditor.js
```diff
@@ -0,0 +1,77 @@
...
+
+			const pbe = window.PodsBlockEditor;
+
+			if ( ! Object.values( pbe.messages ).length ) {
+				// eslint-disable-next-line no-undef
+				return pbe.savePost.apply( this, arguments );
+			}
+
+			return new Promise( function( resolve, reject ) {
+				// Bail early if is autosave or preview.
```
JoryHogeveen, line 29: @sc0ttkclark Should we allow saving for a draft?

sc0ttkclark: I'm on the fence here. Maybe we can make that an option in settings or a constant/hook. Default to allow saves on drafts perhaps.

sc0ttkclark: But there's no way to draft by default when you have a post that's already published. So we'd need to distinguish between save and save draft.

ui/js/dfv/src/components/field-wrapper/index.js
```diff
@@ -192,6 +193,21 @@ export const FieldWrapper = ( props ) => {
...
+			blockEditor.unlockPostSaving( `pods-field-${ name }` );
+		} else {
+			blockEditor.lockPostSaving( `pods-field-${ name }`, validationMessages, () => setHasBlurred( true ) );
+		}
+
+		// Unlock on unmount.
+		return () => {
+			blockEditor.unlockPostSaving( `pods-field-${ name }` );
+		};
+	}, [ validationMessages ] );
```
JoryHogeveen, line 209: Todo: also add meetsConditionalLogic as a hook dependency

ui/js/dfv/src/hooks/useBlockEditor.js
```diff
@@ -0,0 +1,77 @@
+
+const useBlockEditor = () => {
+	const editorSelect = wp.data?.select( 'core/editor' );
+	const editorDispatch = wp.data?.dispatch( 'core/editor' );
+	const notices = wp.data?.dispatch( 'core/notices' );
+
+	// @todo Use hook instead of savePost override once stable.
+	if ( ! window.PodsBlockEditor && editorDispatch && editorDispatch.hasOwnProperty( 'savePost' ) ) {
+		// First init.
+		window.PodsBlockEditor = {
```
JoryHogeveen, line 10: Optionally make this a class with private props and getters/setters. Maybe even an external tool like the API instead of a React hook.

ui/js/dfv/src/components/field-wrapper/index.js
JoryHogeveen:
```suggestion
	}, [ validationMessages, meetsConditionalLogic ] );
```

ui/js/dfv/src/hooks/useBlockEditor.js
JoryHogeveen: Lets keep it like this since we should definitely change this code once a hook is available.
The added todo will inform any dev checking this code in that it's a temp solution.

sc0ttkclark (dismissed):

sc0ttkclark (approved):

PR #7548, opened 2026-06-04
Fix empty number fields saving as zero

### LINKED ISSUE #7110: Empty numeric fields not possible - forces 0
### Description

Hey, 
it seems to be #6226 bug is back for Plain Number/Currency.

I've test it with an field that was prefilled because of duplicating the page, but when emptying the field, its always "0".

### Version

2.9.18

### Testing Instructions

_No response_

### Screenshots / Screencast

pods_numeric_bug

### Possible Workaround

_No response_

### Site Health Information

_No response_

### Pods Package

_No response_

acjohnson08:
## Summary
Fixes #7110.

Preserves intentionally empty values in DFV number formatting/parsing instead of coercing them to `0`. This allows optional Plain Number fields to be cleared and saved blank.

## Details
`parseFloatWithPodsFormat()` previously converted an empty string to `0`, which caused cleared number inputs to repopulate/save as zero.

`formatNumberWithPodsFormat()` also formatted empty/null/undefined values as `"0"`.

This change keeps empty/null/undefined values empty while preserving real numeric zero values.

## Testing
- Applied the fix to a development WordPress site using Pods.
- Rebuilt the DFV bundle and confirmed clearing a previously filled Plain Number field saves and reloads as blank.
- Confirmed actual `0` remains valid.
- Ran focused Jest test:

```bash
npm run jest -- formatNumberWithPodsFormat
```

- Ran ESLint on changed files:

```bash
./node_modules/.bin/eslint ui/js/dfv/src/helpers/formatNumberWithPodsFormat.js ui/js/dfv/src/helpers/test/formatNumberWithPodsFormat.test.js
```

- Ran production webpack build:

```bash
npm run webpack-production
```

Note: full `npm run lint-js` currently reports pre-existing repo-wide lint issues outside this change.

### DIFF
--- a/ui/js/dfv/src/helpers/formatNumberWithPodsFormat.js
+++ b/ui/js/dfv/src/helpers/formatNumberWithPodsFormat.js
@@ -72,9 +72,9 @@ export const parseFloatWithPodsFormat = (
 	newValue,
 	format,
 ) => {
-	// Turn empty string to 0.
-	if ( '' === newValue ) {
-		return 0;
+	// Preserve intentionally empty values.
+	if ( '' === newValue || undefined === newValue || null === newValue ) {
+		return '';
 	}
 
 	// If we get a float value, we don't need to do anything.
@@ -99,9 +99,9 @@ export const formatNumberWithPodsFormat = (
 	decimalHandling = 'none',
 	decimals = 'auto',
 ) => {
-	// Skip empty strings or undefined.
+	// Preserve intentionally empty values.
 	if ( '' === newValue || undefined === newValue || null === newValue ) {
-		return '0';
+		return '';
 	}
 
 	const thousands = getThousandsSeparatorFromPodsFormat( format );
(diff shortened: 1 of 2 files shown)


JoryHogeveen:
Hi @acjohnson08 
Thanks for the PR!

I'm on holiday right now and will revisit this PR afterwards in more detail.

I'm unsure what kind of backwards compatibility issues this might introduce because I see you had to remove some tests for this PR to work as expected.

@sc0ttkclark what are your thoughts?

sc0ttkclark:
This introduces separate behavior for Pods that use table-based storage and those that use meta-based storage (default).

For table-based storage, the column for number fields is defined as `DECIMAL($length, $decimals)`

For meta-based storage, it uses the `meta_value` column of the meta table which ultimately is just a text column. It stores both numbers and text as text so when the form is saved with a blank string, it will save as a blank string. If the form is saved with a zero, it will save as the zero (not blank).

The work in https://github.com/pods-framework/pods/pull/7175 also complicates this PR but it brings up a good idea of making this PR's functionality optional. Something like a new option "Allow empty value" much like date/datetime/time fields have:

```php
			static::$type . '_allow_empty' => [
				'label'   => __( 'Allow empty value, not enforcing a zero', 'pods' ),
				'default' => 0,
				'type'    => 'boolean',
			],
```

With the option added and checking against the option then I'd be good with this returning the empty string. Whether the option should default on or off is up for debate, I'm not tied to one choice or the other there.

PR #7551, opened 2026-06-06
fix(forms): skip required validation for conditionally hidden fields

### LINKED ISSUE #7513: Required fields and logical functions are incompatible.
### Description

Suppose there are three fields:
original => true
original_author => null
original_url => null

Assuming they are all required fields, even if I set conditions under which `original_author` and `original_url` are not needed, they cannot be refreshed and saved successfully due to the required field setting. Is there a way to fix this issue?

If the fields are hidden, the required field check is unnecessary.

### Version

3.3.8

### Testing Instructions

_No response_

### Screenshots / Screencast

_No response_

### Possible Workaround

_No response_

### Site Health Information

```text

```

### Pods Package

```json

```

faisalahammad:
## Summary

`pods_form_get_visible_objects()` only checked the static `hidden` property and missed fields hidden by conditional logic. As a result `pods_form_validate_submitted_fields()` ran required checks on hidden fields, blocking form submission even though the user never saw them.

This change evaluates each field's conditional logic and skips fields whose rules resolve to hidden. It mirrors the existing filter behavior in `PodsAPI::save_pod_item()` so the standalone validator and the actual save path agree on which fields are active.

Fixes #7513

### `includes/forms.php` — `pods_form_get_visible_objects()`

**Before:** the function only filtered out fields where the static `hidden` option was true. Fields hidden by conditional logic passed through and were validated.

**After:** before iterating, the function builds a `$field_values` map from the current submission, applying the same permission and hidden checks used elsewhere. Each field is then checked with `Field::is_visible( $field_values )` and skipped if it resolves to hidden.

```php
// Collect current field values to evaluate conditional logic visibility.
$field_values = [];

foreach ( $groups as $group ) {
    if (
        $options['section']
        && $options['section_field']
        && (
            'any' === $options['section']
            || ! in_array( $options['section'], (array) $group[ $options['section_field'] ], true )
        )
    ) {
        continue;
    }

    if ( ! pods_permission( $group ) ) {
        continue;
    }

    $group_fields = $group->get_fields();

    if ( empty( $group_fields ) ) {
        continue;
    }

    foreach ( $group_fields as $group_field ) {
        if ( ! pods_permission( $group_field ) ) {
            continue;
        }

        if ( pods_v( 'hidden', $group_field, false ) ) {
            continue;
        }

        $field_values[ $group_field['name'] ] = pods_form_get_submitted_field_value( $group_field['name'] );
    }
}
```

```php
// Skip if the field is hidden by conditional logic.
if ( $field instanceof Field && ! $field->is_visible( $field_values ) ) {
    continue;
}
```

**Why:** the existing save path in `PodsAPI::save_pod_item()` already excludes conditionally hidden fields from the validation loop. The standalone `pods_form_validate_submitted_fields()` helper did not, so calling it on a form with required-but-hidden fields would return `WP_Error` even though the save itself would have succeeded. Applying the same visibility logic here keeps both paths consistent.

## Testing

Reproduce the original issue from #7513.

1. Create a Pod with three text fields, all marked as required.
2. Add conditional logic so that `original_author` and `original_url` are hidden when `original` is empty.
3. Add the Pod to a page using the `[pods-form]` shortcode.

**Test 1: submit with all three fields empty**
1. Load the form page
2. Submit without filling anything
3. Result: only `original` reports a validation error; the two hidden fields pass

**Test 2: submit with `original` filled, the other two empty**
1. Fill `original` so the conditional logic reveals the other two fields
2. Submit
3. Result: `original_author` and `original_url` now report required errors as expected

**Test 3: submit with all three filled**
1. Fill all three fields
2. Submit
3. Result: form saves successfully


faisalahammad:

PR #7556, opened 2026-06-12
fix(wpgraphql): preserve Pods field order in GraphQL connections

### LINKED ISSUE #7552: Pods → WPGraphQL relationship/file connections don't preserve Pods field order (fall back to default post_date DESC)
### Bug Description

When a Pods relationship (`pick`) or file/image gallery (`file`) field is exposed through WPGraphQL as a connection, the connected items do **not** come back in the order saved in the Pods admin UI (the relationship "weight" order). Instead they fall back to WordPress's default ordering, `post_date DESC`. The IDs are fetched in the correct order but the connection's `WP_Query` re-sorts them, so the saved ordering is lost.

For a gallery arranged in upload order (ascending `post_date`) — the common case — `post_date DESC` makes the result look like an exact reversal of the intended order, which is the symptom that usually surfaces this. When the field is arranged in any other order, the result is simply date-sorted rather than a clean reversal.

### Environment

- Pods: 3.3.9
- WPGraphQL: 2.15.0
- WordPress: 7.0
- PHP: 8.3.30

### Steps to reproduce

1. Create a Pods **File / Image / Video** (gallery) field — or a **Relationship** field — with WPGraphQL enabled and the GraphQL field format set to **connection**.
2. In the admin, attach several items and arrange them in a deliberate order (items A, B, C, D) that does **not** match their `post_date` ordering.
3. Query the connection via WPGraphQL:
   ```graphql
   {
     post(id: "123", idType: DATABASE_ID) {
       imageGallery {
         nodes { databaseId sourceUrl }
       }
     }
   }
   ```
4. Observe the order of `nodes`.

### Expected behavior

Items returned in the Pods-defined order (A, B, C, D).

### Actual behavior

Items returned sorted by WordPress's default `post_date DESC`, ignoring the Pods order. (For galleries kept in upload/ascending-date order this presents as an exact reversal, e.g. D, C, B, A.)

### Root cause

This is a timing issue between Pods and WPGraphQL's connection resolver.

WPGraphQL **already** preserves `post__in` order. In `WPGraphQL\Data\Connection\PostObjectConnectionResolver::prepare_query_args()` (v2.15.0, ~lines 257-286) it sets `orderby => 'post__in'` whenever `post__in` is present and the client didn't request an `orderby`.

But `prepare_query_args()` runs **inside the resolver's constructor**, and Pods sets `post__in` **after** constructing the resolver, in `src/Pods/Integrations/WPGraphQL/Connection.php` → `get_connection_resolver()`:

```php
// case 'attachment' / 'media'  (~line 198-202)
$connection = new PostObjectConnectionResolver( $source, $args, $context, $info, 'attachment' );
$connection->set_query_arg( 'post__in', $field_value );   // runs AFTER prepare_query_args()
return $connection;
// (same pattern for case 'post_type', ~line 168-171)
```

So when WPGraphQL checks for `post__in`, it isn't there yet; Pods injects it a moment later via `set_query_arg()`, by which point `orderby` has already defaulted. The connection's `WP_Query` then runs with the correct `post__in` but the default `post_date DESC` ordering, discarding the Pods order.

### Suggested fix

When Pods sets `post__in`, also set the matching `orderby` so the order survives — mirroring what WPGraphQL does natively:

```php
$connection->set_query_arg( 'post__in', $field_value );
$connection->set_query_arg( 'orderby', 'post__in' );
```

The same "preserve the order I supplied" need applies to the other branches that pass an explicit ID set (`include`, `comment__in`, etc.).

### Workaround (no plugin edits)

Because the IDs are injected after the connection-args filters run, the order can't be fixed from `graphql_post_object_connection_query_args` / `graphql_connection_query_args` (`post__in` isn't present there yet). It can be restored at query time via `pre_get_posts`, scoped to WPGraphQL connection queries that carry an explicit `post__in` with no requested order (WPGraphQL's own `in` connections already set `orderby`, so they're untouched):

```php
add_action( 'pre_get_posts', function ( $query ) {
    if ( ! isset( $query->query_vars['graphql_args'] ) ) {
        return; // WPGraphQL connection queries only
    }
    $post_in = $query->query_vars['post__in'] ?? [];
    $orderby = $query->query_vars['orderby'] ?? '';
    if ( ! empty( $post_in ) && empty( $orderby ) ) {
        $query->set( 'orderby', 'post__in' );
    }
} );
```

faisalahammad:
## Summary

Pods relationship and file/image gallery fields exposed through WPGraphQL as connections now return nodes in the saved Pods order instead of falling back to WordPress default `post_date DESC`. This fixes the issue where gallery images attached in upload order appear reversed, and relationship items lose their admin-arranged order.

Fixes #7552

### `src/Pods/Integrations/WPGraphQL/Connection.php`

**Before:** Pods injects `post__in` via `set_query_arg()` after the resolver constructor runs, so WPGraphQL's `prepare_query_args()` sees no `post__in` and leaves `orderby` at the default `post_date DESC`.

**After:** Each of 7 standard-resolver branches that pass an explicit ID set now also sets a matching `orderby`:

```php
// post_type / attachment / media / menu_item
$connection->set_query_arg( 'post__in', $field_value );
if ( empty( $args['orderby'] ) ) {
    $connection->set_query_arg( 'orderby', 'post__in' );
}

// taxonomy / user — use include
$connection->set_query_arg( 'include', $field_value );
if ( empty( $args['orderby'] ) ) {
    $connection->set_query_arg( 'orderby', 'include' );
}

// taxonomy_object — use in
$connection->set_query_arg( 'in', $field_value );
if ( empty( $args['orderby'] ) ) {
    $connection->set_query_arg( 'orderby', 'in' );
}

// comment — use comment__in
$connection->set_query_arg( 'comment__in', $field_value );
if ( empty( $args['orderby'] ) ) {
    $connection->set_query_arg( 'orderby', 'comment__in' );
}
```

**Why:** Mirrors WPGraphQL's native behavior for `in:` connections where `orderby` is set before `post__in`. The `empty( $args['orderby'] )` guard ensures client-supplied ordering on the GraphQL query still takes precedence.

The `pod` and `pod_type` branches now pass `$field_value` to the custom resolver constructors so those resolvers can reorder by input order.

### `src/Pods/Integrations/WPGraphQL/Connection_Resolver/Pod.php`

Added `$ordered_ids` property and optional 6th constructor arg (`array $ordered_ids = []`). The `get_ids()` method now reorders its result to match the input order:

```php
public function get_ids() {
    $ids = [];
    if ( ! $this->query ) {
        return $ids;
    }
    while ( $this->query->fetch() ) {
        $id = $this->query->id();
        $ids[ $id ] = $id;
    }
    // Reorder to match the Pods-saved order supplied by the caller.
    if ( ! empty( $this->ordered_ids ) ) {
        $ordered = [];
        foreach ( $this->ordered_ids as $ordered_id ) {
            $ordered_id = (int) $ordered_id;
            if ( isset( $ids[ $ordered_id ] ) ) {
                $ordered[ $ordered_id ] = $ordered_id;
            }
        }
        // Append any IDs the query returned that weren't in the supplied order.
        if ( count( $ordered ) < count( $ids ) ) {
            foreach ( $ids as $id ) {
                if ( ! isset( $ordered[ $id ] ) ) {
                    $ordered[ $id ] = $id;
                }
            }
        }
        $ids = $ordered;
    }
    return $ids;
}
```

**Why:** The `Pod` resolver uses `Pods::find()` (not `WP_Query`), so the WPGraphQL `orderby` workaround doesn't apply. Reordering in `get_ids()` achieves the same result.

### `src/Pods/Integrations/WPGraphQL/Connection_Resolver/Pod_Type.php`

Same pattern as `Pod.php` — accepts `$ordered_ids` as optional 5th constructor arg, reorders in `get_ids()`.

## Testing

**Test 1: File/image gallery order**

1. Create a post-type pod with a File/Image/Video gallery field exposed as a GraphQL connection.
2. Attach images and reorder them in admin so the order differs from upload date.
3. Query the connection via WPGraphQL.
4. Result: nodes come back in the admin-arranged order, not `post_date DESC`.

**Test 2: Relationship field order**

1. Create a Pod with a Relationship field exposed as a GraphQL connection.
2. Attach items in a deliberate non-date order.
3. Query the relationship connection.
4. Result: nodes match the admin order.

**Test 3: Client-supplied `orderby` still works**

1. Query the same relationship with an explicit `orderby: { field: TITLE }`.
2. Result: nodes are sorted by title, not by the saved Pods order.

**Test 4: Table-storage Pod-to-Pod relationships**

1. Two pods with table storage, one has a pick field to the other as a connection.
2. Attach items in deliberate order and query.
3. Result: nodes follow the saved order.


faisalahammad:

PR #7557, opened 2026-06-12
fix(core): preserve <e> error marker in pods_error AJAX die

### LINKED ISSUE #7538: pods_error no longer catched by submit ajax call
### Description

I used to validate some field with the pre_save filter and pods_error function.
Error was displayed with an alert and submit stopped.
Since last update, It doesn't work anymore.

File general.php
`function pods_error( $error, $obj = null ) {
...
die( wp_kses_post( sprintf( '<e>%s</e>', $error ) ) );
...

`
Ajax call in jquery.pods.js search for the pattern '<e>' to define if it's a modal ?
And never reach this code section : 
`alert( 'Error: ' + err_msg );`

wp_kses_post convert <e> and wasn't use before : 

`die( sprintf( '<e>%s</e>', $error ) );`

### Version

3.3.8

### Testing Instructions

1 - add a pre_save filter
2 - call pods_error function
3 - no alert, form is submitted

### Screenshots / Screencast

_No response_

### Possible Workaround

Dirty workaround, use in your pre_save filter :

`die(sprintf( '<e>%s</e>',"My error messsage"));`

Maybe there is a proper way to validate field and display errors. But no documentation or tutorials on this.

### Site Health Information

```text

```

### Pods Package

```json

```

faisalahammad:
## Summary

The `<e>...</e>` error marker that the JS admin form, upgrade wizard, and plupload file uploader rely on to detect `pods_error()` AJAX failures was being stripped by an outer `wp_kses_post()` wrap. The body of `$error` is already sanitized earlier in the function, so the outer wrap was redundant and harmful. Removing it restores the marker and the consumer error path.

Fixes #7538

### `includes/general.php` — `pods_error()`, `'exit'` mode, AJAX branch

**Before:**
```php
} else {
    die( wp_kses_post( sprintf( '<e>%s</e>', $error ) ) );
}
```

**After:**
```php
} else {
    /*
     * The <e>...</e> marker is part of a contract with the JS admin form
     * (ui/js/jquery.pods.js, ui/js/jquery.pods.upgrade.js, and the plupload
     * file uploader) — they detect the marker to short-circuit the AJAX
     * success path and surface the error. Do not run the body through
     * wp_kses_post() here, as it would strip the unknown <e> tag.
     *
     * The error string itself is already sanitized earlier in this function
     * (see the non-array and array branches above).
     */
    die( sprintf( '<e>%s</e>', $error ) );
}
```

**Why:** the outer `wp_kses_post()` was added by an earlier Plugin Checker fixes commit. Since `<e>` is not in the allowed-post tag list, it gets stripped before reaching the client, breaking the JS consumers. The non-AJAX `wp_die()` branch above is unchanged — it does not use the marker and is correct for that context.

## Testing

**Test 1: `pre_save` validation error via AJAX submit (original repro)**

1. Add a `pre_save` filter that calls `pods_error( 'My validation error', $obj )` with `display_errors = true`.
2. Open the edit screen for an item in that Pod.
3. Submit the form.

Result: JS alert shows `Error: My validation error`. Submit aborts. Form stays on screen with the entered values. Before the fix, the modal path posts a `PODS_MESSAGE` to the parent and the form reloads with no feedback.

**Test 2: File uploader error path**

1. Trigger a `pods_error()` exit on a file upload.
2. Upload a file in a Pods form.

Result: the uploader surfaces the error message. Before the fix, the uploader silently fails.

**Test 3: Unit tests**

`npm run jest` — 510 tests pass.

`php -l includes/general.php` — no syntax errors.

CodeRabbit CLI review on the diff — 0 actionable findings.


faisalahammad:

PR #7560, opened 2026-06-24
fix(blocks): add offset option to Pods Item List and Related Item List blocks

### LINKED ISSUE #7523: In the item list block, allow the LIMIT option OFFSET
### Problem to Solve

I have in the same page different pod post of the same type in different locations. It would be very nice to have the OFFSET option to choose which.

Thanks

### Proposed Solution

In the same field where you set the LIMIT, you can write 1, 3 (coma no accepted now) or 1 OFFSET 3.
Or have another field for the OFFSET.

Thanks again

### Possible Workaround

_No response_

### Examples Elsewhere

_No response_

### Screenshots / Screencast

_No response_

faisalahammad:
## Summary
Adds an Offset field to the Pods Item List and Related Item List blocks, allowing users to skip a number of items when listing the same pod in multiple places. The underlying `[pods]` shortcode and data layer already support offset, so this only exposes the existing parameter through the block UI.

Fixes #7523

### Item List block
**Before:**
No offset field; users could only set a limit.

**After:**
Added `offset` field (type number, default 0) in `src/Pods/Blocks/Types/Item_List.php` and mapped it in the shortcode transform. Added `offset` attribute to `ui/js/blocks/block-types/pods-block-list/block.json`.

**Why:** Lets users skip N items before rendering the list without writing a custom shortcode.

### Related Item List block
**Before:**
No offset field.

**After:**
Added `offset` field (type number, default 0) in `src/Pods/Blocks/Types/Related_Item_List.php`.

**Why:** Same list block family, uses the same shortcode path and limit field.

## Testing
**Test 1: Item List block with offset**
1. Add Pods Item List block, choose pod/template, set Limit 2 and Offset 1.
2. View page.
Result: shows items 2-3, skipping item 1.

**Test 2: Related Item List block with offset**
1. Add Pods Related Item List block, choose related field/template, set Limit 2 and Offset 1.
2. View page.
Result: skips first related item and shows next 2.

**Test 3: Shortcode transform**
1. Insert `[pods name="my_pod" limit="2" offset="3" template="my_template"]`.
2. Transform to Pods Item List block.
Result: Offset field shows 3.

Also ran `php -l` on changed files, validated JSON, and ran `npm run build-production` successfully. JS lint `npm run lint-js` fails due to a pre-existing missing parent eslint config (`wordpress`), not from this change.


faisalahammad:

PR #7569, opened 2026-07-24
fix(import): skip old_name on package import, fix count_fields undercount

### LINKED ISSUE #7133: pods.json: incorrect field count with PHP & obscure error message with UI
### Description

Attempting to load the attached `pods.json` in various forms results in either incorrect field counts or not all fields loading:

```php
<?php

add_action(
    'plugins_loaded',
    function(){
        $file = __DIR__ . '/pods-old-names-removed-success-with-ui.json';
        pods_register_config_file( $file, 'json' );
    }
);
```

- [ ] When attempting to load `pods-original.json` via UI, any fields with `"old_name"` set displayed the following error:

Screenshot 2023-08-18 at 8 49 55 PM

- [ ] The above error displayed 23 different variations. Only one is shared above, but it seemed to output the error only per-group then give up. So the file was manually edited to remove `"old_name"` values to get the errors to go away.
- [ ] The error message doesn't give any reason for why the field wasn't imported.
- [ ] Clearing all the `"old_name"` values resulted in a successful import via UI using `pods-old-names-removed-success-with-ui.json`. There is still an odd `null,` on line 17376 of this file, but it didn't seem to cause a problem.

The successful import looks like this:
 (`66 pods, 66 groups, 825 fields`)

Screenshot 2023-08-18 at 8 55 16 PM

Attempting to load the same file to a fresh database using the PHP code at the beginning of this comment results in `60` or `61 fields`:

Screenshot 2023-08-18 at 8 53 38 PM

In testing, it didn't resolve the issue to re-export the file or change the max depth of `json_decode()`. The JSON appears valid, not throwing any errors from PHP when the throw-errors flag was enabled or running through an online linter.

Three JSON files attached: The original [provided in Slack](https://podswp.slack.com/archives/C02SVLHQF/p1692300949266769?thread_ts=1692293162.592349&cid=C02SVLHQF), a manually edited version which imported successfully via UI, but not via code, and a re-export of that imported config which also would not import via code. Some files had Windows line endings. Changing them to Unix didn't fix anything.

[three-json-files.zip](https://github.com/pods-framework/pods/files/12385008/three-json-files.zip)

### Version

2.9.19

### Testing Instructions

1. Start with a fresh install.
2. Backup the database.
3. Import one of the files via the import user interface. Get success with the revised file, but not the original. See obscure errors for the original.
4. Reset the database.
5. Load any of the JSON files with code.
6. Get an incorrect field count.

### Screenshots / Screencast

Below is what the "Settings" menu looks like on the successful import. On the programmatic import, these menus seem to appear with many fields as well, so `it seems it may just be the field counts that are incorrect when using PHP. Will link for feedback from the OP.`

Screenshot 2023-08-18 at 9 20 31 PM

### Possible Workaround

- Removing "old_name" from JSON allowed import to work via UI.
- Clearing Pods cache didn't update the field count.

### Site Health Information

```text
Testing was done in a fresh localwp.com install on PHP 8.1.9.
```

### Pods Package

```json
See .zip attached to description.
```

faisalahammad:
## Description

Two unrelated bugs prevent a clean pods.json package import:

- Fields with `"old_name"` values trigger an obscure error ("Cannot save the X field") and the import aborts partway through.
- A pod registered via `pods_register_config_file()` reports 1 field per pod instead of the actual count.

The import pipeline now strips `old_name` from field and pod data before delegating to `save_field()` / `save_pod()`. The `old_name` lookup in `save_field()` is correct rename semantics for in-place renames; it doesn't belong in the import path that matches objects purely by name.

## Related GitHub issue(s)

Fixes #7133

## Testing instructions

1. On a fresh install, go to Pods Admin → Migrate Packages → Import.
2. Paste the following package JSON:
   ```json
   {
     "@meta": { "version": "3.3.9" },
     "type": "package",
     "pods": [
       {
         "name": "test-rename",
         "label": "Test Rename",
         "type": "post_type",
         "storage": "meta",
         "groups": [
           {
             "name": "main",
             "label": "Main",
             "fields": [
               { "name": "first", "label": "First", "type": "text", "old_name": "previous-first" },
               { "name": "second", "label": "Second", "type": "text", "old_name": "previous-second" }
             ]
           }
         ]
       }
     ]
   }
   ```
3. Before fix: per-group "Cannot save the first field" errors, then crash.
4. After fix: pod "Test Rename" imports cleanly with both text fields visible.

For the register-config-file path:

1. Place a pods.json in your theme directory and add to functions.php:
   ```php
   add_action( 'plugins_loaded', function() {
       pods_register_config_file( get_template_directory() . '/pods.json', 'json' );
   } );
   ```
2. Reload admin.
3. Bug 2: in the Pods admin All Pods table, each pod reports "1" field. After fix: each pod reports its actual field count.

## Screenshots / screencast

Not UI-visible.

Bug: pods.json import no longer drops fields with `old_name` values, and `pods_register_config_file()` now reports correct per-pod field counts. #7133 (@faisalahammad)

## PR checklist


faisalahammad:

PR #7570, opened 2026-07-24
fix(general): rename pods_field_display first arg to $type and accept Pods object

### LINKED ISSUE #7293: Fix parameter name on pods_field_display
### Description

The first parameter of `pods_field_display` is a `$pod`, but if you actually give it a pod object it doesn't work. Looking into this function, it passes this parameter to https://docs.pods.io/code/pods/ as the `$type` parameter. Therefore, the first parameter of `pods_field_display` should probably match `pods()` and be named `$type`.

### Version

all

### Testing Instructions

_No response_

### Screenshots / Screencast

_No response_

### Possible Workaround

_No response_

### Site Health Information

_No response_

### Pods Package

_No response_

faisalahammad:
## Description

`pods_field_display` accepts a pod identifier or name as its first argument
but ships with two mismatches against the rest of the API:

1. The parameter was named `$pod`, but the value is passed straight to
   `pods()` whose first parameter is `$type`. That mismatch made the
   function harder to reason about and to call consistently with `pods()`.
2. The function documentation describes the first argument as the pod,
   suggesting a `Pods` object should be accepted. The implementation passed
   whatever value it received to `pods()` without checking, and
   `Pods::__construct` does not translate a `Pods` instance to a string,
   so a passed `Pods` object silently produced a null pod.

Rename the parameter to `$type` so the signature matches `pods()`, and
add an early short-circuit when a `Pods` object is passed in. Existing
positional callers that pass a pod name or id continue to work unchanged.

## Related GitHub issue(s)

Fixes #7293

## Testing instructions

1. Back up your site and database.
2. Replace the `pods/` plugin directory with the contents of the build zip
   attached to this PR, or upload via the Plugins screen.
3. Activate the Pods plugin.
4. In any Pods template that uses `pods_field_display`, confirm the field
   display value still renders as before.
5. In a template, pass a `Pods` object directly:
   `php`
   echo pods_field_display( pods( 'my_pod', get_the_ID() ), null, 'some_field' );
   ``
   Before this fix this returned an empty value. After this fix it returns
   the field display value.
6. Inside The Loop, confirm the field-name shortcut still works:
   `pods_field_display( 'some_field' )` returns the display value.

Automated wpunit tests live at
`tests/codeception/wpunit/functions/FieldDisplayTest.php` and cover the
string pod name path, the `Pods` object path, the field-name shortcut, an
unknown pod name, an empty field, and a `WP_Post` first argument.

```
slic run wpunit --group=pods-functions-field-display --ext DotReporter
```

## Screenshots / screencast

Not applicable: this is a behavior fix with no UI change.

Bug: `pods_field_display` now accepts a `Pods` object as the first argument and the parameter name matches `pods()`. #7293 (@faisalahammad)

## PR checklist


faisalahammad:

PR #7574, opened 2026-07-26
feat(theme): rewrite WP_Query meta_query to JOIN pod table for table-based fields

### LINKED ISSUE #7280: WP_Query integration for table-based storage
Hook into WP_Query for meta_query and other references to pod fields to support traversal and automatically handle JOINs and proper field aliases for the corresponding fields.

This would allow for supporting fields on Post Types that have table-based storage.

### LINKED ISSUE #7611: WP_Query_Integration: LIKE comparisons skip esc_like() and add no wildcards
## Description

`prepare_meta_query_value()` in `src/Pods/Theme/WP_Query_Integration.php` prepares values with `$wpdb->prepare()` only:

```php
return $is_string ? $wpdb->prepare( '%s', (string) $value ) : $wpdb->prepare( '%d', (int) $value );
```

`WP_Meta_Query` treats LIKE values differently — it wraps them:

```php
'%' . $wpdb->esc_like( $meta_value ) . '%'
```

## Impact

Two distinct problems:

1. **`%` and `_` are not escaped.** A value containing either is interpreted as a SQL wildcard rather than a literal. A search for `50%` matches far more than intended. This is not an injection (the value is still parameterised) but it is incorrect matching.

2. **No surrounding wildcards.** `LIKE 'foo'` here is an exact match, whereas the same clause against postmeta is a substring match. The identical `meta_query` therefore returns different results depending on whether the field happens to use table storage — which is exactly the transparency this integration is meant to provide.

`REGEXP` / `NOT REGEXP` go through the same `prepare_meta_query_value()` path and should be reviewed alongside it.

## Suggested fix

Mirror `WP_Meta_Query`: apply `$wpdb->esc_like()` and add the surrounding `%` wildcards for `LIKE` / `NOT LIKE`, so behaviour matches postmeta.

## Related

- #7574 — PR that introduced this integration.

_Disclosure: this issue was researched and drafted with AI assistance and reviewed by me before filing._

faisalahammad:
## Description

AI disclosure: this PR was written with Claude assistance, reviewed and tested by me.

Adds WP_Query integration so that meta_query clauses naming a table-based pod field are rewritten into LEFT JOINs against `{prefix}pods_{pod}` (plus `{prefix}podsrel` for relationship fields) instead of going through `wp_postmeta`.

Two new callbacks on the existing `Pods\Theme\WP_Query_Integration` singleton:

- `pre_get_posts` (priority 20) walks the meta_query, strips Pods' optional `_pods_`/`pods_` alias prefix on each key, and resolves the key against the fields of every table-based Pod attached to the queried post_type. Matching keys are stashed on the query object and removed from meta_query so WP_Meta_Query never builds postmeta joins/where for them.
- `posts_clauses` (priority 20) splices LEFT JOIN against `{prefix}pods_{pod}` for each spec, plus a LEFT JOIN against `{prefix}podsrel` for relationship fields. WHERE clauses mirror WP_Meta_Query's compare allow-list and sanitize values via `$wpdb->prepare()`.

Behavior: WP_Query with meta_query that names a table-based pod field returns the same IDs as the equivalent `pods()` helper call. Meta-based Pods and non-Pod queries are untouched (early-return in `pre_get_posts`).

### Commit `fix: escape LIKE values and add wildcards for table field queries` (Fixes #7611)

`WP_Meta_Query` wraps LIKE values as `'%' . esc_like( $value ) . '%'` before binding. The integration bound the raw value, so a LIKE on a table field was an exact match and any `%` or `_` in the value acted as a SQL wildcard. The same meta_query could return different rows depending on storage type. This commit mirrors the core esc_like + wildcard wrapping for `LIKE`/`NOT LIKE`. REGEXP/NOT REGEXP are unchanged (core uses plain `%s`). An array LIKE value has no single literal meaning, so the clause is skipped. Backward compatible.

## Related GitHub issue(s)

- Fixes #7280
- Fixes #7611

## Testing instructions

Requires the slic WordPress harness (per `CLAUDE.md`).

1. Activate the plugin from this branch.
2. Create a Post Type Pod (`book`) with **Table Based** storage.
3. Add two fields: `subtitle` (Text) and `related_book` (Pick, single, self-referencing).
4. Create three posts with `subtitle` values Alpha, Beta, Gamma.
5. Run a smoke query:

   ```php
   $q = new WP_Query( [
       'post_type'  => 'book',
       'meta_query' => [ [ 'key' => 'subtitle', 'value' => 'Beta' ] ],
       'fields'     => 'ids',
   ] );
   var_dump( $q->posts );
   ```

   Expected: array containing exactly the post ID whose `subtitle` is `Beta`. Without the fix the array is empty.

Additional manual checks for the LIKE fix (`#7611`): substring LIKE `subtitle LIKE 'pecan'` matches `Butter Pecan`; literal `%`/`_` in the value are escaped (e.g. value `50%` does not act as a wildcard).

Automated regression coverage in the new wpunit group:

```bash
slic run wpunit --group pods-wp-query-table-meta --ext DotReporter
```

Test cases:

- `test_meta_query_by_table_field_returns_same_ids_as_pods_helper`
- `test_pods_prefixed_meta_key_resolves_to_table_field`
- `test_mixed_query_pod_field_and_foreign_meta`
- `test_relationship_field_meta_query_uses_podsrel_join`
- `test_suppress_filters_short_circuits_rewrite`
- `test_no_table_meta_keys_means_no_join_added`
- `test_like_value_gets_wildcards_and_esc_like` (#7611: plain, `%`, `_`, NOT LIKE)
- `test_regexp_value_has_no_added_wildcards` (#7611: REGEXP stays plain)
- `test_like_with_array_value_bails_closed` (#7611: array value skipped)

Testing: verified by the existing `pods-wp-query-table-meta` group. For #7611 the wpunit harness cannot run in this env (no WordPress/DB), so LIKE/regexp coverage was verified by lint + phpstan on both files (clean) and by parallel inspection sub-agents (A: issue + core parity, B: call sites + deprecated APIs); no findings remain to fix on the #7611 lines.

Tested manually: confirmed this resolves the reported issue.

Backward compatible: yes, no breaking changes.

## Screenshots / screencast

Not applicable (no UI changes).

Feature: WP_Query meta_query now resolves table-based pod field keys via JOIN against the pod table and podsrel, so standard `$query->get_posts()` calls return the same IDs as the `pods()` helper. Backward compatible with meta-based Pods. #7280 (@faisalahammad)
Fix: LIKE / NOT LIKE on table-based pod fields now uses `esc_like()` and `%` wildcards, so the same meta_query returns the same rows regardless of storage type and values containing `%` / `_` are treated literally. Fixes #7611 (@faisalahammad)

## PR checklist


faisalahammad:

PR #7576, opened 2026-07-26
fix(form): add labels shortcode attribute

### LINKED ISSUE #7243: User Profile Form: Label options
### Problem to Solve

Related do #7242 , I'm working on pt_pt and naturally the custom field labels are correct, but is there any way to relabel "user_nicename", "display_name" and other internal fields on a shortcode [pods-form]?

### Proposed Solution

Since the fields already exist, they can't be recreated on Pods. A "labels" parameter should fix it, but since custom fields are already labeled it doesn't make much sense. Can't find anything on Polylang, and I really don't want to translate a site - just some labels.

### Possible Workaround

_No response_

### Examples Elsewhere

_No response_

### Screenshots / Screencast

_No response_

faisalahammad:
## Description

Add a `labels` attribute to the `[pods-form]` (and `[pods form=1]`) shortcode so per-field labels can be overridden from the shortcode itself. This is useful for built-in user fields like `user_nicename`, `display_name`, and `user_email` where labels are hardcoded in English and site translation tools do not pick them up.

Usage:

```
[pods-form name="user" fields="user_nicename,display_name,user_email" labels="user_nicename:Utilizador,display_name:Nome Completo,user_email:Correio Eletronico"]
```

The attribute uses `:` as field:label separator and `,` as pair separator. Label values are sanitized with `sanitize_text_field()`.

## Related GitHub issue(s)

Fixes #7243

## Testing instructions

1. Create a WordPress page containing `[pods-form name="user" fields="user_nicename,display_name" labels="user_nicename:Nickname,display_name:Full Name"]`
2. View the page and confirm the form shows "Nickname" and "Full Name" instead of the default English labels
3. Create a second shortcode without `labels` on the same page and confirm default labels still show (no leak)
4. Insert a Pods Form block in Gutenberg editor and verify it still renders without error

## Screenshots / screencast

N/A — no visible UI changes beyond shortcode output labels.

Feature: You can now override per-field labels in the [pods-form] shortcode using the labels attribute. #7243 (@faisalahammad)

## PR checklist


faisalahammad:

faisalahammad:
## Review: the override works, but the new entry point is untested

The mechanism is sound. `ui/forms/list-row.php` calls `PodsForm::output_label( $field['name'], ... )` with the **unprefixed** field name — the `name_prefix` is applied later inside `label()` at `classes/PodsForm.php:117` — so the array key matches and the override genuinely reaches the render path.

The existing tests are real too: `PodsFormLabelsTest.php` follows the convention (`namespace Pods_Unit_Tests\...`, extends `Pods_UnitTestCase`, no `WPBrowser` type hint) and would fatal on unpatched code.

Three things to tighten.

### 1. Nothing tests `pods_shortcode_parse_labels()`

That's the actual new user-facing surface — the `field:Label,field2:Label2` parsing, the `unset( $tags['labels'] )`, and the `finally` cleanup — and it's the part most likely to regress. The tests only exercise the static directly. Worth covering:

- a `labels` string parses into the expected map
- malformed input (missing `:`, trailing `,`, a label containing `:`) degrades safely
- `labels` is removed from `$tags` so it doesn't reach the query
- the static is cleared after `pods_shortcode()` returns, including when it throws

### 2. The block gains an attribute with no control

`labels` is added to the `attributes` maps in `src/Pods/Blocks/Types/Form.php` (both places) but not to `Form::get_fields()`, which holds the editor controls (e.g. `'name' => 'label'` at `:198`). So the Gutenberg Pods Form block accepts the attribute but offers no way to set it.

### 3. Redundant double-parse

`pods_shortcode_form()` calls `pods_shortcode_parse_labels( $tags )`, then `pods_shortcode()` calls it again. Harmless only because the first call `unset()`s the key — worth dropping one.

### Design note

`apply_filters( 'pods_form_ui_label_text', $label, $name, $help, $options )` already exists at `classes/PodsForm.php:132` and does this job without new global static state. If the shortcode attribute is the goal, routing it through that filter for the duration of the render would avoid adding a static that has to be cleaned up in a `finally`.

PR #7578, opened 2026-07-26
fix(admin): add custom capability option for settings pages

### LINKED ISSUE #7245: Custom Settings Pages should support specifying a capability for access
### Problem to Solve

Right now, you must have access as a Pods admin (caps: `pods`, `pods_content`, or `pods_edit_{pod_name}`).

It would be great to override what capability check(s) are used for the Custom Settings Pages specifically.

### Proposed Solution

Add a new option to the Pod edit screen for Settings pods so that you can choose from these capabilities:

* Default: Anyone with access to pods, pods_content, or pods_edit_{pod_name} capabilities
* Custom
 * Specify the custom capability to reference for access

### Possible Workaround

```php
add_filter( 'pods_is_admin', 'my_custom_override_for_custom_settings_page_access', 10, 3 );

function my_custom_override_for_custom_settings_page_access( $has_access, $capabilities, $capability ) {
	$my_pod_name = 'my_pod_name_here';
	$my_custom_capability = 'my_custom_capability_here';

	if (
		in_array( 'pods_edit_' . $my_pod_name, $capabilities, true )
		&& current_user_can( $my_custom_capability )
	) {
		return true;
	}

	return $has_access;
}
```

### Examples Elsewhere

_No response_

### Screenshots / Screencast

_No response_

faisalahammad:
## Description

Settings pods now expose a per-pod capability option. The admin menu stays gated on `pods`, `pods_content`, or `pods_edit_{name}` by default. Picking "Custom" adds an editable cap name (e.g. `manage_options`) that grants access in the same OR-list, so existing admins keep their access and `pods_is_admin` filter still works.

## Related GitHub issue(s)

Fixes #7245

## Testing instructions

1. Go to Pods Admin.
2. Open an existing settings pod or create a new one (name `site_settings` for the test below).
3. Open the Admin UI tab.
4. New field "User Capability" — choose "Custom Capability".
5. In the new "Custom User Capability" input, type `manage_options` (or any cap your test user has).
6. Save pod.
7. As a user with `manage_options` but none of the Pods caps: the settings page menu item shows.
8. As a user without `manage_options` or any Pods cap: the menu item stays hidden.
9. Set "User Capability" back to "Default", save: original three-cap check restored.

## Screenshots / screencast

n/a

* Enhancement: Settings pods now support specifying a custom capability to gate access from the admin menu. #7245 (@faisalahammad)

## PR checklist

- [x] My code follows the WordPress Coding Standards.
- [x] My code follows the WordPress Inline Documentation Standards.

## Screenshot
image


faisalahammad:

faisalahammad:
## Review: the capability is only enforced on menu registration

No privilege escalation on the default path — the custom cap is only appended when `capability_type === 'custom'` *and* the value is non-empty, so the existing three-cap check at `classes/PodsAdmin.php:484` is preserved for everyone else. That part is sound.

Two gaps before this delivers what the issue asks for.

### 1. The real access check is untouched

The custom cap is only consulted when registering the menu. The actual capability resolution for settings objects lives in `includes/access.php:535-538`, which still hardcodes:

```php
$capabilities['read'] = 'manage_options';
$capabilities['edit'] = 'pods_edit_' . $info['object_name'];
```

`PodsAdmin::admin_content_settings()` (`classes/PodsAdmin.php:760`) performs no capability check of its own — it relies entirely on menu registration. So a user granted only the custom cap gets the menu item, but the path that governs actual read/edit access never learns about it.

Worth mirroring the custom cap into that `access.php` branch so both agree.

### 2. Unsanitized option

`admin_capabilities()` injects the raw user-supplied string, where the neighbouring post-type branch uses `pods_v_sanitized()` (`classes/PodsAdmin.php:3995`). Not an escalation vector — that method is only hooked to `members_get_capabilities` — but it's inconsistent with the surrounding code. Suggest `pods_v_sanitized( 'capability_type_custom', ... )` in both places.

### Minor

The PR checklist ticks "includes automated tests", but neither changed file is a test. A wpunit test asserting that a user with only the custom cap can reach the settings page (and that one without it cannot) would cover both points above.

PR #7580, opened 2026-07-26
fix(data): guard pods_trim() against null input for PHP 8.1+

### LINKED ISSUE #7262: Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /wp-content/plugins/pods/includes/data.php on line 298
### Description

Hey!

Testing Pods 3.1.1 on WP 6.4.3, running PHP 8.1.27 - so far, no issues, nice work 👏

But with WP_DEBUG enabled, I get the following deprecation warning:
`Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /wp-content/plugins/pods/includes/data.php on line 298`

No biggie, sites work fine, just a heads-up. Let me know if I can provide any helpful information :)

### Version

3.1.1

### Testing Instructions

_No response_

### Screenshots / Screencast

_No response_

### Possible Workaround

_No response_

### Site Health Information

```text
on request
```

### Pods Package

```json
on request
```

faisalahammad:
## Description

A PHP 8.1+ `Deprecated: trim(): Passing null to parameter #1 ($string)` notice was emitted from `includes/data.php` whenever `pods_trim()` reached `trim()`/`ltrim()`/`rtrim()` via `call_user_func_array` with a null value. The function had no early guard and relied on a `(string)` cast only in its scalar branch, leaving array and object recursion paths exposed.

Added a 3-line early-return guard at the top of `pods_trim()`, matching the existing `pods_sanitize()` idiom (line 27). One change covers the only entry point of the trim dispatcher, so every caller (top-level and recursive) is protected. `null` is now returned unchanged instead of producing a deprecation notice. All non-null behavior is unchanged.

```diff
 function pods_trim( $input, $charlist = " \t\n\r\0\x0B", $lr = null ) {
+	if ( '' === $input || is_int( $input ) || is_float( $input ) || empty( $input ) ) {
+		return $input;
+	}
+
 	if ( is_object( $input ) ) {
```

## Related GitHub issue(s)

Fixes #7262

## Testing instructions

1. Install on a WordPress 6.4+ / PHP 8.1+ site with `WP_DEBUG=true`.
2. Trigger any code path that calls `pods_trim()` internally (saving a Pod, registering a Pod with a custom REST namespace, etc.).
3. Confirm `wp-content/debug.log` no longer contains entries like `Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /wp-content/plugins/pods/includes/data.php on line [298-299]`.
4. Confirm all Pods admin screens still load, Pods save normally, and REST routes still resolve.
5. Automated: `slic run wpunit --ext DotReporter tests/codeception/wpunit/functions/DataTest.php` runs the new `test_pods_trim_null()` method, covering scalar null with/without charlist/direction, empty string, `'0'`, integer `0`, float `0.0`, null leaf inside an array, and null leaf inside an object.

## Screenshots / screencast

Not applicable.

Bug: pods_trim() no longer triggers a PHP 8.1+ trim() deprecation notice when called with null values. #7262 (@faisalahammad)

## PR checklist


faisalahammad:

faisalahammad:
## Review: this looks already fixed on the base branch

I verified `pods_trim()` on `release/3.4.0` and the deprecation this targets is gone. The function builds its arguments as:

```php
$args = [
    (string) $input,
];
```

Every leaf — including the recursive array and object branches above it — is cast before `call_user_func_array()`, so `trim()` can no longer receive `null`. I also confirmed the original bug was real at tag `3.1.1`, where the same line was `$args = array( $input );` with no cast. The cast has since been added.

### The guard also changes return types

The new early return alters behaviour for existing callers, in ways the added test locks in rather than catches:

| call | before | after |
|---|---|---|
| `pods_trim( 5 )` | `'5'` | `5` |
| `pods_trim( 0.0 )` | `'0'` | `0.0` |
| `pods_trim( false )` | `''` | `false` |
| `pods_trim( null )` | `''` | `null` |

The docblock still says `@return array|object|string`. Anything doing a strict comparison on the result of `pods_trim()` would silently change behaviour.

The test isn't hollow — it does fail on base — but it fails because it asserts the *new* types, not because a deprecation is being triggered.

### Suggestion

Close as already-fixed against the existing `(string)` cast. If you'd still like belt-and-braces hardening, keep it type-preserving:

```php
if ( null === $input ) {
    return '';
}
```

That guards the null case without changing the return type for any other input.

PR #7594, opened 2026-08-09
test: add regression coverage for Media pod REST save

### LINKED ISSUE #6420: Cannot modify custom field via REST API if added to Media content type
### Description

I have added a field named "checksum" to a custom post type and to the media type:

Of course 1) REST API tab / Enable is checked and 2) the field options / REST API / both Read and Write via REST is enabled.

Via the rest endpoint, I can view the value of checksum on each media page that I've added via the WP editor. However I cannot update it via the REST endpoint. 

- In the json uploaded, the property looks like: ```{ ... "checksum": "42" ... }```
- The JSON returned via the post method: contains ```{..."checksum": [False] ... }```
and the value is not updated.

This happens only to the Media type. The same exact field works correctly in a custom content type I created.

### Version

2.8.8.1

### Testing Instructions

_No response_

### Screenshots / Screencast

image

### Possible Workaround

_No response_

### Site Health Information

_No response_

### Pods Package

_No response_

faisalahammad:
## Description

Adds a regression test for the Media pod REST save path. A Pods custom field added to the Media (attachment) pod could be read via the WP REST API but could not be written: a `POST` to `/wp/v2/media/<id>` returned the field as `[false]` and the value never persisted. The same field worked on a custom post type.

Root cause was a lookup key mismatch in `PodsRESTHandlers::save_handler()`. `register_rest_field( 'attachment', ... )` stores the field under the `attachment` key in the `$wp_rest_additional_fields` global, but the handler was keying by the pod name `media`, so the field was never matched during the save loop. The fix keying the lookup by the WP post type is already present in `classes/PodsRESTHandlers.php`.

The new test directly invokes `save_handler()` on a media pod and asserts the field is persisted. It fails if the lookup regresses to the pod name, locking the fix in place. No production code changes in this PR.

## Related GitHub issue(s)

Fixes #6420

## Testing instructions

1. Run the wpunit REST group: `./vendor/bin/codecept run wpunit --group=pods-rest` (covered by the existing `wpunit` matrix entry in `.github/workflows/tests-php.yml`).
2. Confirm the 3 tests in this class pass.
3. To confirm the test catches the original bug, revert the `$type`-vs-`$pod_name` lookup in `PodsRESTHandlers::save_handler()` (change `$wp_rest_additional_fields[ $type ]` back to `$wp_rest_additional_fields[ $pod_name ]` on lines 287 and 297) and re-run. `should_save_field_on_media_pod_via_save_handler` fails with `'' != '42'`. Restore the lookup to re-green.

## Screenshots / screencast

None. No UI changes.

Bug: A Pods custom field on the Media (attachment) pod can now be updated via the WP REST API. #6420 (@faisalahammad)

## PR checklist

### DIFF
--- a/tests/codeception/wpunit/Pods/PodsRESTHandlersMediaTest.php
+++ b/tests/codeception/wpunit/Pods/PodsRESTHandlersMediaTest.php
@@ -1,0 +1,211 @@
+<?php
+
+namespace Pods_Unit_Tests\Pods;
+
+use PodsRESTFields;
+use PodsRESTHandlers;
+use Pods_Unit_Tests\Pods_UnitTestCase;
+use WP_Post;
+use WP_REST_Request;
+
+/**
+ * Regression test for Pods Framework issue #6420.
+ *
+ * A Pods custom field added to the Media (attachment) pod could not be
+ * updated via the WP REST API. Reading worked, but the save path failed
+ * silently, leaving the field empty and the read returning [false]. The
+ * same field worked on a custom post type.
+ *
+ * Root cause was a lookup key mismatch in PodsRESTHandlers::save_handler():
+ * register_rest_field( 'attachment', ... ) stores the field under the
+ * 'attachment' key in $wp_rest_additional_fields, but save_handler was
+ * keying by the pod name 'media' instead. The fix (commit 6a62c35b0) keys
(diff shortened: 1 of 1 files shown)


faisalahammad:

PR #7605, opened 2026-08-14
feature: add Minimum Length option for text fields

### LINKED ISSUE #7413: Add a "Minimum Length" Option for Text Fields (Quick Win!)
### Problem to Solve

Use case: We are using Pods to describe (e.g. books, theirs uploads etc). When it comes to "description" (text-) fields, many authors/editors/users are a bit too lazy to write enough text. That's why we would like to force users to write at least -n- characters, but it seems that there is currently no solution for this?!
Currently, for most of your text-based field types, such as “Plain Paragraph Text” or “Plain Text” or “WYSIWYG” etc., there is a “Maximum Length” option. But a “Minimum length” option is missing.

### Proposed Solution

Add a "Minimum length" option (Its more or less a "copy and paste" job of the existing "Maximum length" field.

### Possible Workaround

nothing found yet.

### Examples Elsewhere

_No response_

### Screenshots / Screencast

_No response_

faisalahammad:
## Description

Adds a **Minimum Length** option to the Plain Text, Plain Paragraph Text, and WYSIWYG field types. It mirrors the existing Maximum Length option, but enforces a floor instead of a ceiling: a non-empty value shorter than the configured minimum is rejected on save with a clear error.

Maximum Length works by silently truncating the value, which is not possible for a minimum. So instead the new option validates in each field's `validate()` method and blocks the save when the cleaned value is shorter than the minimum. Blank values are left to the existing required-field check, so a minimum is never enforced on an empty field.

## Related GitHub issue(s)

Fixes #7413

## Testing instructions

1. Edit (or add) a Plain Text, Plain Paragraph Text, or WYSIWYG field.
2. Open its options and set **Minimum Length** to a number (for example `10`). Save the field.
3. Edit an item and enter a value shorter than the minimum, then save.
4. See the save blocked with the message "<Field Label> must be at least N characters long."
5. Enter a value at or above the minimum and save again.
6. Confirm it saves successfully, and that leaving the field empty saves normally.

## Screenshots / screencast

None. This is an admin option plus server-side validation, with no visible front-end change beyond the existing error messaging.

Enhancement: Text, Paragraph, and WYSIWYG fields now support a Minimum Length option that requires non-empty values to meet a character floor on save. #7413 (@faisalahammad)

## PR checklist


faisalahammad:

sc0ttkclark (changes requested):
This should probably abstract something the min length logic into a Field_Validation trait, maybe in src/Pods/Data/Traits/Field_Validation.php

Also add tests for the fields in `tests/codeception/wpunit/Pods/Field` or add a test for the trait itself

faisalahammad:
Hi @sc0ttkclark, thanks for the review.

I moved the min length check into a trait at `src/Pods/Data/Traits/Field_Validation.php`. Text, Paragraph, and WYSIWYG fields now use it instead of each having their own copy.

I also added tests for all three fields in `tests/codeception/wpunit/Pods/Field/`. They cover value below minimum, value at minimum, empty value with minimum set, and minimum not set.

Changes are in commit 648b20b5f. Ready for another look.

PR #7606, opened 2026-08-16
fix: check for custom plugins directory in init.php

### LINKED ISSUE #7361: Check for custom plugins directory.
https://github.com/pods-framework/pods/blob/20f681a61d80f9f4d272d7d407bf92d61ef8ade7/init.php#L153

in `init.php` several deactivation functions check the hard-coded `plugins/pods-ui` directory. 

This may prevent devs from changing the name of the `plugins` directory. Which often a security recomendation. Eg:
```
define('WP_PLUGIN_DIR', dirname(__FILE__) . '/../wp-content/extensions'); // no host name, no trailing slash
define('WP_PLUGIN_URL', 'http://localhost.com/site/wp-content/extensions');
```

Can we improve this so it checks for alternative directory structures? I believe wp-migrate-lite does something like this here: https://github.com/deliciousbrains/wp-migrate-db/blob/f0bd63fae078f3c6added0b3db18983a2a6d4e07/compatibility/wp-migrate-db-pro-compatibility.php#L17

(Also, was `pods-ui` the original directory name way back in the day?)

faisalahammad:
## Description

The deactivation functions in `init.php` (`pods_deactivate_pods_duplicate()` and `pods_deactivate_pods_ui()`) referenced the old Pods UI plugin using a hard-coded `WP_CONTENT_DIR . 'plugins/pods-ui/pods-ui.php'` path. This ignores a customized plugin directory (sites that define `WP_PLUGIN_DIR`, a common security hardening step) and could silently fail to detect/deactivate a conflicting copy.

The paths now use `trailingslashit( WP_PLUGIN_DIR ) . 'pods-ui/pods-ui.php'`, which follows the plugin directory the site actually uses. This also fixes a latent bug: the old string was missing a slash between `WP_CONTENT_DIR` and `plugins`, so its `file_exists()` check always returned false.

## Related GitHub issue(s)

Fixes #7361

## Testing instructions

1. Install and activate Pods normally (no custom plugin dir) and confirm activation and admin work as expected.
2. In `wp-config.php`, define `WP_PLUGIN_DIR` and `WP_PLUGIN_URL` to a non-default directory (e.g. `wp-content/extensions`).
3. Place a conflicting `pods-ui/pods-ui.php` in that custom directory.
4. Confirm Pods resolves the path against the custom directory and can detect/deactivate the conflicting plugin instead of always failing the check.

Run `vendor/bin/phpstan analyse init.php` (no errors) and `vendor/bin/phpcs init.php` (only pre-existing errors on `main` remain).

## Screenshots / screencast

No visible UI changes.

Bug: Pods now respects a custom plugin directory when checking for a conflicting Pods UI plugin. #7361 (@faisalahammad)

## PR checklist


faisalahammad:

PR #7617, opened 2026-08-21
Rebuild DFV bundle to drop stale tabindex="2"

### LINKED ISSUE #7402: Custom Setting Page - file fields `Add File` buttons do not have correct tabbing order
### Description

In a custom settings page created by Pods, file fields `Add File` buttons have `tabindex=2` added, which breaks the tabbing order if I'm not wrong.

### Version

3.2.8.2

### Testing Instructions

1. Create a custom setting page.
2. Create a field group, then add a text field, a file upload field.
3. Focus on the text field.
4. Then press `Tab` key and "Add File" button of the file upload field is not focused by the Tab key.

### Screenshots / Screencast

This is the current tabbing order.

Image

### Possible Workaround

- Update the `tabindex` attribute.
- Or, updating html tag to `button` so we can get rid of `tabindex` attribute and let browser handle that automatically.

### Site Health Information

```text

```

### Pods Package

```json

```

faisalahammad:
## Description

Regenerates the committed DFV bundle (`ui/js/dfv/pods-dfv.min.js`) on `release/3.4.0`.

The React file field refactor removed the hard-coded positive `tabindex="2"` from the DFV source, but the committed minified artifact was stale and still shipped the old markup at runtime. Because the bundle is what actually loads, keyboard users on custom settings pages and in the block editor kept tabbing to the wrong control or skipping the file field — the accessibility problem originally reported in #7402.

This PR rebuilds from current source so the committed artifact matches. The only code change in the bundle is the removal of both `tabindex="2"` occurrences (2 → 0).

## Related GitHub issue(s)

Fixes #7402

## Testing instructions

1. On a Pods custom settings page with a text field followed by a file upload field, focus the text field.
2. Press `Tab` and confirm focus reaches the file field's `Add File` control at its natural document position.
3. Press `Enter` on the `Add File` control and confirm the uploader opens.
4. Repeat in the block editor where the file field is used.

## Verification

```
grep -c 'tabindex="2"' ui/js/dfv/pods-dfv.min.js   # => 0 (was 2 before rebuild)
```

The `ui/js/blocks/pods-blocks-api.*` files carry only a regenerated cache-busting hash in the `.asset.json`; no source change.

Bug: Rebuilt DFV bundle so the file field Add File control no longer ships a stale hard-coded tabindex. #7402 (@faisalahammad)

## PR checklist

- [x] My code follows the WordPress Coding Standards.

---

_Disclosure: this PR and its description were prepared with AI assistance and reviewed by me before submitting._


sc0ttkclark:
Fixed via 3b61b9f595d580d4b7898a413eb2feb9dcd414ce

